PR #33, opened 2019-06-11
Report cluster state as part of artifacts

### LINKED ISSUE #32: collect cluster state for every run
To assist in debugging failed tests, as well as to know how a cluster was configured for a specific test run, we need to collect additional information from the cluster.

Please evaluate and integrate as much of this as possible into osde2e:

I looked into what origin is doing:

https://github.com/openshift/release/blob/80af9fee7a9f63a79e01da0c74d9dd323118daf0/ci-operator/templates/openshift/installer/cluster-launch-installer-e2e.yaml#L666

It runs a bunch of commands, and also runs the `must-gather` tool, which is meant to define what gets collected for debugging openshift clusters:

https://github.com/openshift/must-gather/

ethernetdan:
This PR:
* Adds helper method to collect defined set of resources from cluster
* Add test suite to call method and write results
* Add new test suite to osde2e
* Modify testgrid package to support gzip data 

Closes #32

### DIFF
--- a/e2e_test.go
+++ b/e2e_test.go
@@ -7,6 +7,7 @@ import (
 
 	// import suites to be tested
 	_ "github.com/openshift/osde2e/test/openshift"
+	_ "github.com/openshift/osde2e/test/state"
 	_ "github.com/openshift/osde2e/test/verify"
 )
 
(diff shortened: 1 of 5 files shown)


jeremyeder:
@stevekuznetsov fyi

PR #55, opened 2019-07-04
Set cluster expiration timestamp

jhernand:
This patch changes the tests so that they always set an
`expiration_timestamp` for the created clusters. That way even if the
tests crash or forget to delete the cluster for whatever the reason the
cluster will be automatically removed after eight hours.

### DIFF
--- a/pkg/osd/cluster.go
+++ b/pkg/osd/cluster.go
@@ -14,6 +14,10 @@ import (
 func (u *OSD) LaunchCluster(cfg *config.Config) (string, error) {
 	log.Printf("Creating cluster '%s'...", cfg.ClusterName)
 
+	// Calculate an expiration date for the cluster so that it will be automatically deleted if
+	// we happen to forget to do it:
+	expiration := time.Now().Add(8 * time.Hour)
+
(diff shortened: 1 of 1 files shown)


jhernand:
@ethernetdan @jeremyeder @vkareh please review.

jhernand:
Good point @vkareh. However I am not completely sure of what should be done when `NO_DESTROY` is enabled. I assume that this flag is intended for development: you run the tests locally and you may not want to immediately destroy the cluster because you may want to inspect it after running the tests. But even in that case woudl't it be good to automatically destroy the cluster after a few hours? In a few hours you should have enough time to inspect the cluster. What if you forget then to delete it?

vkareh:
@jhernand, makes sense to me.
@ethernetdan, thoughts about this? Is that part of the intention behind the `NO_DESTROY` flag?

ethernetdan:
@jhernand  @vkareh That makes sense to me

We should make sure we get all the info we need to triage the failures that cause the suite to hit the limit. I'll merge this now and we can add additional archiving as needed.

PR #82, opened 2019-08-13
take NO_DESTROY env for docker-test runs

clcollins:
Allow NO_DESTROY environment variable to be used by default in the Makefile target for docker-test.  If unset, cluster will continue to be destroyed as it currently is.

In practice, this considerably reduces the amount of times I accidentally delete a test cluster while doing iterative tests (I can source a file with cluster details and NO_DESTROY=true).

Signed-off-by: Christopher Collins <collins.christpher@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,7 @@ docker-test:
 	docker run \
 		-t \
 		--rm \
+		-e NO_DESTROY=$(NO_DESTROY) \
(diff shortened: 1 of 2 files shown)


clcollins:
Accidentally got further commits included in this PR.  I'll close and squash.

PR #92, opened 2019-09-16
Increase olm grace period

clcollins:
Moved poller timeouts to single global setting for pollers; increased to 60m.  OLM in normal operation takes ~15 minutes to be installed and begin installing operators, so the previous 20m timeout left little wiggle room for any other latency.


jeefy:
/lgtm

Yay consistency

PR #95, opened 2019-09-23
SDAR-30. Remove CLEAN_RUNS from upgrade tests.

meowfaceman:
CLEAN_RUNS have been removed from upgrade tests so that upgrade tests
run more regularly.


jeefy:
I'm also pro keeping the separate upgrade events script in the event we need to change other upgrade-specific test params. 

/lgtm

PR #96, opened 2019-09-24
SDAR-31. Add an hour wait after test completion.

meowfaceman:
After tests complete (successfully or otherwise), wait one hour to allow
for any active debugging.

Note: The lack of env/sect for time.Duration is intended, as apparently the doc generation barfs otherwise.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -42,6 +42,9 @@ type Config struct {
 	// MinorTarget is the minor version to target. If specified, it is used in version selection.
 	MinorTarget int64 `env:"MINOR_TARGET" sect:"version"`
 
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -42,6 +42,9 @@ type Config struct {
 	// MinorTarget is the minor version to target. If specified, it is used in version selection.
 	MinorTarget int64 `env:"MINOR_TARGET" sect:"version"`
 
+	// AfterTestClusterWait is how long to keep a cluster around after tests have run.
+	AfterTestClusterWait time.Duration
```
jeefy: At some point we may want to be able to configure this per-run via an env var.

PR #98, opened 2019-09-26
Remove operator testing, make debug sleep optional

jeefy:
The debug sleep thing kind of hurts doing local testing/development because the clusters linger around. So, being able to toggle it with an env var seemed appropriate.

Also this PR comments out the dedicatedadmin test while we try to achieve some cleaner signal and I try to optimize its testing 

/cc @clcollins 
/assign @meowfaceman

### DIFF
--- a/setup.go
+++ b/setup.go
@@ -67,13 +67,17 @@ var _ = ginkgo.AfterSuite(func() {
 			cfg.AfterTestClusterWait = 60 * time.Minute
 		}
 
-		log.Printf("Sleeping for %d minutes before destroying cluster '%s'", cfg.AfterTestClusterWait/time.Minute, cfg.ClusterID)
-		startTime := time.Now()
-		for time.Now().Sub(startTime) < cfg.AfterTestClusterWait {
-			time.Sleep(1 * time.Minute)
-			log.Print(".")
+		if cfg.NoDestroyDelay {
+			log.Printf("Skipping sleep for cluster debugging")
+		} else {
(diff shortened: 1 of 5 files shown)


jeremyeder:
Wouldn't this disable operator tests for all pipelines, and not just the 4.2 pipeline?  We still need these tests to run in the 4.1 pipelines.

Consider a more granular approach (such as a gingko focus for each suite of tests?).

jeefy:
Point. I'll whip up something better this morning

jeefy:
PTAL

This is much more flexible. Once merged in, I'll update the 4.2-int job to skip the Dedicated Admin test.

setup.go
```diff
@@ -67,13 +67,17 @@ var _ = ginkgo.AfterSuite(func() {
 			cfg.AfterTestClusterWait = 60 * time.Minute
 		}
 
-		log.Printf("Sleeping for %d minutes before destroying cluster '%s'", cfg.AfterTestClusterWait/time.Minute, cfg.ClusterID)
-		startTime := time.Now()
-		for time.Now().Sub(startTime) < cfg.AfterTestClusterWait {
-			time.Sleep(1 * time.Minute)
-			log.Print(".")
+		if cfg.NoDestroyDelay {
```
meowfaceman: Good catch. Sorry about that!

jeefy: All good.

jeefy:
Had two lgtm's, decided to merge it

meowfaceman:
Jeez, I can't believe that PR check still hasn't run. Hopefully the executor increase will come through soon.

jeefy:
Same considering we're going to be adding even more tests LOL

PR #100, opened 2019-09-27
Cross-stream support, Dedicated Admin Operator timeout changes

jeefy:
This PR has two main things (it should be two PRs, won't lie.)

- Adds "TargetStream" option. Previously you could not target an install version and an upgrade version from different streams/channels. Now you can say "install 4.1-stable and upgrade to 4.2-nightly" \o/

- The Dedicated Admin Operator Test was bumping into the overall test timeout and causing the whole `ginkgo` suite to fail. This would cause `AfterSuite` to not trigger, and we would not get results reported out to TestGrid. 

  This PR re-tools some of the tests to be less-redundant (they're run serially so we don't need to see if the operator is installed if the first check passes) and then adds timeouts to every individual context. 

Questions/critiques welcome. Once this gets in, we can build 4.1-4.2 and 4.2-4.3 upgrade tests. :)


pkg/upgrade/releases.go
```diff
@@ -19,9 +19,7 @@ const (
 )
 
 // LatestRelease retrieves latest release information for given releaseStream. Will use Cincinnati for stage/prod.
-func LatestRelease(cfg *config.Config) (name, pullSpec string, err error) {
-	releaseStream := cfg.UpgradeReleaseStream
-
+func LatestRelease(cfg *config.Config, releaseStream string) (name, pullSpec string, err error) {
```
meowfaceman: Nit: Do we need the releaseStream argument since it's part of config anyway?

test/operators/dedicatedadmin.go
```diff
@@ -57,34 +57,31 @@ var roleBindings = []string{
...
 
+	// Check that the operator deployment exists in the operator namespace
+	ginkgo.Context("deployments", func() {
+		ginkgo.It("should exist", func() {
 			deployments, err := pollDeploymentList(h)
 
 			Expect(err).ToNot(HaveOccurred(), "failed fetching deployments")
 			Expect(deployments).NotTo(BeNil())
-		})
+		}, float64(globalPollingTimeout * 60)
```
meowfaceman: Kind of confused about the * 60s in this code. Rather than do this, why not make the globalPollingTimeout 30 * 60 and skip all of the multiplies?

clcollins: I would have defaulted to keeping the variable a human-friendly (heh, or just, non-math-y) single value, but you have a point too.  I could go either way and don't feel strongly about either.

pkg/upgrade/releases.go
jeefy: Yes. `LatestRelease` looks for the latest valid release in a specific stream. In some cases we pass `cfg.UpgradeReleaseStream` and in some we pass `cfg.TargetStream`

We also pass the `cfg` var because LatestRelease will update it with version info based on the stream passed.

meowfaceman: SGTM. Will resolve!

test/operators/dedicatedadmin.go
meowfaceman: I'd elect for baking the 60 into the variable if you don't mind. I think it'll minimize complexity for adding in tests later.

test/operators/dedicatedadmin.go
```diff
@@ -28,7 +28,7 @@ import (
 )
 
 // timeout is the duration in minutes that the polling should last
```
clcollins: This is no longer true - globalPollingTimeout is now the time in seconds

jeefy: Good catch. Fixed.

PR #102, opened 2019-09-27
fixing syntax error in version.go

clcollins:


jeefy:
Thanks. Now I need to figure out why vscode didn't catch that.

PR #107, opened 2019-09-29
Look for version prefix expected by cluster

jeefy:
Clusters are expecting versions (from streams) to have `openshift-v`. This looks to see if the returned version contains the string and, if not, prepends it. Tested extensively locally with great success~

Also sneaking in a 300s wait time between cluster-install and tests. 

/assign @meowfaceman @clcollins

### DIFF
--- a/version.go
+++ b/version.go
@@ -28,8 +28,12 @@ func ChooseVersions(cfg *config.Config, osd *osd.OSD) (err error) {
 // chooses between default version and nightly based on target versions.
 func setupVersion(cfg *config.Config, osd *osd.OSD) (err error) {
 	if cfg.MajorTarget == 0 && cfg.MinorTarget == 0 {
-		// use defaults if no version targets
-		if cfg.ClusterVersion, err = osd.DefaultVersion(); err == nil {
+		// check to see if a target stream is set.
+		// a target stream and a major/minor target should not be set at the same time.
+		if cfg.ClusterVersion, _, err = upgrade.LatestRelease(cfg, cfg.TargetStream); err == nil {
+			log.Printf("Target Release Stream set, using version  '%s'", cfg.ClusterVersion)
+			// use defaults if no version targets
+		} else if cfg.ClusterVersion, err = osd.DefaultVersion(); err == nil {
 			log.Printf("CLUSTER_VERSION not set, using the current default '%s'", cfg.ClusterVersion)
 		}
 	} else {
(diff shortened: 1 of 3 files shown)


version.go
```diff
@@ -29,7 +30,13 @@ func ChooseVersions(cfg *config.Config, osd *osd.OSD) (err error) {
 func setupVersion(cfg *config.Config, osd *osd.OSD) (err error) {
 	if cfg.MajorTarget == 0 && cfg.MinorTarget == 0 {
 		// use defaults if no version targets
-		if cfg.ClusterVersion, err = osd.DefaultVersion(); err == nil {
+		if cfg.ClusterVersion, _, err = upgrade.LatestRelease(cfg, cfg.TargetStream); err == nil {
+			if !strings.Contains(cfg.ClusterVersion, "openshift-v") {
+				log.Println("Version didn't have prefix. Adding....")
```
meowfaceman: Would be good to know which version didn't have the prefix.

version.go
```diff
@@ -38,6 +45,19 @@ func setupVersion(cfg *config.Config, osd *osd.OSD) (err error) {
 			cfg.MajorTarget = -1
 		}
 
+		if len(cfg.TargetStream) != 0 {
+			log.Printf("Looking for a release on stream %s", cfg.TargetStream)
+			if cfg.ClusterVersion, _, err = upgrade.LatestRelease(cfg, cfg.TargetStream); err == nil {
+				if !strings.Contains(cfg.ClusterVersion, "openshift-v") {
```
meowfaceman: It seems like we could make this into a function and combine the above logic with it.

version.go
```diff
@@ -52,9 +72,18 @@ func setupUpgradeVersion(cfg *config.Config, osd *osd.OSD) (err error) {
 		return fmt.Errorf("couldn't get latest release from release-controller: %v", err)
 	}
 
+	log.Printf("Target stream: '%s', Upgrade stream: '%s'", cfg.TargetStream, cfg.UpgradeReleaseStream)
+
 	if len(cfg.TargetStream) != 0 {
+		log.Printf("Looking for a release on target stream %s", cfg.TargetStream)
 		if cfg.ClusterVersion, _, err = upgrade.LatestRelease(cfg, cfg.TargetStream); err == nil {
+			if !strings.Contains(cfg.ClusterVersion, "openshift-v") {
```
meowfaceman: Ditto.

jeefy:
This was something I ran into over the weekend and didn't JIRA. /shrug

I'll start creating a JIRA entry for any bug I find and then fix it. I'll also refactor this to be a bit cleaner. :)

version.go
jeefy: added in new func

version.go
jeefy: absolutely

jeefy:
@meowfaceman PTAL when around :) Should be cleaner.

PR #108, opened 2019-09-30
Wrap ginkgo skip in quotes

jeefy:
Quick fix for any Ginkgo skip that's multiple words.


jeefy:
Quick fix I wanted merged in before another int-4.2 run :)

jeefy:
As a note, this was for SDAR-32

PR #109, opened 2019-09-30
if upgrade and target are the same, look for the previous version

jeefy:
Addresses [SDAR-39](https://jira.coreos.com/browse/SDAR-39)

/assign @meowfaceman


meowfaceman:
Do we have a JIRA for this?

PR #128, opened 2019-10-09
fixing extended timeout times; reduced from 30h to 30m

clcollins:
Fixing a regression from previous commit that does not take into account the new unit size for the globalPollingTimeout variable.

Fixes the 30h polling wait (reduced to 30m).


clcollins:
@jeefy beat me to it: #127

PR #130, opened 2019-10-16
SDCICD-40. Use MAJOR_TARGET and MINOR_TARGET to target CIS versions.

meowfaceman:
The MAJOR_TARGET and MINOR_TARGET variables will now target versions
that come directly from the OSD cluster versions endpoint without being
targeted at a specific type of nightly.


clcollins:
:+1:

PR #135, opened 2019-10-24
Use cvo and osd-verify checks to determine cluster health

jeefy:
This mimics the osd-verify script SREs use to determine cluster health. It's checking the CVO for cluster state, then checking operator, node, and pod health. 

Once a cluster is deemed healthy, it will continue polling until 5 clean polls occur, or will attempt to do so until the setup timeout.

### DIFF
--- a/pkg/helper/operators.go
+++ b/pkg/helper/operators.go
@@ -1,0 +1,32 @@
+package helper
+
+import (
+	"log"
+
+	osconfig "github.com/openshift/client-go/config/clientset/versioned"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// CheckOperatorReadiness attempts to look at the state of all operator and returns true if things are healthy.
+func CheckOperatorReadiness(oscfg *osconfig.Clientset) (bool, error) {
+	success := true
+	log.Print("Checking that all Operators are running or completed...")
+
+	configClient, getOpts := oscfg.ConfigV1(), metav1.ListOptions{}
+	list, err := configClient.ClusterOperators().List(getOpts)
+	if err != nil {
+		log.Printf("Error getting CVS: %v\n", err)
+		return false, nil
+	}
+
+	for _, co := range list.Items {
+		for _, cos := range co.Status.Conditions {
+			if (cos.Type != "Available" || cos.Status != "True") && cos.Type != "Upgradable" {
+				log.Printf("Operator %v type %v is %v: %v", co.ObjectMeta.Name, cos.Type, cos.Status, cos.Message)
+				success = false
+			}
+		}
+	}
+
+	return success, nil
+}
--- a/pkg/helper/pods.go
+++ b/pkg/helper/pods.go
@@ -8,6 +8,7 @@ import (
 
 	kubev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 )
 
 // WaitForPodPhase until in target, checking n times and sleeping dur between them. Last known phase is returned.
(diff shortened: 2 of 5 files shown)


pkg/helper/operators.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"log"
+
+	osconfig "github.com/openshift/client-go/config/clientset/versioned"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// CheckOperatorHealth attempts to look at the state of all operator and returns true if things are healthy.
+func CheckOperatorHealth(oscfg *osconfig.Clientset) (bool, error) {
+	fail := true
```
meowfaceman: Let's name this "success," since this more accurately reflects how this is being used. e.g. "success is true unless it's set to false by the later check.

pkg/helper/operators.go
```diff
@@ -0,0 +1,40 @@
...
+	configClient, getOpts := oscfg.ConfigV1(), metav1.ListOptions{}
+	list, err := configClient.ClusterOperators().List(getOpts)
+	if err != nil {
+		log.Printf("Error getting CVS: %v\n", err)
+		return false, nil
+	}
+
+	for _, co := range list.Items {
+		for _, cos := range co.Status.Conditions {
+			if cos.Type == "Available" && cos.Status != "True" {
```
meowfaceman: For this, why don't we say:

```
if (cos.Type != "Available" || cos.Type != "True") && cos.Type != "Upgradable" {
    log.Printf("Operator %v type %v is %v: %v", co.ObjectMeta.Name, cos.Type, cos.Status, cos.Message)
    fail = false
}
```

This should be logically equivalent. I think. Feel free to poke holes in that.

setup.go
```diff
@@ -109,13 +109,10 @@ func setupCluster(cfg *config.Config) (err error) {
 		log.Printf("CLUSTER_ID of '%s' was provided, skipping cluster creation and using it instead", cfg.ClusterID)
 	}
 
-	if err = OSD.WaitForClusterReady(cfg.ClusterID, cfg.ClusterUpTimeout); err != nil {
+	if err = OSD.WaitForClusterReady(cfg, cfg.ClusterID, cfg.ClusterUpTimeout); err != nil {
```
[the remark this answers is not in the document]
jeefy: Good point.

pkg/osd/cluster.go
```diff
@@ -120,19 +125,97 @@ func (u *OSD) DeleteCluster(clusterID string) error {
...
+	cleanRuns := 0
 
 	return wait.PollImmediate(45*time.Second, timeout, func() (bool, error) {
 		if state, err := u.ClusterState(clusterID); state == v1.ClusterStateReady {
-			return true, nil
+			healthy, err := u.PollClusterHealth(cfg)
+			if err != nil {
+				return false, err
+			}
+			if healthy == true && cleanRuns < 5 {
```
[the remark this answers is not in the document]
jeefy: so fresh so clean

pkg/helper/pods.go
```diff
@@ -32,3 +33,32 @@ func (h *H) WaitForPodPhase(pod *kubev1.Pod, target kubev1.PodPhase, n int, dur
 	Expect(phase).NotTo(BeEmpty())
 	return
 }
+
+// CheckPodHealth attempts to look at the state of all pods and returns true if things are healthy.
+func CheckPodHealth(kubeclient kubernetes.Interface) (bool, error) {
+	var (
```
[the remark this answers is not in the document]
jeefy: We totally can, this was my IDE doing weird things as I copied/cleaned up.

pkg/osd/cluster.go
```diff
@@ -120,19 +125,97 @@ func (u *OSD) DeleteCluster(clusterID string) error {
 }
 
 // WaitForClusterReady blocks until clusterID is ready or a number of retries has been attempted.
-func (u *OSD) WaitForClusterReady(clusterID string, timeout time.Duration) error {
+func (u *OSD) WaitForClusterReady(cfg *config.Config, clusterID string, timeout time.Duration) error {
 	log.Printf("Waiting %v for cluster '%s' to be ready...\n", timeout, clusterID)
+	cleanRuns := 0
 
 	return wait.PollImmediate(45*time.Second, timeout, func() (bool, error) {
```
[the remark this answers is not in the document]
jeefy: I'll switch this to 30s. It might make the logs somewhat chattier during install but it will resolve this faster AND give us more granularity on events.

pkg/helper/operators.go
jeefy: I'll nit further and change it to `CheckOperatorReadiness`

pkg/helper/operators.go
jeefy: nice

pkg/helper/operators.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"log"
+
+	osconfig "github.com/openshift/client-go/config/clientset/versioned"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// CheckOperatorReadiness attempts to look at the state of all operator and returns true if things are healthy.
+func CheckOperatorReadiness(oscfg *osconfig.Clientset) (bool, error) {
+	fail := true
```
meowfaceman: This still needs to change. :-)

meowfaceman:
One last nit, then I'll merge.

PR #136, opened 2019-10-25
SDCICD-35. Excise TestGrid publishing.

meowfaceman:
TestGrid publishing is no longer necessary since we directly build on
prow.


jeremyeder:
This is fine.

PR #138, opened 2019-10-25
SDICD-48 - Use CVO / Cluster State to determine when to start tests

jeefy:
This is #135 with multiple fixes and additional unit tests for the logic. 

/assign @meowfaceman

### DIFF
--- a/pkg/helper/nodes_test.go
+++ b/pkg/helper/nodes_test.go
@@ -22,7 +22,7 @@ func TestCheckNodeHealth(t *testing.T) {
 		expected    bool
 		objs        []runtime.Object
 	}{
-		{"no pods", false, nil},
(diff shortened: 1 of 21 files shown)


pkg/helper/nodes_test.go
```diff
@@ -0,0 +1,77 @@
...
+			Conditions: conditions,
+		}}
+}
+func TestCheckNodeHealth(t *testing.T) {
+	var tests = []struct {
+		description string
+		expected    bool
+		objs        []runtime.Object
+	}{
+		{"no pods", false, nil},
```
meowfaceman: nit: Should this be "no nodes?"

jeefy: Yes. :)

PR #139, opened 2019-10-25
Fix typo, add one more operator test

jeefy:
I can't spell, and this causes another loop-till-timeout when checking Operators.


jeefy:
Merging since without this tests **will** fail.

PR #142, opened 2019-10-29
Allow polling timeout to be configurable

cblecker:


docs/Options.md
```diff
@@ -40,6 +40,13 @@ These options are required to run osde2e.
 
 - Type: `string`
 
+### `POLLING_TIMEOUT`
+
+- PollingTimeout is how long to wait for an object to be created before
```
jeefy: Small discussion point: Is this for "objects" or for-specifically operators? Should we rename this completely to be operator-specific on the likelihood we will have additional objects we'll want to create-or-timeout with?

pkg/config/load.go
```diff
@@ -15,6 +15,23 @@ func (c *Config) LoadFromEnv() {
 	v := reflect.ValueOf(c).Elem()
 	for i := 0; i < v.Type().NumField(); i++ {
 		f := v.Type().Field(i)
+		if defaultValue, ok := f.Tag.Lookup(DefaultTag); ok {
```
jeefy: We actually have a couple areas where we do if-nil-hard-coded-default. I like this better and I'll have a follow-up PR to update those.

docs/Options.md
cblecker: Right now it's only being used for operators. I'm happy to rename if there is a better suggestion.

jeefy: It might warrant renaming later, for now I think it's reasonable. It was mainly a point for discussion. Thoughts @meowfaceman?

PR #143, opened 2019-10-30
SDCICD-63. Reintroduce NodeVertical tests.

meowfaceman:
The NodeVertical tests that @ethernetdan had been working on have been
reintroduced into osde2e. Additionally, the tests have been split into
general and scale tests, as the scale tests take much longer to run than
other tests, and as a result need to be split out.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,9 @@
 .PHONY: check generate build-image push-image push-latest test
 
 PKG := github.com/openshift/osde2e
+E2E_PKG := $(PKG)/suites/e2e
+SCALE_PKG := $(PKG)/suites/scale
+
 DIR := $(dir $(realpath $(firstword $(MAKEFILE_LIST))))
 
 IMAGE_NAME := quay.io/app-sre/osde2e
@@ -10,9 +13,9 @@ CONTAINER_ENGINE ?= docker
 
 check: cmd/osde2e-docs
 	go run $(PKG)/$< --check
-	CGO_ENABLED=0 go test -v ./cmd/... ./pkg/... 
+	CGO_ENABLED=0 go test -v ./cmd/... ./pkg/...
 	go get -u github.com/golangci/golangci-lint/cmd/golangci-lint && \
-    golangci-lint run ./... 
+	golangci-lint run ./...
 
 generate: docs/Options.md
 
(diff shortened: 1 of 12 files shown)


jeremyeder:
@chaitanyaenr please review

Makefile
```diff
@@ -27,7 +27,10 @@ push-latest:
 	@$(CONTAINER_ENGINE) --config=$(DOCKER_CONF) push "$(IMAGE_NAME):latest"
 
 test: out/osde2e
-	$< -test.v -ginkgo.skip="$(GINKGO_SKIP)" -test.timeout 8h
+	$< -test.v -ginkgo.skip="$(GINKGO_SKIP)" -test.timeout 8h -test.run TestE2E
+
+scale: out/osde2e
```
jeefy: I'd change this to be `make test-scale`

Makefile
```diff
@@ -27,7 +27,10 @@ push-latest:
...
+	$< -test.v -ginkgo.skip="$(GINKGO_SKIP)" -test.timeout 8h -test.run TestScale
 
 docker-test:
```
jeefy: And now that I think about it, maybe this should be `make test-docker`

jeefy:
Overall lgtm though I'll leave tagging it to @chaitanyaenr per @jeremyeder 

Once this is in I'd propose a separate set of jobs that run on each environment for the latest z-stream but not on upgrades once a day.

jeefy:
/lgtm

@meowfaceman on you to remove the hold

test/scale/nodevertical.go
```diff
@@ -0,0 +1,44 @@
...
+		scaleCfg := scaleRunnerConfig{
+			Name:         "node-vertical",
+			PlaybookPath: "workloads/nodevertical.yml",
+		}
+		r := scaleCfg.Runner(h)
+
+		// only test on 3 nodes
+		r.PodSpec.Containers[0].Env = append(r.PodSpec.Containers[0].Env, kubev1.EnvVar{
+			Name:  "NODEVERTICAL_NODE_COUNT",
+			Value: "3",
```
chaitanyaenr: The max_pods count needs to be changed to 750 instead of the default ( 1000 pods ) in case we are setting the nodevertical node count to 3 to avoid scheduling issues.

meowfaceman: This actually isn't necessary as the script multiplies the vertical node count * 250 and uses that to set the max pods count in the script used in scale.go.

PR #146, opened 2019-10-31
CICD-73 - Fix bool default logic

jeefy:
Original logic assumed if the default is set for a bool, it _must_ be true.

The broken tests indicate that that was a lie.


jeefy:
A) Hit the wrong button.
B) I'm going to manually push this in to fix our broken tests.

jeefy:
Yeah I'm doing that in a follow-up PR that I'll leave open.

PR #148, opened 2019-11-05
SDCICD-27. Make cluster expiry configurable and update to ocm-go-sdk.

meowfaceman:
Cluster expiry is now configurable and the old cluster cleanup code has
been excised. Additionally, the code has been migrated to use ocm-go-sdk
rather than uhc-go-sdk, as uhc-go-sdk is deprecated.

### DIFF
--- a/docs/Options.md
+++ b/docs/Options.md
@@ -84,14 +84,6 @@ before failing the test.
 - Type: `bool`
 - Default: `false`
 
-### `NO_DESTROY_DELAY`
-
-- NoDestroyDelay circumvents the 60min delay before a cluster is deleted
-This is highly useful when trying to debug things locally. :)
-
-- Type: `bool`
-- Default: `false`
-
 ### `OSD_ENV`
 
 - OSDEnv is the OpenShift Dedicated environment used to provision clusters.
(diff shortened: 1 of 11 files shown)


docs/Options.md
```diff
@@ -102,6 +94,13 @@ This is highly useful when trying to debug things locally. :)
 ## cluster
 
 
+### `CLUSTER_EXPIRY_IN_HOURS`
```
jeefy: This isn't defined in config.go -- ExpiryInMinutes is

pkg/config/config.go
```diff
@@ -51,12 +51,15 @@ type Config struct {
...
-	// NoDestroy leaves the cluster running after testing.
-	NoDestroy bool `env:"NO_DESTROY" sect:"cluster" default:"false"`
+	// DestroyClusterAfterTest set to false if you want OCM to clean up the cluster itself after the test completes.
+	DestroyClusterAfterTest bool `env:"DESTROY_CLUSTER" sect:"cluster" default:"true"`
 
 	// Kubeconfig is used to access a cluster.
 	Kubeconfig []byte `env:"TEST_KUBECONFIG" sect:"cluster"`
 
+	// ClusterExpiryInMinutes is how long before a cluster expires and is deleted by OSD
+	ClusterExpiryInMinutes int64 `env:"CLUSTER_EXPIRY_IN_MINUTES" sect:"cluster" default:"210"`
```
jeefy: The documentation for ExpiryInHours says 5 hours, yet this would be 3.5 hours. Which is correct?

meowfaceman: I meant for this to be 3.5 hours, so I'll update the docs. I'm a little worried about quota issues and figure we can configure it differently if we'd like.

PR #152, opened 2019-11-11
SDCICD-50: Create example for test workloads

jeefy:
This PR provides an example for testing workloads on new and post-upgrade clusters. Using the example supplied from git.k8s.io/examples (simple guestbook), we submit the workload and validate that it's functioning (in this case, simply returning a webpage).

If the test suite is executed again in the same run, we track what "workloads" have been installed and only validate that the workload is still running again.

Setting this as WIP for additional testing.


.golang-ci.yml
```diff
@@ -5,15 +5,16 @@ run:
   deadline: 30m
   issues-exit-code: 1
   tests: true
-  modules-download-mode: readonly
```
jeefy: Changes in this file were required to skip certain strict golang checks.

Dockerfile
```diff
@@ -6,17 +6,19 @@ WORKDIR ${PKG}
...
-ADD . ${PKG}
+COPY . ${PKG}
 
 RUN make check
 
 RUN make out/osde2e
 
 FROM gcr.io/distroless/base
+
+COPY artifacts artifacts
```
jeefy: We now have YAML artifacts we need copied into the container. Typically it's better to use COPY over ADD, so I updated the above lines as well. Functionally they're the same.

Makefile
```diff
@@ -12,7 +12,7 @@ check: cmd/osde2e-docs
 	go run $(PKG)/$< --check
 	CGO_ENABLED=0 go test -v ./cmd/... ./pkg/... 
 	go get -u github.com/golangci/golangci-lint/cmd/golangci-lint && \
-    golangci-lint run ./... 
+    golangci-lint run -c .golang-ci.yml ./...
```
jeefy: Actually use the config (I believe it was not auto-detecting the config previously)

artifacts/workloads/tests/badpod.yaml
```diff
@@ -0,0 +1,2 @@
+apiVersion: v1
+kind: invalidThing
```
jeefy: invalid thing is invalid.

pkg/helper/workloads.go
```diff
@@ -0,0 +1,152 @@
...
+		// If this times out, there's something wrong.
+		wait.PollImmediate(2*time.Second, 1*time.Minute, func() (bool, error) {
+			if _, err := kube.CoreV1().Namespaces().Get(ns, metav1.GetOptions{}); err != nil {
+				return false, nil
+			}
+			return true, nil
+		})
+	}
+
+	switch obj.GetObjectKind().GroupVersionKind().Kind {
```
jeefy: This is the only thing I don't like.

Currently we only support the below-defined K8s objects. The way to do dynamic objects is hacky and frankly more pain than it's worth. 

If additional workloads are added that don't fall under these objects we will need to add support here as well.

jeremyeder: :+1:

jeefy:
Local tests came back looking great, ready for review. 

/assign @meowfaceman

jeefy:
Conflicts resolved @meowfaceman PTAL

pkg/helper/workloads.go
```diff
@@ -0,0 +1,152 @@
...
+// ApplyYamlInFolder reads a folder and attempts to create objects in K8s with the yaml
+func ApplyYamlInFolder(folder, namespace string, kube kubernetes.Interface) ([]runtime.Object, error) {
+	var (
+		objects []runtime.Object
+		obj     runtime.Object
+		files   []string
+		err     error
+	)
+	pwd, _ := os.Getwd()
+	err = filepath.Walk(pwd+folder, func(path string, info os.FileInfo, err error) error {
```
meowfaceman: Let's do filepath.Join here to avoid concatenation issues.

pkg/helper/workloads.go
```diff
@@ -0,0 +1,152 @@
...
+		if strings.Contains(path, ".yaml") || strings.Contains(path, ".yml") {
```
meowfaceman: nits: Use info.Name() instead of path, and use strings.HasSuffix() instead of strings.Contains(). Unlikely to cause issues, but reduces the surface area for complication.

pkg/helper/workloads.go
meowfaceman: Let's add a TODO here to convey that for future implementers.

pkg/helper/workloads_test.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+func TestCreateWorkload(t *testing.T) {
+	pwd, _ := os.Getwd()
+
+	var tests = []struct {
+		description string
+		file        string
+	}{
+		{"pod", pwd + "/../../artifacts/workloads/tests/pod.yaml"},
```
meowfaceman: Should update these as well with the filepath.Join change.

PR #154, opened 2019-11-14
CICD-6: Run tests pre and post upgrade

jeefy:
We should validate that a cluster is valid and healthy prior to running an upgrade. 

There was some massaging to be done because of the way the test suite set itself up (if kubeconfig was set it assumed it was a filename not a valid kubeconfig and how helpers would spin up projects)  but it's been tested locally without issue.

### DIFF
--- a/e2e.go
+++ b/e2e.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/openshift/osde2e/pkg/config"
 	"github.com/openshift/osde2e/pkg/osd"
+	"github.com/openshift/osde2e/pkg/upgrade"
 )
 
 // OSD is used to deploy and manage clusters.
@@ -60,8 +61,7 @@ func RunE2ETests(t *testing.T, cfg *config.Config) {
 	}
 
 	// setup reporter
-	err = os.Mkdir(cfg.ReportDir, os.ModePerm)
-	if err != nil {
+	if err = os.Mkdir(cfg.ReportDir, os.ModePerm); err != nil {
 		log.Printf("Could not create reporter directory: %v", err)
 	}
 	reportPath := path.Join(cfg.ReportDir, fmt.Sprintf("junit_%v.xml", cfg.Suffix))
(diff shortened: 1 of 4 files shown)


common/e2e.go
```diff
@@ -60,8 +61,7 @@ func RunE2ETests(t *testing.T, cfg *config.Config) {
 	}
 
 	// setup reporter
-	err = os.Mkdir(cfg.ReportDir, os.ModePerm)
-	if err != nil {
+	if err = os.Mkdir(cfg.ReportDir, os.ModePerm); err != nil {
```
jeefy: I like the readability of this more and I've been changing those as I run into them. /shrug

common/e2e.go
```diff
@@ -70,5 +70,24 @@ func RunE2ETests(t *testing.T, cfg *config.Config) {
 	if !cfg.DryRun {
 		log.Println("Running e2e tests...")
 		ginkgo.RunSpecsWithDefaultAndCustomReporters(t, "OSD e2e suite", []ginkgo.Reporter{reporter})
+		// upgrade cluster if requested
```
jeefy: We have to shuffle around some logic (upgrade, destroy) to put it where we call the ginkgo tests. That way it's very clean/clear (not a dryRun? run tests. upgrade? upgrade. run tests. destroy cluster? destroy cluster.)

common/setup.go
```diff
@@ -106,6 +92,15 @@ func setupCluster(cfg *config.Config) (err error) {
 // useKubeconfig reads the path provided for a TEST_KUBECONFIG and uses it for testing.
 func useKubeconfig(cfg *config.Config) (err error) {
 	filename := string(cfg.Kubeconfig)
+
+	_, err = clientcmd.RESTConfigFromKubeConfig(cfg.Kubeconfig)
```
jeefy: The kubeconfig stuff mentioned above. This is checking "hey is this kubeconfig a valid kubeconfig? no? okay then try reading it as a file. yes? sweet return."

PR #157, opened 2019-11-15
SDCICD-91. Collect metadata and report it during a run.

meowfaceman:
osde2e will now collect metadata and serialize it to JSON when running
tests. These will then be reported in Spyglass when examining prow job
results.

### DIFF
--- a/pkg/metadata/metadata.go
+++ b/pkg/metadata/metadata.go
@@ -1,0 +1,42 @@
+package metadata
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"os"
+)
+
+// metadata houses the metadata that will be written to the report directory after
+type Metadata struct {
+	// Cluster information
+	ClusterID      string `json:"cluster-id"`
+	ClusterName    string `json:"cluster-name"`
+	ClusterVersion string `json:"cluster-version"`
+	Environment    string `json:"environment"`
+	UpgradeVersion string `json:"upgrade-version,omitempty"`
+
+	// Metrics
+	TimeToOCMReportingInstalled float64 `json:"time-to-ocm-reporting-installed,string"`
+	TimeToClusterReady          float64 `json:"time-to-cluster-ready,string"`
+}
+
+// Instance is the global metadata instance
+var Instance *Metadata
+
+func init() {
+	Instance = &Metadata{}
+}
+
+// WriteToJSON will marshall the metadata struct and write it into the given file.
+func (m *Metadata) WriteToJSON(outputFilename string) (err error) {
+	var data []byte
+	if data, err = json.Marshal(m); err != nil {
+		return err
+	}
+
+	if err = ioutil.WriteFile(outputFilename, data, os.FileMode(0644)); err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/pkg/metadata/metadata_test.go
+++ b/pkg/metadata/metadata_test.go
@@ -1,0 +1,96 @@
+package metadata
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"reflect"
+	"testing"
+)
+
+func TestMetadata(t *testing.T) {
+	tests := []struct {
+		name     string
+		m        *Metadata
+		expected map[string]interface{}
+	}{
+		{
+			name: "test all fields",
+			m: &Metadata{
+				ClusterID:                   "test-id",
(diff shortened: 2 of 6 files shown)


pkg/metadata/metadata.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"os"
+)
+
+// metadata houses the metadata that will be written to the report directory after
+type metadata struct {
+	ClusterID      string `json:"cluster-id"`
```
jeefy: Thoughts on making a clusterMetadata Struct? 

Should then output like metadata.cluster.id and the like.

pkg/metadata/metadata.go
```diff
@@ -0,0 +1,44 @@
...
+	ClusterName    string `json:"cluster-name"`
+	ClusterVersion string `json:"cluster-version"`
+	Environment    string `json:"environment"`
+	UpgradeVersion string `json:"upgrade-version,omitempty"`
+
+	// Metrics
```
jeefy: And maybe turn Metrics into map[string]string so we can do metadata.metrics.timetoclusteready

pkg/metadata/metadata.go
```diff
@@ -0,0 +1,44 @@
...
+// Instance is the global metadata instance
+var Instance *metadata
+
+func init() {
+	Instance = &metadata{}
+}
+
+// WriteToJSON will marshall the metadata struct and write it into the given file.
+func (m *metadata) WriteToJSON(outputFilename string) error {
+	data, err := json.Marshal(m)
```
jeefy: change this to be 
```
if data, err := json.Marshal(m); err != nil {
```

pkg/metadata/metadata.go
```diff
@@ -0,0 +1,44 @@
...
+
+	if err != nil {
+		return err
+	}
+
+	err = ioutil.WriteFile(outputFilename, data, os.FileMode(0644))
```
jeefy: Change to mirror above syntax

pkg/metadata/metadata_test.go
```diff
@@ -0,0 +1,99 @@
...
+		if err := writeAndTestMetadata(test.m, test.expected); err != nil {
+			t.Errorf("%s: error while testing metadata: %v", test.name, err)
+		}
+	}
+}
+
+func writeAndTestMetadata(m *metadata, expected map[string]interface{}) error {
+	tempDir, err := ioutil.TempDir("", "")
+
+	if err != nil {
```
jeefy: Sorry this will be redundant. Change syntax as above.

pkg/metadata/metadata_test.go
```diff
@@ -0,0 +1,99 @@
...
+		return err
+	}
+
+	outputFilename := filepath.Join(tempDir, "test.json")
+	err = m.WriteToJSON(outputFilename)
+
+	if err != nil {
```
jeefy: Sorry this will be redundant. Change syntax as above.

pkg/metadata/metadata_test.go
```diff
@@ -0,0 +1,99 @@
...
+		return err
+	}
+
+	data, err := ioutil.ReadFile(outputFilename)
+
+	if err != nil {
```
jeefy: Sorry this will be redundant. Change syntax as above.

pkg/metadata/metadata_test.go
```diff
@@ -0,0 +1,99 @@
...
+		return err
+	}
+
+	readData := map[string]interface{}{}
+	err = json.Unmarshal(data, &readData)
+
+	if err != nil {
```
jeefy: Sorry this will be redundant. Change syntax as above.

pkg/metadata/metadata.go
meowfaceman: Hrm. Not opposed to it. I'll do that.

meowfaceman: This seems to be not as easy as I expected, so I'm going to punt on this. It's doable, I'm just not sure of the value vs. how much workk to put into it.

pkg/metadata/metadata.go
meowfaceman: I actually intentionally resisted making this a map so that the naming of things had a little bit more immediate oversight. I'd prefer to keep it this way if that's alright with you.

jeefy: kk

PR #160, opened 2019-11-28
Add e2e test for certman

wanghaoran1988:


wanghaoran1988:
@clcollins Could you take a look?
/cc @c-e-brumm

test/operators/certman.go
```diff
@@ -0,0 +1,30 @@
...
+	. "github.com/onsi/gomega"
+	"github.com/openshift/osde2e/pkg/helper"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = ginkgo.Describe("[OSD] Certman Operator", func() {
+	h := helper.New()
+	secretName := h.ClusterName + "-" + "primary-cert-bundle-secret"
+	ginkgo.Context("certificate secret  should be applied when cluster installed", func() {
```
meowfaceman: nit: Double space between "secret" and "should."

c-e-brumm:
Code looks good +1 on that. If I understand the intention then we're checking a provisioned test cluster for the secrets and certs that we expect Hive to have passed it during creation. Just double checking that osde2e clusters are built by Hive. If that's accurate then this test looks good to confirm certman was working.

wanghaoran1988:
> Just double checking that osde2e clusters are built by Hive. If that's accurate then this test looks good to confirm certman was working.

Yes, the clusters are built by Hive.

PR #161, opened 2019-11-29
Add ginkgo.focus option

wanghaoran1988:


meowfaceman:
/approve

I have no problem with this, but I'm gonna defer to @jeefy on this as he has some ideas relating to this functionality w/r/t configurable osde2e jobs via yaml configuration. I think this will be good in the short term though.

PR #164, opened 2019-12-03
SDCICD-56: Migrate to go modules

jeefy:
This is the initial work needed to migrate our dependencies to go modules. I've tested this locally with success 🎉 but want more eyes 👀on this before we consider migrating.

/hold
/assign @meowfaceman


go.mod
```diff
@@ -0,0 +1,37 @@
+module github.com/openshift/osde2e
+
+go 1.12
```
meowfaceman: I'm guessing this would need to change if we moved to the most recent go version.

jeremyeder:
Just a sec - the base image here is a debian image from dockerhub?  We should be using an RH base image.  For example registry.svc.ci.openshift.org/ocp/builder:golang-1.12 ?

jeefy:
/hold cancel

ready for lgtm

jeefy:
Oooh one sec, just saw that comment @jeremyeder

jeefy:
@jeremyeder Thoughts on using the alpine tag? We're building then copying artifacts into `distroless` anyhow.

jeremyeder:
Use RH container images please. (openshift policy)

jeefy:
@jeremyeder @meowfaceman PTAL

Dockerfile
```diff
@@ -1,24 +1,24 @@
-FROM golang:1.11.9
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.13
+
 ENV PKG=/go/src/github.com/openshift/osde2e/
+ENV GOMODULE111=on
```
meowfaceman: We don't need this now, right?

jeefy: You saw nothing

image

jeefy:
Get credentials to the internal registry and `make build-image` :)

PR #168, opened 2019-12-04
SDCICD-101. Get cluster name for existing clusters.

meowfaceman:
If a cluster ID is provided, the cluster name was not being properly
set. It will now be retrieved from OCM if the cluster name has not
otherwise been provided. Additionally, the cluster name is now recorded
in the custom prow metadata.

### DIFF
--- a/common/setup.go
+++ b/common/setup.go
@@ -78,8 +78,23 @@ func setupCluster(cfg *config.Config) (err error) {
 		}
 	} else {
 		log.Printf("CLUSTER_ID of '%s' was provided, skipping cluster creation and using it instead", cfg.ClusterID)
+
+		if cfg.ClusterName == "" {
+			cluster, err := OSD.GetCluster(cfg.ClusterID)
+			if err != nil {
+				return fmt.Errorf("could not retrieve cluster information from OCM: %v", err)
+			}
+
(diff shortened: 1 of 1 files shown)


common/setup.go
```diff
@@ -78,8 +78,19 @@ func setupCluster(cfg *config.Config) (err error) {
 		}
 	} else {
 		log.Printf("CLUSTER_ID of '%s' was provided, skipping cluster creation and using it instead", cfg.ClusterID)
+
+		if cfg.ClusterName == "" {
+			cluster, err := OSD.GetCluster(cfg.ClusterID)
+			if err != nil {
+				return fmt.Errorf("could not retrieve cluster name: %v", err)
```
jeefy: Technically this message isn't correct. If this fails it means that it couldn't retrieve data on the cluster as a whole, not the Name.

common/setup.go
```diff
@@ -78,8 +78,19 @@ func setupCluster(cfg *config.Config) (err error) {
...
+			}
+
+			cfg.ClusterName = cluster.Name()
```
jeefy: Related to above, it might make sense to check for a zero-length string returned from `cluster.Name()` and return an error in that case.

PR #171, opened 2019-12-05
Add tests for osd prune jobs

wanghaoran1988:


test/operators/prunejob.go
```diff
@@ -0,0 +1,94 @@
...
+				job := createJobFromCronJob(cjob)
+				job, err = h.Kube().BatchV1().Jobs(namespace).Create(job)
+				defer func() {
+					err = h.Kube().BatchV1().Jobs(namespace).Delete(job.Name, &metav1.DeleteOptions{})
+					Expect(err).NotTo(HaveOccurred())
+				}()
+				Expect(err).NotTo(HaveOccurred())
+				err = waitJobComplete(h, namespace, job.Name)
+				Expect(err).NotTo(HaveOccurred())
+			})
```
jeefy: Can you also set a timeout for the individual test? See https://github.com/openshift/osde2e/blob/master/test/operators/dedicatedadmin.go#L75 for an example.

cblecker:
My only concern with this aw written would be that there wouldn't be any builds/images to prune on a fresh CI cluster.

jeefy:
So this is an interesting distinction that I want to point out.

Most of the operator tests, as they are, don't test the *functionality* of
the Operator. It just tests the health and validity of its installation in
the cluster. The more we can test, the more we should, but I'd opt to get
wider coverage before we get deeper coverage.

100% agree that tests should eventually validate the behavior of the
Operator though.

jeremyeder:
@jeefy across the existing operator tests, can we create a reference doc like 'phase 1, test these 5 things'...'phase 2, test these other, deeper things'?

jeefy:
Absolutely. I'll card that and try and get that out soon.

cblecker:
> The more we can test, the more we should, but I'd opt to get wider coverage before we get deeper coverage.

I agree with this, as long as we record somewhere that this doesn't really deeply test that these work. Just that a CLI command exits zero. I don't want us to have a false sense of security that "oh, pruning is being tested" when it really isn't.

jeefy:
image

wanghaoran1988:
QE team already have a lot tests that testing the `prune` feature itself, I don't think we should test the features itself, making sure the job is running successfully should be enought in my mind. 
or maybe we need add a simple test that making sure we are using this feature correctly ?
Anyway, let's get this in first, making sure the jobs are running well.

test/operators/prunejob.go
wanghaoran1988: @jeefy Updated

jeefy: Thanks!

jeremyeder:
> QE team already have a lot tests that testing the `prune` feature itself, I don't think we should test the features itself, making sure the job is running successfully should be enought in my mind.
> or maybe we need add a simple test that making sure we are using this feature correctly ?
> Anyway, let's get this in first, making sure the jobs are running well.

Then those tests should be merged here, since the purpose of this repository is to actively gate our ability to release. @yufchang @yasun1

yasun1:
These 'prune' jobs are OCP features which are owned by OCP devexp team. I agree with that before OSD release, we need make sure the functions are worked as expected but not only the prune job is running. Currently, for 3.x OSD release, we always collaborate with all of the OCP QE teams to finish the big testing, we only focus on OSD specific functions.   
To test OCP features in a completed flow is good idea in osde2e. It helps us find the issues as soon as possible. But before agree with including such tests in osde2e,  we need reach a consensus that how much extent of the testing are acceptable for the OCP features.  In fact, it is impossible for us to includes all of the scenarios for an OCP feature.

yufchang:
> > QE team already have a lot tests that testing the `prune` feature itself, I don't think we should test the features itself, making sure the job is running successfully should be enought in my mind.
> > or maybe we need add a simple test that making sure we are using this feature correctly ?
> > Anyway, let's get this in first, making sure the jobs are running well.
> 
> Then those tests should be merged here, since the purpose of this repository is to actively gate our ability to release. @yufchang @yasun1

IMO，if we need test "Prune" feature in OSD testing or not depends on if any OSD specific difference or overridden exists compared with OCP.We can regard a full functional testing on OSD as two parts with different importance. One is OSD specific testing, the other is the same function testing with OCP.The latter part can use QE current automation testing.

PR #172, opened 2019-12-05
SDCICD-95. Preliminary metrics-sync.sh script.

meowfaceman:
This is the preliminary metrics-sync.sh script. This will be used to
verify that the job that runs this script is working properly.


jeefy:
@meowfaceman Generally lgtm. One thought, could we document either in-script or in-markdown the dependencies to run that script? (ie python version, etc)

jeefy:
You. You're awesome.

/lgtm

PR #174, opened 2019-12-06
SDCICD-103. Fix panic in certman test.

meowfaceman:
The certman test now verifies if the NamedCertificates array is not
empty before trying to grab a certificate out of it.

/cc @wanghaoran1988


wanghaoran1988:
thanks @meowfaceman

PR #176, opened 2019-12-10
SDCICD-83 - Support job configuration by optional YAML file

jeefy:
This PR restructures the Config package to have embedded structs. This was becoming necessary as the config options grew. In addition, when `osde2e` is run, it can now be supplied a `yaml` file with the config options.

By default, it looks for `./.osde2e.yaml` and if not found, continues on attempting to use environment variables and defaults. You can specify a file by passing `-e2e-config` when running `osde2e`

While a large PR, the majority of changes by and large involve changing various config options across the whole project. 

/assign @meowfaceman

### DIFF
--- a/.osde2e.yaml
+++ b/.osde2e.yaml
@@ -1,1 +1,0 @@
-dryRun: true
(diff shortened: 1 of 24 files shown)


jeefy:
@meowfaceman Should be good now. Take a look when you can.

PR #177, opened 2019-12-10
bump(*): Vendor dependencies

tnozicka:
This ensures offline builds and ensures the project requirements are met.

Also the project level golang builder is about to be switched to force vendoring mode.

/priority critical-urgent
/assign clcollins jeefy jeremyeder meowfaceman


meowfaceman:
Left a message in Slack, but do you have any context about this? I'm not averse to it but I'm not sure of what's driving this from a policy/business sense.

tnozicka:
The "policy/business sense" requires our product to be build offline. Being dependent on golang cache breaks the requirement.

I've sent and org wide mail, but to cite a bit here:

> There is also an unfortunate change that the Golang has forgotten to
address when enabling modules by default - vendor dir is ignored in
golang 1.13 unless using `-mod=vendor` explicitly. In other words, with
switch to golang-1.13 every component will ignore its vendor folder on
builds. As in some repos like origin we patch vendor content, this has
severe impact on ignoring patches / backports at scale, not only the
issue with offline mode. Golang is also tracking the fix, I think it was this one:
  https://github.com/golang/go/issues/33848
As for now Golang team is recommending the following:
  "Some people will want to routinely opt-in to vendoring by setting a
GOFLAGS=-mod=vendor environment variable." [1]
As the workflow for OpenShift org mandates using vendoring we have
decided to fix the golang-1.13 image by adding the environment variable
GOFLAGS='-mod=vendor' there. With the scale of the org (new repos are
growing exponentially with the split) fixing and hoping to maintain
individual repos is just too risky. And the whole org requires
vendoring anyways

The origin golang-1.13 builder image you use will be enforcing vendoring mode in https://github.com/openshift/release/pull/6337 

> If you are on a team owning these please take actions asap. The options
are:
 a) add vendor dir to your repo with `go mod vendor` and commit
 b) run `go mod vendor` in CI as a pre step before running build
 c) explicitly use -mod=mod in your build system to retain non-
vendoring module mode
 d) in CI config for your job, unset the global GOFLAGS variable
 e) use different builder image
Again, as the org is required to use vendoring, a) is the preferred
option, unless there are exceptional circumstances.

meowfaceman:
/lgtm
Okay. The reasoning here sounds good, so I'll merge this.

PR #183, opened 2019-12-11
SDCICD-87. Addon test harness support.

meowfaceman:
Addon test harnesses can now be supplied to osde2e. This will allow for
addon authors to test against OSD clusters. Results from addons will be
collected by osde2e for later analysis.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -27,6 +27,8 @@ type Config struct {
 
 	OCM OCMConfig `yaml:"ocm"`
 
+	Addon AddonConfig `yaml:"addon"`
+
 	// Name lets you name the current e2e job run
 	JobName string `json:"job_name" env:"JOB_NAME" sect:"tests" yaml:"jobName"`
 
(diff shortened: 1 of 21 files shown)


pkg/config/config.go
```diff
@@ -130,6 +138,9 @@ type TestConfig struct {
 	// OperatorSkip is a comma-delimited list of operator names to ignore health checks from. ex. "insights,telemetry"
 	OperatorSkip string `env:"OPERATOR_SKIP" sect:"tests" default:"insights" yaml:"ginkgoFocus"`
 
+	// SkipClusterHealthChecks skips the cluster health checks. Useful when developing against a running cluster.
+	SkipClusterHealthChecks bool `env:"SKIP_CLUSTER_HEALTH_CHECKS" sect:"tests" yaml:"skipClusterHealthChecks"`
```
jeefy: I think it's better to specify `default:"false"` here

pkg/runner/command.go
```diff
@@ -7,7 +7,7 @@ import (
 )
 
 // testCmd configures default Service Account as a kubeconfig, runs openshift-tests, and serves results over HTTP
-const testCmd = `
+const testCmd = `#!/bin/bash
```
jeefy: We should explicitly call this out and ensure documentation states the test container images require `bash` to be installed. 

Often times containers don't actually have `bash` installed, only sh. This isn't so much a critique on the PR but raising the need to document this once addon testing starts.

pkg/config/config.go
meowfaceman: Good catch -- had meant to do this.

pkg/runner/command.go
meowfaceman: Agreed here -- I'll add this to the doc.

meowfaceman:
/hold cancel
Saw some must-gather failures, but those aren't new. This looks good.

PR #186, opened 2019-12-13
SDCICD-87. Make sure runners run as root.

meowfaceman:
Runners will ensure that they will run as the root user for a container
so that the necessary directory /test-run-results can be created.


jeefy:
I'll unhold and lgtm this so we can get scale tests back but I'm going to make a card about improving this. 

/hold cancel
/lgtm

jeefy:
Just saw that message. Didn't update. :( 

/hold cancel

PR #187, opened 2019-12-16
SDCICD-108. Counter event support.

meowfaceman:
osde2e can now track counter events. This will allow for tracking the
number of successful/failed events, like cluster provisions, upgrades,
etc. This will be useful for alerting.


meowfaceman:
/hold
Holding while I wait on a test run.

PR #191, opened 2019-12-19
Update Options.md

NautiluX:
fix `DESTROY_CLUSTER` description


meowfaceman:
Could you update this in pkg/config/config.go as well? This change used to be automated but is not now -- we have an action item to improve this. Otherwise lgtm.

PR #194, opened 2019-12-20
SDCICD-112 - Update README and Makefile

jeefy:
Updates README to better reflect project trajectory. Also now allows yaml configs to be used through `make test`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,13 +41,13 @@ build:
 	CGO_ENABLED=0 go test ./suites/scale -v -c -o ./out/osde2e-scale
 
 test:
-	go test $(E2E_PKG) -test.v -ginkgo.skip="$(GINKGO_SKIP)" -ginkgo.focus="$(GINKGO_FOCUS)" -test.timeout 8h -e2e-config $(E2ECONFIG)
+	go test $(E2E_PKG) -test.v -ginkgo.skip="$(GINKGO_SKIP)" -ginkgo.focus="$(GINKGO_FOCUS)" -test.timeout 8h -e2e-config=$(E2ECONFIG)
 
 test-scale:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -41,13 +41,13 @@ build:
 	CGO_ENABLED=0 go test ./suites/scale -v -c -o ./out/osde2e-scale
 
 test:
-	go test $(E2E_PKG) -test.v -ginkgo.skip="$(GINKGO_SKIP)" -ginkgo.focus="$(GINKGO_FOCUS)" -test.timeout 8h 
+	go test $(E2E_PKG) -test.v -ginkgo.skip="$(GINKGO_SKIP)" -ginkgo.focus="$(GINKGO_FOCUS)" -test.timeout 8h -e2e-config $(E2ECONFIG)
```
meowfaceman: Just confirming: this behaves as expected when no E2ECONFIG is provided?

README.md
```diff
@@ -1,38 +1,148 @@
...
-The options available are found [here](./docs/Options.md).
+Set OCM_TOKEN environment variable:
+```
+$ export OCM_TOKEN=<token from step 1>
+```
 
-Common ones are:
-- [`NO_DESTROY`](./docs/Options.md#no_destroy): don't delete clusters after testing
-- [`CLUSTER_ID`](./docs/Options.md#cluster_id): test an existing cluster specified by ID
+Run tests:
```
meowfaceman: I'm wondering if we should excise this section and label the Test Examples section as "Running the Tests." Reading this again, the Example section basically covers this in much greater detail.

Makefile
jeefy: Yes and no. 

Yes, it will.

No it doesn't because I didn't put `=` between the flag and the string. New commit inc.

README.md
jeefy: Sounds good, I'll shuffle it around and give it another pass.

PR #198, opened 2020-01-08
SDCICD-123. Use cluster image sets for upgrades if specified.

meowfaceman:
Cluster image sets can now be used directly for upgrades if specified.


meowfaceman:
/hold
Waiting for a local test to run through on this, but otherwise this is close to where I want it.

jeefy:
So happy some of this logic is getting cleaned up

/lgtm

PR #199, opened 2020-01-09
e2e test for testing newest Hive with a set of older ClusterImageSets

yasun1:
https://issues.redhat.com/browse/SDCICD-9

Based on the current framework, the cluster will be created once before suite. The only method for me is separating the oldest and newest ClusterImageSets into two suites.  The versions used for testing are not the default version, that means three versions will be covered. 

Certainly, it can be enhanced to test all the versions, but need to think whether it is worth doing. In my opinion, the newest, the oldest, the middle one can cover the most cases.

### DIFF
--- a/common/version.go
+++ b/common/version.go
@@ -96,3 +96,25 @@ func setupUpgradeVersion(cfg *config.Config, osd *osd.OSD) (err error) {
 		cfg.Cluster.Version, cfg.Upgrade.ReleaseName, cfg.Upgrade.ReleaseStream)
 	return
 }
+
+// GetNewestAndOldestVersions returns a list which contains the newest and the oldest non-default versions.
+func GetNewestAndOldestVersions(cfg *config.Config) ([]string, error) {
+	OSD, err := osd.New(cfg.OCM.Token, cfg.OCM.Env, cfg.OCM.Debug)
+	if err != nil {
+		return nil, fmt.Errorf("could not setup OSD: %v", err)
+	}
+
+	versionList, err := OSD.EnabledNoDefaultVersionList()
+	if err != nil {
+		return nil, err
+	}
+
+	if len(versionList) < 2 {
+		return versionList, nil
+	}
+
+	return []string{
+		versionList[0],
+		versionList[len(versionList)-1],
+	}, nil
+}
--- a/common/version.go
+++ b/common/version.go
@@ -97,24 +97,12 @@ func setupUpgradeVersion(cfg *config.Config, osd *osd.OSD) (err error) {
 	return
 }
 
-// GetNewestAndOldestVersions returns a list which contains the newest and the oldest non-default versions.
-func GetNewestAndOldestVersions(cfg *config.Config) ([]string, error) {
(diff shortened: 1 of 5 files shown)


yasun1:
@meowfaceman  @jeefy  Please help review.

common/version.go
```diff
@@ -96,3 +96,25 @@ func setupUpgradeVersion(cfg *config.Config, osd *osd.OSD) (err error) {
...
+		return nil, err
+	}
+
+	if len(versionList) < 2 {
+		return versionList, nil
+	}
+
+	return []string{
+		versionList[0],
+		versionList[len(versionList)-1],
```
meowfaceman: I wonder if we should pick something in the middle here, e.g. len(versionList)/2) instead of -1. Or something like that -- @jeefy, @yasun1, what do you guys think?

yasun1: I can understand your concern. I think we have covered a kind of the middle one. The middle one is based on where the default version(which is covered by the regular testing) is:
1. The oldest < the default < the newest
2. The default < the oldest < the newest -- the oldest = the default + 1
3. The oldest < the newest < the default  -- the newest =  the default - 1

If the newest in above three scenarios has been covered the other cases, it is better for us to use len(versionList)/2) instead of -1 to cover as many ClusterImageSets as possible.

pkg/osd/versions.go
```diff
@@ -89,6 +89,30 @@ func (u *OSD) LatestVersion(major, minor int64) (string, error) {
 	return VersionPrefix + latest.Original(), nil
 }
 
+// EnabledNoDefaultVersionList returns a sorted list of the enabled but not default versions currently offered by OSD.
+func (u *OSD) EnabledNoDefaultVersionList() ([]string, error) {
+	semverVersions, err := u.getSemverList(-1, -1, "")
```
meowfaceman: This list is not sorted -- it'll need to be sorted by semver in order to work properly. Sorry, just realized this today. Other than that and the other comment, I think this is good to go.

common/version.go
meowfaceman: Hey @yasun1. Yeah, so you're right: the other three test scenarios are covered by other test cases, so I think len(versionList)/2 is the way to go. So basically we need the oldest version and the middle version.

pkg/osd/versions.go
yasun1: Thanks for point out this..  It is my mistake that I forget doing sorting.

yasun1: @meowfaceman  You guys have achieved Sorting at the last of the function `getSemverList`: `sort.Sort(semver.Collection(versions))`

meowfaceman: Whoops, good catch. Sorry about that!

yasun1:
@jeefy  May require you approve the new changes again, and then merged directly? Not familiar with the process, seems like the `/retest` calls all the members to do review again.

meowfaceman:
/retest
The merge should happen automatically once the verify passes, but it looks like the verify job got wedged for some reason. I'll try retesting again, and then if it's still wonky I'll merge it manually.

PR #210, opened 2020-01-30
SDCICD-104. Introduce state, refactor mutable config fields.

meowfaceman:
The concept of a global state across an osde2e run has been introduced.
As osde2e has expanded, we were introducing more and more mutable
objects. The idea here is that config should be immutable and that this
new concept, state, should be mutable depending on how the run of osde2e
progresses. To support this:

* The code has overall been refactored.
* Config and state objects are no longer passed around, electing to use
  the singletons instead.
* Config loading has been separated into a separate concept so that it
  can arbitrarily load config options into any object. This for the
  state to be seeded with initial values to do things like reuse existing
  cluster IDs.
* New special load values have been introduced so that, on load, the
  config object doesn't need any post processing. In particular:

  __TMP_DIR__: this will create a temporary directory and assign the
               location to the associated field.
  __RND_N__: this will create a random ASCII string of length N. For
             example, __RND_3__ will create a 3 character long random
             string.


pkg/load/load.go
```diff
@@ -0,0 +1,209 @@
...
+
+	if objectType := reflect.TypeOf(object); objectType.Kind() != reflect.Ptr {
+		return fmt.Errorf("the supplied object must be a pointer")
+	}
+
+	if err := loadDefaults(object); err != nil {
+		return fmt.Errorf("error loading config defaults: %v", err)
+	}
+
+	if err := loadFromEnv(object); err != nil {
```
jeefy: This should occur after the YAML config is loaded. We want to be able to override a YAML config using Environment Variables.

jeefy:
/lgtm
:taco: :taco: :taco:

PR #214, opened 2020-02-04
SDCICD-129: Update metadata json file upon every metadata instance change

jeefy:


pkg/metadata/metadata.go
```diff
@@ -40,6 +42,58 @@ func init() {
 	Instance = &Metadata{}
```
meowfaceman: As a thought, should we not export Instance here and elevate all of the setters to package level?

I'm +0 to this, just wondering your thoughts.

jeefy: I feel like the way it is now is more intuitive? I dunno. 

I get the idea I'm just +0 to it as well. And if we're both +0 I think maybe we let it slide?

PR #217, opened 2020-02-06
SDCICD-158. Increase health check window.

meowfaceman:
This effecitvely makes the health check window require 10 minutes of
good health instead of 5. This will hopefully allow for OSD operators to
have more time to set up properly and reduce flakes in our builds.


jeefy:
/lgtm
:pray:

PR #218, opened 2020-02-06
SDCICD-51: Support arbitrary metrics derived from log data

jeefy:


pkg/metadata/metadata_test.go
```diff
@@ -10,11 +10,31 @@ import (
 	"testing"
 )
 
+// generateExpected marshals/unmarshals a supplied metadata object.
+// Rather than hand-curating our own expected data, this functionally
+// achieves the same test result but it's cleaner.
+func generateExpected(m *Metadata) map[string]interface{} {
```
meowfaceman: The advantage of hard coding the test strings in here is that it forces the developer to be a little more intentional about the design of the JSON. With marshalling the JSON programmatically, it won't catch things like misnamed fields or missed fields.

I think I'd rather have the hard coding here rather than this function.

jeefy:
I disagree.

We're already hardcoding the Metadata we expect. On top of that, we're not
actually writing JSON. We're writing a hand-crafted unmarshalled interface{}

Functionally this is here to test whether writeToJSON, which besides
json.Marshal/Unmarshal, has logic to place and write a file correctly.
Therefore we shouldn't validate that json.Marshal/Unmarshal works
correctly, but that the data placed and read is what/where we expect.

Let the json library do its thing, let's just test what we care about:
Where the data is placed and it being what we expect.

meowfaceman:
/lgtm
Fair enough, good counterpoint!

PR #222, opened 2020-02-10
SDCICD-161. Fix cluster image set upgrade logic.

meowfaceman:
Cluster image set upgrades were not behaving as expected due to the
logic surrounding whether to run an upgrade and upgrade tests. It now
simply tests if the release name has been set.


jeefy:
/lgtm

:pray:

PR #223, opened 2020-02-10
SDCICD-162 - Automatically update to fast-x.y channel

jeefy:
If the upgrade image is set, force the upgrade to use that.

If not, if minor versions aren't the same, automatically change the channel to be fast-x.y of the upgrade path and sleep 15 (per the SRE script)

Then attempt the upgrade.

### DIFF
--- a/pkg/upgrade/upgrade.go
+++ b/pkg/upgrade/upgrade.go
@@ -7,6 +7,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Masterminds/semver"
+
 	configv1 "github.com/openshift/api/config/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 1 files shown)


pkg/upgrade/upgrade.go
```diff
@@ -83,22 +85,50 @@ func RunUpgrade(OSD *osd.OSD) error {
...
+			Force:   true, // Force if we have an image specified
+		}
+	} else {
+		upgradeVersion := strings.Replace(h.Upgrade.ReleaseName, "openshift-v", "", -1)
+		installVersion := strings.Replace(state.Instance.Cluster.Version, "openshift-v", "", -1)
+
+		upgradeVersionParsed := semver.MustParse(upgradeVersion)
+		installVersionParsed := semver.MustParse(installVersion)
+
+		if upgradeVersionParsed.Minor() > installVersionParsed.Minor() {
```
meowfaceman: We should add a TODO here to make sure we address major eventually. This will work for OSD 4.x, but will break for 5.x. If we put a TODO we'll at least have a breacrumb to fix the issue when it actually becomes a problem (years down the line).

PR #225, opened 2020-02-10
SDCICD-163. Make cluster image set tests work.

meowfaceman:
The cluster image set tests have not been running as expected. This will
make them function as intended.


jeefy:
/lgtm

Excellent excellent.

PR #226, opened 2020-02-11
Add test to verify volume expansion

sedroche:
Adding a test to verify volume expansion is set correctly as per: https://issues.redhat.com/browse/OSD-2390

During an upgrade [a bug](.https://bugzilla.redhat.com/show_bug.cgi?id=1751641) was introduced where `AllowVolumeExpansion` was unset. This test verifies this value is true on all StorageClasses.

### DIFF
--- a/test/verify/storage.go
+++ b/test/verify/storage.go
@@ -1,0 +1,25 @@
+package verify
+
+import (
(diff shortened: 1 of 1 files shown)


meowfaceman:
Thank you for the new test!

PR #230, opened 2020-02-12
Added checkSecrets assistant function

mrbarge:
Added a function checkSecrets to verify the presence of one or more secrets.

Tickets OSD-2455 and OSD-2456, concerning the creation of additional operator tests, will both be needing to verify the presence of secrets installed by the operators that are being tested. This commit adds a common function that can be used by future operator tests requiring this ability.

Refs:
https://issues.redhat.com/browse/OSD-2455
https://issues.redhat.com/browse/OSD-2456

### DIFF
--- a/test/operators/operators.go
+++ b/test/operators/operators.go
@@ -111,6 +111,20 @@ func checkRoleBindings(h *helper.H, namespace string, roleBindings []string) {
 		}, float64(config.Instance.Tests.PollingTimeout))
 	})
 }
+
+//nolint
+func checkSecrets(h *helper.H, namespace string, secrets []string) {
+	// Check that deployed secrets exist
+	ginkgo.Context("secrets", func() {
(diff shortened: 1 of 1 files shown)


jeefy:
golint is mad that the function is defined but not used. 

I'd be fine manually merging this in asassuming OSD-2455/2456 are coming soon but would also be okay pushing this PR back till one of those two is going to be PR'd in.

@meowfaceman Thoughts?

meowfaceman:
Let's add //nolint above the function for now. Looks like this will eliminate the linting for this function.

jeefy:
+1

mrbarge:
Thank you for the assistance, folks! Yes, the work on those two tickets should be coming soon - they're actively being worked and should arrive in the coming days. We wanted the function present so that both tickets could be worked indepdently, as they share the need for the behaviour. Appreciate the help, again!

PR #232, opened 2020-02-12
SDCICD-168: Add timetoupgraded metric and ensure timetoinstall is honored

jeefy:


jeefy:
/hold

I think I see something wrong.

jeefy:
/hold cancel

@meowfaceman take a peek now, I'm adding two new metrics specific to upgrades.

PR #233, opened 2020-02-13
SDCICD-171. Use composable configs instead of separate exeuction paths.

meowfaceman:
Rather than use separate exeuction paths, osde2e now has composable
configs using multiple baked in config objects.

Users can now specify CONFIGS=config1,config2,config3 and osde2e will
attempt to load those configs from the new /configs directory in the
repository. Users can still specify environment variables and can still
use a custom yaml config. The custom yaml config now uses the config
option setting -custom-config rather than -e2e-config.

Additionally, the config and state objects have been lightly refactored
and the middle/oldest cluster image set object has been migrated into
the version.go code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,7 @@
 .PHONY: check generate build-image push-image push-latest test
 
 PKG := github.com/openshift/osde2e
-ADDONS_PKG := $(PKG)/suites/addons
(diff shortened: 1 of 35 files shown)


jeefy:
I like the direction this is going.

/lgtm

PR #237, opened 2020-02-14
Add PagerDuty e2e test and fix dockerfile

karthikperu7:
Testing if the secrets exist in the target cluster after installation.
Also added `RUN go mod vendor` in the `dockerfile` to fix the image build which was failing.

### DIFF
--- a/test/operators/pagerduty.go
+++ b/test/operators/pagerduty.go
@@ -1,0 +1,16 @@
+package operators
+
+import (
+	"github.com/onsi/ginkgo"
+	"github.com/openshift/osde2e/pkg/helper"
+)
+
+var _ = ginkgo.Describe("[OSD] Pagerduty Operator", func() {
+	h := helper.New()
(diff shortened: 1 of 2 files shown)


test/operators/pagerduty.go
```diff
@@ -0,0 +1,16 @@
+package operators
+
+import (
+	"github.com/onsi/ginkgo"
+	"github.com/openshift/osde2e/pkg/helper"
+)
+
+var _ = ginkgo.Describe("[OSD] Pagerduty Operator", func() {
```
meowfaceman: Could you add the string `[Suite: operators]` to the beginning of this Ginkgo.describe string? This is a brand new thing. If you rebase on master you'll see examples in the other operator tests.

karthikperu7: Cool will do that now! thanks

karthikperu7:
@meowfaceman Added [Suite: operators]

PR #242, opened 2020-02-24
Test upgrades of SRE operators

sedroche:
https://issues.redhat.com/browse/OSD-2398

These tests verify that upgrading for configure-alertmanager-operator and rbac-permissions-operator are successful

### DIFF
--- a/pkg/e2e/operators/operators.go
+++ b/pkg/e2e/operators/operators.go
@@ -179,12 +179,8 @@ func ensureCSVIsInstalled(h *helper.H, csvName string, namespace string) error {
 }
 
 func getChannel() string {
-	if config.Instance.OCM.Env == "stage" {
+	if config.Instance.OCM.Env == "stage" || config.Instance.OCM.Env == "int" {
 		return "staging"
-	}
-
-	if config.Instance.OCM.Env == "int" {
-		return "int"
 	}
 
 	return "production"
--- a/pkg/e2e/operators/operators.go
+++ b/pkg/e2e/operators/operators.go
@@ -2,6 +2,8 @@ package operators
 
 import (
 	"fmt"
+	kerror "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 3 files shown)


pkg/e2e/operators/operators.go
```diff
@@ -125,6 +127,131 @@ func checkSecrets(h *helper.H, namespace string, secrets []string) {
...
+		return false, nil
+	});
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func getChannel() string {
```
sedroche: I'm not sure if this is the correct way to do this. The channel name is different on different environments.

jeefy: Looks good to me. 

Do you know if there are other uses of this function? It might make more sense to put this logic in something like pkg/common or pkg/osd depending on how widely used it would be.

If you don't know any off the top of your head we'll keep it here and address it later.

meowfaceman:
Let's remove the commented out code in dedicatedadmin.go and rename it to rbac.go while we're here.

Is this code expected to only run against upgraded clusters? If not, is it okay for it to run on fresh installs?

pkg/e2e/operators/operators.go
sedroche: This is specific to the CatalogSource resource so I think it's ok to leave this here.
fyi a small update -> https://github.com/openshift/osde2e/pull/242/commits/4ed4cf675da0c4ffec38db6342658f7f47eb9b0a

sedroche:
@meowfaceman This should be run on fresh installs.

As for the dedicated-admin to rbac change, I looked into this and its slightly more work than just uncommenting the current code. I think we should merge this change now and do a follow up for the dedicated-admin to rbac changes.

I can create a ticket and start on it straight away.

wdyt>

sedroche:
@meowfaceman https://issues.redhat.com/browse/OSD-2990

PR #243, opened 2020-02-24
SDCICD-159: Start generating metrics based on logs

jeefy:
Starting metric generation using five basic but frequently-occurring errors observed.

One note: These metrics do not need to be negative or error metrics: We can generate positive metrics, or whatever we want.

### DIFF
--- a/configs/log-metrics.yaml
+++ b/configs/log-metrics.yaml
@@ -1,0 +1,6 @@
+logMetrics:
+  cluster-mgmt-500: "CLUSTERS-MGMT-500"
+  access-token-500: "can't get access token: token response status is: 500 Internal Server Error"
+  eof: "EOF"
+  cluster-pending: "Cluster is not ready, current status 'pending'"
+  host-dns-lookup: "dial tcp: lookup api\\.ci-cluster.* no such host occurred"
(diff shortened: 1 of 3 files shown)


configs/log-metrics.yaml
```diff
@@ -0,0 +1,6 @@
+logMetrics:
+  cluster-mgmt-500: "CLUSTERS-MGMT-500"
+  access-token-500: "can't get access token: token response status is: 500 Internal Server Error"
+  eof: "EOF"
+  cluster-pending: "Cluster is not ready, current status 'pending'"
+  host-dns-lookup: "dial tcp: lookup api\\.ci-cluster.* no such host occurred"
```
jeefy: Escape your backslashes or it will break. :smile:

pkg/e2e/e2e.go
```diff
@@ -223,13 +223,15 @@ func runTestsInPhase(t *testing.T, phase string, description string) bool {
...
 	for _, file := range files {
+		//log.Printf("Parsing log metrics in %s", filepath.Join(cfg.ReportDir, file.Name()))
 		if logFileRegex.MatchString(file.Name()) {
 			data, err := ioutil.ReadFile(filepath.Join(cfg.ReportDir, file.Name()))
 			if err != nil {
 				t.Fatalf("error opening log file %s: %s", file.Name(), err.Error())
 			}
 			for name, matchRegex := range logMetricsRegexs {
-				matches := matchRegex.Find(data)
+				matches := matchRegex.FindAll(data, -1)
```
jeefy: This was a mistake (now fixed) on my part.

PR #245, opened 2020-02-25
SDCICD-182. Make sure the out directory is created before building.

meowfaceman:
The out directory is now created when building the osde2e binary.


jeefy:
/lgtm
🌮

PR #250, opened 2020-02-27
SDCICD-190. Move echo into processing block.

meowfaceman:
An echo statement was not moved with the rest of the logic when adding
in support for new test metrics files.


meowfaceman:
This is pretty minor and has been sitting for a week. I'm going to merge this manually.

PR #255, opened 2020-03-03
Hopefully fix pbench secrets.

meowfaceman:
The pbench secrets don't look like they're propagating to the scale
tests. This fix will hopefully fix the issue.

Note: I believe this will fix the credentials issue. I believe there is a problem in the openshift-scale repo related to https://github.com/openshift-scale/workloads/commit/69365243a849cdae4e7705da55f0f3cd561dfa83 that is contributing to the scale failures.


jeefy:
This makes sense. I wish we could properly scope the secrets to the specific tests but short term

/lgtm

PR #256, opened 2020-03-04
Update tests for dedicated-admin with RBAC operator

sedroche:
https://issues.redhat.com/browse/OSD-2990

Updating the dedicated admins test after the move to the RBAC operator.

The ClusteRoles/ClusterRoleBindings/RoleBindings names are now taken from the SubjectPermissions custom resource.


pkg/e2e/operators/rbac.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("Error getting permissions")
+	}
+
+	var roleBindings = []string{}
+	for _, p := range permissions {
+		perm := p.(map[string]interface{})
+		cpName, _, err := unstruct.NestedString(perm, "clusterRoleName")
+		if err != nil {
+			return nil, nil, nil, fmt.Errorf("Error getting permission clusterRoleName", cpName)
```
meowfaceman: Looks like this is the cause of the test failure.

sedroche: Thanks @meowfaceman!

Updated now :+1:

PR #257, opened 2020-03-04
SDCICD-173. Gate report script.

meowfaceman:
A gate report script has been added that will wrap gate report
production and S3 uploading into an executable script. This will make
job creation of the gate report easier.


jeremyeder:
lgtm

PR #259, opened 2020-03-06
SDCICD-173. Bugfixes for the gate-report.sh script.

meowfaceman:
The report name for the generated gate report was interpolated
incorrectly. Additionally, if a report that was not viable was
generated, it would fail the report generation, which was not intended.


meowfaceman:
Same deal as before -- due to the simplicity, the availability of reviews, and to get this in and get reports generated, I'm going to merge this.

PR #260, opened 2020-03-06
SDCICD-174. Introduce gate report analysis command and script.

meowfaceman:
A command has been introduced that will analyze gate reports and
pass/fail based on the values in the gate report.

### DIFF
--- a/cmd/osde2e/gate/gate_report.go
+++ b/cmd/osde2e/gate/gate_report.go
@@ -11,8 +11,8 @@ import (
 	"github.com/openshift/osde2e/pkg/gate"
 )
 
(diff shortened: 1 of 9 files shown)


jeefy:
distracted overall lgtm, only thing that stuck out was TestCommand -> Command -- if we prepend all commands with their intent this seems like an oddity even if it's the default behavior (if that makes sense)

Like the writer solution.

meowfaceman:
Honestly I did the TestCommand => Command due to a liner suggestion, which is that test.TestCommand "stutters," and it suggested to rename it to "test.Command" (since the package should imply the correct command. I don't feel strongly about this, but it made sense when reading it. Thoughts?

PR #263, opened 2020-03-06
SDCICD-194. Exit gracefully for middle/oldest imageset tests.

meowfaceman:
If a version cannot be found for the middle/oldest imageset test, we
should exit gracefully instead of failing.


jeefy:
/lgtm

Saw the failures on this last night.

PR #265, opened 2020-03-08
SDCICD-195. Fail if no versions can be found for gate report.

meowfaceman:
If, when generating a gate report, no versions can be found in
Prometheus, we should fail.


jeefy:
ezpz

/lgtm

PR #266, opened 2020-03-09
SDCICD-174. Scripts should always grab the most recent image.

meowfaceman:
The scripts will now always grab the most recent docker image before
running the docker run command.


jeefy:
/lgtm

Just noting another place than slack: Eventually we may want to pin a specific image tag for this, but for now it's using `latest`

PR #270, opened 2020-03-12
SDCICD-152: Pass the build number and cluster id to prom metadata

jeefy:
This will add the Cluster ID and the Build Number to the prometheus file generated at the end of a run. It also updates the docs and is covered by tests.

Validated it locally, subset of output:

```
cicd_metadata{cluster_id="1brhm6dg6dq3u76sa7amsa0m3bbte7iq",environment="prod",install_version="openshift-v4.3.0",job_id="",metadata_name="log-metrics.host-dns-lookup",phase="",upgrade_version=""} 0
cicd_metadata{cluster_id="1brhm6dg6dq3u76sa7amsa0m3bbte7iq",environment="prod",install_version="openshift-v4.3.0",job_id="",metadata_name="time-to-cluster-ready",phase="",upgrade_version=""} 1260.291193986
cicd_metadata{cluster_id="1brhm6dg6dq3u76sa7amsa0m3bbte7iq",environment="prod",install_version="openshift-v4.3.0",job_id="",metadata_name="time-to-ocm-reporting-installed",phase="",upgrade_version=""} 1950.130662779
cicd_metadata{cluster_id="1brhm6dg6dq3u76sa7amsa0m3bbte7iq",environment="prod",install_version="openshift-v4.3.0",job_id="",metadata_name="time-to-upgraded-cluster",phase="",upgrade_version=""} 0
cicd_metadata{cluster_id="1brhm6dg6dq3u76sa7amsa0m3bbte7iq",environment="prod",install_version="openshift-v4.3.0",job_id="",metadata_name="time-to-upgraded-cluster-ready",phase="",upgrade_version=""} 0
```

In this example there is no BUILD_NUMBER environment variable from when I ran it locally so it just sends it as a zero-length string. From a bunch of reading, this behavior seems fine/acceptable.


pkg/e2e/metrics.go
```diff
@@ -230,11 +230,7 @@ func (m *Metrics) jsonToPrometheusOutput(gatherer *prometheus.GaugeVec, phase st
 
 			// We're only concerned with tracking float values in Prometheus as they're the only thing we can measure
 			if floatValue, err := strconv.ParseFloat(stringValue, 64); err == nil {
-				if phase != "" {
```
jeefy: Since my reading yielded "empty label values are A-OKAY" we can simplify this to just always pass through phase.

meowfaceman: So I'm not sure about this -- basically, `cicd_metadata` isn't supposed to report the phase as it exists outside of any install/upgrade phase. It looks like phase is inserted above, which (IMO) is not desirable.

jeefy: I get what you're layin' down. 

Is the same true for the Addon Prometheus object?

meowfaceman: So actually no, at least IMO. The addon metadata could potentially be captured in the install or upgrade phases. We don't currently have any upgrade addon tests, but they could potentially exist.

PR #271, opened 2020-03-12
SDCICD-199. Move gate report AWS read/write functionality into osde2e.

meowfaceman:
osde2e will now directly interact with AWS for gate reports rather than
rely on scripting glue.


jeefy:
/lgtm
Holding to confirm w/ @meowfaceman 
/hold

meowfaceman:
/hold cancel
If @jeefy's fine with it, I'm fine with it.

PR #275, opened 2020-03-13
SDCICD-206. Filter upgrade version CISes if they're not in Cincinnati.

meowfaceman:
If cluster image sets are not present in Cincinnati, then they will now
be filtered out of electable upgrade versions if trying to use cluster
immge sets.

Additionally, upgrading clusters via Cincinnati has now been excised
since we no longer do that directly. We only use cluster image sets or
the release controller.

### DIFF
--- a/pkg/common/upgrade/releases.go
+++ b/pkg/common/upgrade/releases.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"net/http"
 	"strings"
+	"sync"
 
 	"github.com/Masterminds/semver"
 	"github.com/openshift/osde2e/pkg/common/config"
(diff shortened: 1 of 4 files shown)


meowfaceman:
```
➜  osde2e git:(master) UPGRADE_TO_CIS_IF_POSSIBLE=true osde2e test -configs stage,dry-run
...
2020/03/13 14:56:20 Selecting version 'openshift-v4.3.5' to be able to upgrade to 'openshift-v4.4.0-rc.1'
...
➜  osde2e git:(master) git checkout SDCICD-206
Switched to branch 'SDCICD-206'
➜  osde2e git:(SDCICD-206) UPGRADE_TO_CIS_IF_POSSIBLE=true osde2e test -configs stage,dry-run
...
2020/03/13 14:56:29 Selecting version 'openshift-v4.3.5' to be able to upgrade to 'openshift-v4.4.0-rc.0'
...
```

pkg/common/upgrade/releases.go
```diff
@@ -97,13 +85,70 @@ func LatestRelease(releaseStream string, useReleaseControllerForInt bool) (name,
...
+// IsVersionInCincinnati returns true if the version can be found in Cincinnati
+func IsVersionInCincinnati(version *semver.Version) (bool, error) {
+	channel := VersionToChannel(version)
+	cincinnatiVersions, err := cache.Get(channel)
+
+	if err != nil {
+		return false, fmt.Errorf("error loading Cincinnati data: %v", err)
+	}
+
+	for _, cincinnatiVersion := range cincinnatiVersions {
```
jeefy: Instead of a for loop here, we should make cincinatiVersions a map[string]bool and just check to see if the key exists.

meowfaceman:
Definitely agreed on tests. In practice the switch to the map[*semver.Version]bool isn't working as cleanly as I'd like.

jeefy:
kk. It's an optimization that we don't really need to worry about right now anyhow. 

/lgtm

PR #277, opened 2020-03-14
SDCICD-186: Test privileged container permissions

jeefy:
1. Expose the RandomStr generator to not be an internal-only method
2. Add Impersonate and AddProject methods to the Helper
3. Added the test itself

### DIFF
--- a/pkg/common/load/load.go
+++ b/pkg/common/load/load.go
@@ -179,7 +179,7 @@ func processValueFromString(f reflect.StructField, field reflect.Value, value st
 			}
 		} else if rndStringRegex.MatchString(value) {
 			if rndStringLen, err := strconv.Atoi(rndStringRegex.FindStringSubmatch(value)[1]); err == nil {
-				rndString := randomStr(rndStringLen)
+				rndString := RandomStr(rndStringLen)
 				log.Printf("Generated random string %s for field %s", rndString, f.Name)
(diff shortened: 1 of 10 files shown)


pkg/common/load/load.go
```diff
@@ -218,7 +218,7 @@ func processValueFromString(f reflect.StructField, field reflect.Value, value st
 	return nil
 }
 
-func randomStr(length int) (str string) {
+func RandomStr(length int) (str string) {
```
meowfaceman: Since we've got three instances of this function now, let's make a new folder (spitballing: `pkg/common/util`?) that has this so we're not duplicating code here.

pkg/e2e/osd/privileged.go
```diff
@@ -0,0 +1,82 @@
...
+                    Image: "registry.access.redhat.com/ubi8/ubi-minimal",
+                    SecurityContext: &v1.SecurityContext{
+                        Privileged: &privileged,
+                    },
+                },
+            },
+        },
+    }
+}
+var _ = ginkgo.Describe("[Suite: service-definition] [OSD] Privileged Containers", func() {
```
meowfaceman: Just a general thing, but can we add a bit more whitespace in this file separating logically separable parts of the code? Just makes it a little more readable to my eyes.

PR #280, opened 2020-03-15
Introduce shellcheck.

meowfaceman:
shellcheck has been introduced and the various scripts that osde2e uses
have been adjusted for this.


jeefy:
Huge +1
/lgtm

PR #285, opened 2020-03-17
Revert SDCICD-210 short term fix.

meowfaceman:
Looks like the short term fix for SDCICD-210 doesn't work well on prod.

### DIFF
--- a/pkg/e2e/operators/operators.go
+++ b/pkg/e2e/operators/operators.go
@@ -187,7 +187,7 @@ func getChannel() string {
 	return "production"
 }
 
-func checkUpgrade(h *helper.H, sub *operatorv1.Subscription, previousCSV string) {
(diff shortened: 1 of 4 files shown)


pkg/e2e/operators/operators.go
```diff
@@ -142,7 +141,7 @@ func getInstallPlan(h *helper.H, sub *operatorv1.Subscription) (*operatorv1.Inst
 		}
 
 		return false, nil
-	})
+	});
```
jeefy: Semicolon?

pkg/e2e/operators/operators.go
```diff
@@ -171,7 +170,7 @@ func ensureCSVIsInstalled(h *helper.H, csvName string, namespace string) error {
 			return true, nil
 		}
 		return false, nil
-	})
+	});
```
jeefy: Here too?

PR #286, opened 2020-03-17
SDCICD-210. Resubmit fixed operator upgrades.

meowfaceman:
Fixed operator upgrades have are now using more appropriate versions for
the current cluster versions.

Note: Basically just reverted the revert but updated versions. This has been verified against 4.3.0. Trying to verify against 4.4.0 but running into testing issues unrelated to this.

### DIFF
--- a/pkg/e2e/operators/configurealertmanager.go
+++ b/pkg/e2e/operators/configurealertmanager.go
@@ -3,8 +3,8 @@ package operators
 import (
 	"github.com/onsi/ginkgo"
 	"github.com/openshift/osde2e/pkg/common/helper"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


pkg/e2e/operators/configurealertmanager.go
```diff
@@ -42,16 +42,19 @@ var _ = ginkgo.Describe("[Suite: operators] [OSD] Configure AlertManager Operato
...
+			},
+			Spec: &operatorv1.SubscriptionSpec{
+				Package:                "configure-alertmanager-operator",
+				Channel:                getChannel(),
+				CatalogSourceNamespace: "openshift-monitoring",
+				CatalogSource:          "configure-alertmanager-operator-registry",
+			},
 		},
-	})
+		"configure-alertmanager-operator.v0.1.116-0b1cafc",
```
jeefy: Are we hard-coding an operator here now?

PR #287, opened 2020-03-18
SDCICD-210. Remove operator upgrade tests.

meowfaceman:
At the moment, it looks like Openshift 4.4 doesn't really have a good
way of returning the previous available version of an operator. We can
use a fixed version rather than try to detect a previously available
version, but it looks like these upgrades take >20 minutes per, and we
have a max limit of 4 hours per test run.

At the moment, we're going to just have to remove these tests until we
figure out a better approach that's compatible with Openshift 4.4.

cc @sedroche


meowfaceman:
/hold
While we discuss this.

PR #291, opened 2020-03-19
Set operator upgrades to automatic after first manual test refs OSD-3141

mrbarge:
This fixes an issue identified in OSD-3141 / SDCICD-210 whereby
once the static previousCSV version got two hops out of date
the test would fail. The fix sets the operator subscription
installplan to automatic once a first manual upgrade hop
has successfully been undertaken.

Additionally the previousCSVs for affected operators have been
updated to the most recent N-1 version.

### DIFF
--- a/pkg/e2e/operators/configurealertmanager.go
+++ b/pkg/e2e/operators/configurealertmanager.go
@@ -41,7 +41,7 @@ var _ = ginkgo.Describe("[Suite: operators] [OSD] Configure AlertManager Operato
 	checkRoleBindings(h, operatorNamespace, roleBindings)
 })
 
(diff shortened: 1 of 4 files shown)


sedroche:
Code lgtm

@mrbarge Can you just double check those versions are in production too?

mrbarge:
> Code lgtm
> 
> @mrbarge Can you just double check those versions are in production too?

Thanks for pointing that out @sedroche - I checked a prod CI cluster and for rbac and splunk operators, no, they weren't present. I've updated to reference prod's N-1 version for those two operators.

meowfaceman:
/lgtm
Thanks for figuring this out all. Glad to have these tests back!

PR #298, opened 2020-03-24
SDCICD-167. Test for critical cluster alerts.

meowfaceman:
A new test has been introduced to look for critical cluster alerts.
While here, I moved the /artifacts directory to /assets in line with the
suggested golang project structure at
https://github.com/golang-standards/project-layout.


pkg/e2e/state/alerts.go
```diff
@@ -0,0 +1,105 @@
...
+
+	alertsCmdTpl = template.Must(template.New("alerts-cmd").Parse(string(data)))
+}
+
+var _ = ginkgo.Describe("[Suite: e2e] Cluster state", func() {
+	defer ginkgo.GinkgoRecover()
+	h := helper.New()
+
+	alertsTimeoutInSeconds := 900
+	ginkgo.FIt("should have no alerts", func() {
```
jeefy: This shouldn't be focused.

pkg/e2e/state/alerts.go
```diff
@@ -0,0 +1,105 @@
...
+		for _, result := range queryJSON.Data.Results {
+			if result.Metric.Severity == "critical" {
+				foundCritical = true
+			}
+
+			log.Infof("Active alert: %s, Severity: %s", result.Metric.AlertName, result.Metric.Severity)
+		}
+		Expect(foundCritical).To(BeFalse(), "found a critical alert")
+
+	}, float64(alertsTimeoutInSeconds+30))
```
jeefy: Why add 30 here?

pkg/e2e/state/alerts.go
meowfaceman: Whoops! Thanks.

pkg/e2e/state/alerts.go
meowfaceman: We actually do this everywhere -- I did this because we use the timeout as part of the runner, so that we'll timeout at X seconds during the run command and then add an additional 30 seconds for the test to report as possible before cleaning up. Make sense?

jeefy: Totes!

PR #300, opened 2020-03-25
OSD-2132 Add Redmine workload

mrbarge:
In support of OSD-2132 to improve customer app testing, this PR adds a new workload to build and host the Redmine web application via a Kubernetes deployment, using persistent storage for both the backend MySQL database service and the frontend web service, and secrets to manage the database credentials.

An initial validation method has been written to check that the service is responding. The deployment should complete within several minutes.

It has been successfully tested on staging and prod CI clusters so far.

If accepted as an additional workload, more sophisticated validation will be added as part of OSD-2132.

Thank you! :)

### DIFF
--- a/assets/workloads/e2e/redmine/mysql-deployment.yaml
+++ b/assets/workloads/e2e/redmine/mysql-deployment.yaml
@@ -1,0 +1,47 @@
+apiVersion: apps/v1 # for versions before 1.9.0 use apps/v1beta2
+kind: Deployment
+metadata:
+  name: redmine-mysql
+  labels:
+    app: redmine
+spec:
+  selector:
+    matchLabels:
+      app: redmine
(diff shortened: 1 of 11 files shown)


meowfaceman:
/approve
This is awesome. It looks like the failure in doTest is a linter issue -- did you mean for that to be a log.Printf instead of an assignment to err? Either way, once that's in I think this is looking good.

mrbarge:
> /approve
> This is awesome. It looks like the failure in doTest is a linter issue -- did you mean for that to be a log.Printf instead of an assignment to err? Either way, once that's in I think this is looking good.

Thanks @meowfaceman ! I've addressed that issue now and updated the branch accordingly. Cheers for the review.

jeefy:
/lgtm

Awesome work!

PR #304, opened 2020-03-26
SDCICD-214. Enable dynamic version selection for fresh installs.

meowfaceman:
Fresh installs now can be specified as a delta from the production
default value. As an example:

- Prod default: 4.3, delta 1 == 4.4.0-nightly.
- Prod default: 4.3, delta 2 == 4.5.0-nightly.
- Prod default: 4.4, delta 1 == 4.5.0-nightly.
- Prod default, 4.4, delta 2 == 4.6.0-nightly.

There needs to be a follow on to support upgrades in the same way, but
this will get us half of the way there.

### DIFF
--- a/pkg/common/osd/versions.go
+++ b/pkg/common/osd/versions.go
@@ -6,9 +6,11 @@ import (
 	"log"
 	"sort"
 	"strings"
+	"sync"
 
 	"github.com/Masterminds/semver"
 	v1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/openshift/osde2e/pkg/common/config"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


pkg/common/osd/versions.go
```diff
@@ -149,6 +188,112 @@ func (u *OSD) OldestVersion() (string, bool, error) {
 	return versionList[0], true, nil
 }
 
+// NextReleaseAfterProdDefault will return the latest version of a release after the default in production.
+// The integer "releasesFromProdDefault" is the "distance" from the current prod default to get the latest version for.
+// With a prod default of 4.3.2, a releaseFromProdDefault value of 1 would produce 4.4.0-0.nightly... on integration.
+// With a prod default of 4.3.2, a releaseFromProdDefault value of 1 would produce 4.5.0-0.nightly... on integration.
```
jeefy: Shouldn't this be a value of 2?

meowfaceman: Good catch. Bad copy paste. Fixed!

PR #311, opened 2020-03-30
CICD-197: Run must gather and collect state info after tests have run

jeefy:
The OCM queries should be tweaked once we know what more info we want to grab.


pkg/e2e/setup.go
```diff
@@ -71,11 +76,88 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
...
+
+		log.Print("Running Must Gather...")
+		mustGatherTimeoutInSeconds := 900
+		h.SetServiceAccount("system:serviceaccount:%s:cluster-admin")
+		r := h.Runner(fmt.Sprintf("oc adm must-gather --dest-dir=%v", runner.DefaultRunner.OutputDir))
+		r.Name = "must-gather"
+		r.Tarball = true
+		stopCh := make(chan struct{})
+		err := r.Run(mustGatherTimeoutInSeconds, stopCh)
+		Expect(err).NotTo(HaveOccurred())
```
meowfaceman: Rather than use Expect(), let's just do manual err checks and that way Id on't think we'll need to use ginkgo.GinkgoRecover() here at all. Should also make the error messages clearer/less alarming.

pkg/e2e/setup.go
```diff
@@ -71,11 +76,88 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
...
+	// Run Cluster State
+	func() {
+		defer ginkgo.GinkgoRecover()
+
+		log.Print("Gathering Cluster State...")
+		clusterState := h.GetClusterState()
+		stateResults := make(map[string][]byte, len(clusterState))
+		for resource, list := range clusterState {
+			data, err := json.MarshalIndent(list, "", "    ")
+			Expect(err).NotTo(HaveOccurred())
```
meowfaceman: Ditto.

pkg/e2e/setup.go
```diff
@@ -71,11 +76,88 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
...
+			filename := fmt.Sprintf("%s-%s-%s.json.gzip", resource.Group, resource.Version, resource.Resource)
+			stateResults[filename] = gbuf.Bytes()
+		}
+
+		// write results to disk
+		h.WriteResults(stateResults)
+	}()
+
+	// Get state from OCM
+	func() {
```
meowfaceman: I don't know that this is necessary since we're not using ginkgo.GinkgoRecover() in here.

pkg/e2e/setup.go
```diff
@@ -71,11 +76,88 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
 	// only needs to run once
 })
 
-// Collect logs after each test
-var _ = ginkgo.AfterSuite(func() {
+// After all tests run, clean up within the Ginkgo context
+var _ = ginkgo.SynchronizedAfterSuite(func() {
```
meowfaceman: Does this run even if BeforeInstall fails?

pkg/common/helper/helper.go
```diff
@@ -37,6 +37,17 @@ func New() *H {
 	}
 
 	ginkgo.BeforeEach(h.Setup)
+
+	return h
+}
+
+func NewOutsideGinkgo() *H {
+	h := &H{
```
meowfaceman: Let's extract this logic into a helper function and have New() and NewOutsideGinkgo() both call it. It's trivial, I know, but will reduce the amount of repetition here.

pkg/e2e/setup.go
jeefy: That makes sense.

PR #316, opened 2020-04-01
SDCICD-235. Calculate pass rate as metadata metric.

meowfaceman:
The pass rate for a job is now calculated as a metadata metric.


jeefy:
/lgtm

Thanks for cleaning up those debug log-prints

PR #320, opened 2020-04-02
Disable operator upgrade tests.

meowfaceman:
The operator upgrade tests are flaking consistently on int and stage,
which makes our full cluster upgrade tests fail to run. We need to
disable these until we can get them more stable.

We will re-enable these once we can introduce test tagging to indicate
that these are known to be flaky.

cc @sedroche


sedroche:
:+1:

PR #321, opened 2020-04-02
SDCICD-201: Test daemonset creation in OSD with various nodelabels

jeefy:
This test creates a daemonset using:
1) No nodeLabel
2) nodeLabel for worker
3) nodeLabel for infra

The criteria in the service definition is now "Create a DaemonSet using the worker NodeLabel" but technically others can be created. I'm testing for all three, and if OSD eventually restricts creation, we can tweak these accordingly.

### DIFF
--- a/pkg/e2e/osd/daemonsets.go
+++ b/pkg/e2e/osd/daemonsets.go
@@ -1,0 +1,84 @@
+package osd
+
+import (
+	"github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/openshift/osde2e/pkg/common/config"
+	"github.com/openshift/osde2e/pkg/common/helper"
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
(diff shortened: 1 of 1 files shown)


pkg/e2e/osd/daemonsets.go
```diff
@@ -0,0 +1,84 @@
...
+				},
+			},
+		},
+	}
+
+	return ds
+}
+
+var _ = ginkgo.Describe("[Suite: service-definition] [OSD] DaemonSets", func() {
+	ginkgo.Context("DaemonSets are not allowed", func() {
```
meowfaceman, line 47: Does this context make sense now? I'm assuming, since this is testing for positive existence of DaemonSets, that this isn't the intention any more.

jeremyeder:
I believe this should also ensure that customer-created daemonsets do not run on masters or infra nodes.

jeefy:
Currently there's no restriction on that unless we just found an issue with
the dedicated-admin role.

These 3 cases are all being run as a dedicated-admin-project and they pass.

jeremyeder:
> Currently there's no restriction on that

Incoming:
https://issues.redhat.com/browse/OSD-3217

jeefy:
Excellent! I was talking with Mike about this: With the test as-is it'll
start squawking once they have this restriction in place and we can then
update this test accordingly. :)

PR #323, opened 2020-04-03
SDCICD-239. Increase redmine timeout.

meowfaceman:
The redmine workload is frequently timing out, so this increases the
timeout of the redmine workload test to 15 minutes. Additionally, the
timeout waiting for the pods to come up has been increased from 5 to 10
minutes.


mrbarge:
/lgtm
Thanks for the pick-up!

PR #324, opened 2020-04-03
SDCICD-229: Attempt to retry failed tests 3 times

jeefy:
Sets the ginkgo flag to retry a failed test 3 times.

Per Ginkgo, if there's one success it deems the test a success.


meowfaceman:
So, interestingly I almost put up a PR for this myself. We can definitely do this, and this is the mechanism I'd use. My question: is it worthwhile to do this given the fact that OCP gating may not behave the way we're thinking? Thoughts?

jeefy:
I'd be interested in seeing if this cleans up any of our signal over the next week. If there's no real impact we can either leave it in or make it a config option we pass through (rather than hardcode in)

meowfaceman:
/lgtm
Fair enough, sounds like a plan. :-)

PR #330, opened 2020-04-07
Fix pkger again.

meowfaceman:
The /configs directory was not included in the previously generated
pkged.go file. This has been corrected. Additionally, a dry run check
has been added to the new helper cleanup method.


jeefy:
/lgtm

Have a taco :taco:

PR #336, opened 2020-04-09
SDCICD-248. Convert gate report into a weather report.

meowfaceman:
The gate report functionality has now been converted into a weather
report, which will be given to the Service Delivery CI/CD developers
daily covering the last 24 hours in a human readable format. The intent
of the weather report is to describe the failures occurring over the
last 24 hours.

The ability to output Markdown is now supported, and the output of a
weather report will be sorted in order of the job name.

The ability to post a report to a slack webhook is now supported.


assets/reports/markdown.template
```diff
@@ -0,0 +1,21 @@
+# osde2e weather report
+
+## Report Information
+This report was generated on {{.ReportDate}}
```
jeefy, line 4: I feel like this section is redundant. It could just go as the first line under Summary OR do something like `OSDe2e Weather Report {{.ReportDate}}`

assets/reports/markdown.template
```diff
@@ -0,0 +1,21 @@
+# osde2e weather report
```
jeefy, line 1: Nit: OSDe2e Weather Report

assets/reports/markdown.template
```diff
@@ -0,0 +1,21 @@
...
+
+## Report Information
+This report was generated on {{.ReportDate}}
+
+## Summary
+
+{{range .Jobs}}
+* [{{.Name}}](#{{.Name}}) (Viable: {{.Viable}}){{end}}
```
jeefy, line 9: Add a link to the job type in Prow.

assets/reports/markdown.template
```diff
@@ -0,0 +1,21 @@
...
+
+{{range .Jobs}}
+## {{.Name}}
```
jeefy, line 12: Add a link to the job type in Prow.

assets/reports/markdown.template
```diff
@@ -0,0 +1,21 @@
...
+
+## Report Information
+This report was generated on {{.ReportDate}}
+
+## Summary
+
+{{range .Jobs}}
+* [{{.Name}}](#{{.Name}}) (Viable: {{.Viable}}){{end}}
+
+{{range .Jobs}}
```
jeefy, line 11: I feel like this is redundant. Can we only list jobs that have failures?

assets/reports/markdown.template
```diff
@@ -0,0 +1,21 @@
...
+## {{.Name}}
+
+Viability: {{.Viable}}
```
jeefy, line 14: If we list only jobs that have failures, this becomes redundant and can be removed (because presumably if there are details it's because it's not viable)

pkg/common/report/weather_report.go
```diff
@@ -0,0 +1,135 @@
...
+	Versions     []string `json:"versions"`
+	FailingTests []string `json:"failingTests,omitempty"`
+}
+
+// Len is the number of jobs in the weather report.
+func (w WeatherReport) Len() int {
+	return len(w.Jobs)
+}
+
+// Less reports whether the element with index i should sort before the element with index j.
```
jeefy, line 50: Is this used anywhere?

pkg/common/report/weather_report.go
```diff
@@ -0,0 +1,135 @@
...
+func (w WeatherReport) Less(i, j int) bool {
+	return w.Jobs[i].Name < w.Jobs[j].Name
+}
+
+// Swap swaps the elements with indexes i and j.
```
jeefy, line 55: Same Q as above

pkg/common/report/weather_report.go
meowfaceman: Yeah, these are implicitly used by sort -- these are part of the sorting Interface.

assets/reports/markdown.template
meowfaceman: I'd personally rather list all jobs -- that way we get a sense of what's actually being run as opposed to needing to know which jobs are being run. Thoughts?

assets/reports/markdown.template
meowfaceman: Like this idea.

PR #340, opened 2020-04-14
SDCICD-257. Add in health check for certman certs.

meowfaceman:
Certman cert issuance is now part of osde2e's extended health checks
after cluster provisioning. Additionally, metadata has been added to
track this.


jeefy:
+1 to Healthcheck package refactor

/lgtm

Leaving to you to unhold after local tests. :)

PR #342, opened 2020-04-15
SDCICD-257. Perform all health checks each time.

meowfaceman:
All health checks are now performed each time when checking for cluster
readiness.


jeefy:
/lgtm

Short-term okay with this.

Long-term I want to see how long each iteration takes and tune our "20 Clean runs" accordingly. :)

PR #344, opened 2020-04-17
SDCICD-269: Run upgrade testing regardless of install phase results

jeefy:
An idealistic behavior that in retrospect was a mistake.

This will ensure we have some form of upgrade-signal regardless of the results of install-phase testing. 

/assign @meowfaceman


meowfaceman:
Forcing this merge due to the prow degredation.

PR #346, opened 2020-04-20
SDCICD-187: Add more documentation to our code (pkg/common edition)

jeefy:
First pass on everything up through pkg/common. :eyes:


pkg/common/runner/service.go
```diff
@@ -10,6 +10,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
+// createSer4vice returns a v1.Service pointing to a given v1.Pod object
```
meowfaceman, line 13: nit: Looks like there was a fat-finger on the 4 key in `createSer4vice`.

jeefy: new keyboard lol

PR #347, opened 2020-04-21
SDCICD-264. Remove type dependent failure check.

meowfaceman:
The failure check was inadvertently type dependent. This has been fixed.


meowfaceman:
@jeefy Same goes for this, though this is a small change.

jeefy:
This surprises me significantly that it doesn't take type into consideration.

/lgtm in any case.

PR #352, opened 2020-04-23
Fix MkdirAll so that it creates the directory and not the filename.

meowfaceman:
After discovering my very polluted cwd, I realized that I thought
`filename.Base` was the same as `filename.Dir`. Don't ask me how.
Anyway, it's fixed.


jeefy:
/lgtm

:taco:

PR #353, opened 2020-04-27
SDCICD-274. Add in configurable cluster provisioning layer.

meowfaceman:
A new configurable cluster provisioning layer has been added. This will
allow for osde2e to provision clusters from multiple paths should we
chooose to add cluster provisioning providers in the future.

One interesting note here is that, no matter what, right now everything
still goes through OCM, so we still have to poll OCM regardless of
whether we're using OCM directly to create clusters. This may change in
the future if we decide to add ARO to our testing capabilities, but for
now this is the simplest way forward.

It does create some somewhat annoying cross dependencies between the
provisioning package and the osd package, but I think these are worth it
since most of this OCM functionality will be shared.

An OCM connection cache has been added in since there are now two
objects that can create OCM connections. This will allow for reuse
across packages.

We'll need to take a follow on to extract the reliance on
config.Instance.OCM.Env from the code into something more general. In
order to do this, however, I'd like to remove the dependency that we
have on environment variables in our prow jobs, which will allow us to
change things like environment configs without prow job spec changes.


pkg/e2e/setup.go
```diff
@@ -99,44 +101,52 @@ func setupCluster() (err error) {
...
 		log.Printf("CLOUD_PROVIDER_REGION set to %s from OCM.", state.CloudProvider.Region)
+
+		log.Printf("Found addons: %s", strings.Join(cluster.Addons(), ","))
 	}
 
 	metadata.Instance.SetClusterName(state.Cluster.Name)
 	metadata.Instance.SetClusterID(state.Cluster.ID)
 
-	if err = OSD.WaitForClusterReady(); err != nil {
+	if err = cluster.WaitForClusterReady(provisioner, state.Cluster.ID); err != nil {
```
jeefy, line 145: This may sound weird but having a `cluster` method that we pass a provisioner to, feels somewhat clunky, especially, then, using a `provisioner` method to the grab the Kubeconfig

Let me know if that makes sense, it's surprisingly hard for me to articulate right now lol

pkg/common/cluster/clusterutil.go
```diff
@@ -0,0 +1,145 @@
...
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+const (
+	// errorWindow is the number of checks made to determine if a cluster has truly failed.
+	errorWindow = 5
+)
+
+// WaitForClusterReady blocks until the cluster is ready for testing.
+func WaitForClusterReady(provisioner spi.Provisioner, clusterID string) error {
```
jeefy, line 26: Should waiting for cluster readiness be part of the provisioner?

pkg/e2e/setup.go
meowfaceman: Yeah, good point -- this should probably just be a method that has the kubeconfig. +1. I'll change this.

This has the added benefit of *not* making an OCM call every time, too.

meowfaceman: Ugh, I forgot this also waits for the provisioner to report "ready." Let me think on this.

pkg/common/cluster/clusterutil.go
meowfaceman: IMO no, and the reason is that we need direct access to the cluster in order to assess whether a cluster is ready -- because there's a difference between OCM's notion of "ready" and our notion of "ready." I don't think the provisioner should be doing direct cluster access here, just letting us know what the provisioner APIs are reporting back.

If OCM did osd-verify, I'd be fine with it, but since that doesn't seem to be the case, I'd rather extract it out. Thoughts?

PR #355, opened 2020-04-27
SDCICD-274. Prevent cluster panic if err occurs in wait.

meowfaceman:
During the cluster health wait, if an error occurs when hitting the
provisioner API, an NPE can be hit because the err isn't being checked.

Additionally, when no upgrade candidates are found, the job doesn't
fail, but just fails gracefully (again).


jeefy:
oops

/lgtm

PR #366, opened 2020-05-01
SDCICD-273: Change the color by hand

jeefy:
Remember when we thought it was such a simple but awesome change? :\


meowfaceman:
/lgtm
Eh, so long as it works. Too bad though!

PR #368, opened 2020-05-04
SDCICD-283. Select versions prior to default version.

meowfaceman:
A test can be configured to run N versions prior to the default version.
Since production will soon support running multiple installation
versions on production, this will be necessary to ensure that we have
automated tests for all versions trying to upgrade to the current
default.

### DIFF
--- a/pkg/common/state/state.go
+++ b/pkg/common/state/state.go
@@ -46,6 +46,9 @@ type ClusterState struct {
 
 	// EnoughVersionsForOldestOrMiddleTest is true if there were enough versions for an older/middle test.
 	EnoughVersionsForOldestOrMiddleTest bool `default:"true"`
+
+	// PreviousVersionFromDefaultFound is true if a previous version from default was found.
+	PreviousVersionFromDefaultFound bool `default:"true"`
 }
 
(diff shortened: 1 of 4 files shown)


pkg/common/state/state.go
```diff
@@ -46,6 +46,9 @@ type ClusterState struct {
 
 	// EnoughVersionsForOldestOrMiddleTest is true if there were enough versions for an older/middle test.
 	EnoughVersionsForOldestOrMiddleTest bool `default:"true"`
+
+	// PreviousVersionFromDefaultFound is true if a previous version from default was found.
+	PreviousVersionFromDefaultFound bool `default:"true"`
```
jeefy, line 51: 1. Should this really default to true?
2. Should this have a default at all?

meowfaceman: So mentioned this in Slack, but just so the discussion is here:

Sort of depends on the perspective.

From the perspective of "does this work and do what it should," then yes it should be here.

From the perspective of "we need to refactor our version selection code entirely to make it more testable," no it should not.

How about we file a follow on to address this and get this code testable in the first place? Right now it's all too tightly coupled together to be tested well.

PR #373, opened 2020-05-07
Remove all references to pbench.

meowfaceman:
All references to pbench have been excised from osde2e.


jeefy:
/lgtm

/hold

@meowfaceman Feel free to unhold when you're sure/set.

PR #375, opened 2020-05-09
SDCICD-313. Move log metrics reset before the calculation.

meowfaceman:
It looks like the log metrics reset was inadvertently directly after
collecting the log metrics, which was causing all log metrics to be zero
all the time. This has been moved.


meowfaceman:
Because this is so minor, I'm going to push this through and see if the next jobs alert at all.

PR #380, opened 2020-05-12
SDCICD-315. Add in new cluster property tags.

meowfaceman:
New clusters will be tagged with two new tags when being launched:

* MadeByOsdE2E to distinguish clusters that are created by osde2e.
* OwnedBy with a value of the current system user to distinguish who has
  created the cluster, or at least provide some idea of who has done so.

### DIFF
--- a/pkg/common/providers/ocmprovider/cluster.go
+++ b/pkg/common/providers/ocmprovider/cluster.go
@@ -3,6 +3,7 @@ package ocmprovider
 import (
 	"fmt"
 	"log"
+	"os/user"
 	"time"
 
 	v1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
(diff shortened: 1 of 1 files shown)


pkg/common/providers/ocmprovider/cluster.go
```diff
@@ -14,6 +15,12 @@ import (
 const (
 	// DefaultFlavour is used when no specialized configuration exists.
 	DefaultFlavour = "osd-4"
+
+	// MadeByOsdE2E property to attach to clusters.
+	MadeByOsdE2E = "MadeByOsdE2E"
```
jeefy, line 20: This is pure opinion but I like "OSDe2e" better.

meowfaceman: Dang, I was actually trying to match your preference but I inverted it in my head, hahaha. Will change.

PR #381, opened 2020-05-12
SDCICD-316: Check for error message before generating diff

jeefy:
If we see a 404 (or really anything-not-200) status from the bucket browser, return an error rather than attempt to generate a diff.


pkg/debug/diff.go
```diff
@@ -28,6 +29,14 @@ func GenerateDiff(baseURL, phase, dependencies, jobName string, jobID int) error
 		return err
 	}
 
+	if resp.StatusCode == 400 {
```
meowfaceman, line 32: This returns a 400 instead of a 404?

jeefy: MY BAD

PR #384, opened 2020-05-13
SDCICD-315. Restore cluster properties with Prow specific logic.

meowfaceman:
It looks like, on prow, user.Current() fails. Now, if a BUILD_NUMBER is
detected, it's assumed we're running on prow. This should prevent
failures when running on prow.

### DIFF
--- a/pkg/common/providers/ocmprovider/cluster.go
+++ b/pkg/common/providers/ocmprovider/cluster.go
@@ -3,6 +3,7 @@ package ocmprovider
 import (
 	"fmt"
 	"log"
+	"os/user"
 	"time"
 
 	v1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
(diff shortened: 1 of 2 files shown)


pkg/common/providers/ocmprovider/cluster.go
```diff
@@ -30,6 +37,16 @@ func (o *OCMProvider) LaunchCluster() (string, error) {
 	// we happen to forget to do it:
 	expiration := time.Now().Add(time.Duration(cfg.Cluster.ExpiryInMinutes) * time.Minute).UTC() // UTC() to workaround SDA-1567.
 
+	user, err := user.Current()
+
+	var username string
+	if err != nil {
+		log.Printf("Unable to get current user. Assuming this is being run from prow.")
+		username = "prow"
```
jeefy, line 45: I would add an additional check to verify this is run in Prow by checking the Job Name and Job ID (mainly Job ID) since those are only populated by Prow. That way all cases are covered.

meowfaceman: Sounds good. Just submitted an update with this in place.

PR #386, opened 2020-05-14
SDCICD-310: Flesh out MockProvider

jeefy:
This PR fills out the stubs left in the MockProvider, and adds tests around them.

Of note, if you create a mock provider using `fail` as the Environment, several pieces of MockProvider functionality will fail (so we can test failure cases using MockProvider)

The mock Kubeconfig is a kubeconfig generated using `kind` locally so none of that is sensitive info.


pkg/common/providers/mock/mock.go
```diff
@@ -66,27 +78,108 @@ func (m *MockProvider) GetCluster(clusterID string) (*spi.Cluster, error) {
...
+	return len(addonIDs), nil
 }
 
 // Versions mocks a versions operation.
 func (m *MockProvider) Versions() (*spi.VersionList, error) {
-	return nil, fmt.Errorf("versions is currently unsupported by the mock provider")
+	if m.env == "fail" {
+		return nil, fmt.Errorf("Fake error returning version list")
+	}
+	versions := []*spi.Version{
```
meowfaceman, line 152: I feel like we should make a public helper function here so that tests can populate their own version lists -- this should be fine, as tests will likely be using the MockProvider directly.

pkg/common/providers/mock/mock.go
```diff
@@ -66,27 +78,108 @@ func (m *MockProvider) GetCluster(clusterID string) (*spi.Cluster, error) {
 
 // ClusterKubeconfig mocks a cluster kubeconfig operation.
 func (m *MockProvider) ClusterKubeconfig(clusterID string) ([]byte, error) {
-	return nil, fmt.Errorf("cluster kubeconfig is currently unsupported by the mock provider")
+	if clusterID == "fail" {
+		return nil, fmt.Errorf("failed to get versions: Some fake error")
+	}
+	// This kubeconfig is valid and can be parsed, but attmping to use it will cause failures :)
+	kubeconfig := `apiVersion: v1
```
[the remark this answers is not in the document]
jeefy: done!

pkg/common/providers/mock/mock.go
jeefy: I agree. We may want to stress version selection logic in novel ways so we should be able to set different use cases. I'll update it now.

PR #393, opened 2020-05-20
SDCICD-281. Introduction of the MOA provider.

meowfaceman:
The MOA provider has been introduced into osde2e. This will allow
clusters and tests to be run through MOA. This first pass will initially
allow for cluster provisioning directly through the MOA tool, and then
subsequently wrap every other call with the OCM provider. This will
hopefully change as we work on refactoring moactl itself.


meowfaceman:
cc @jharrington22, @vkareh 

I've got more work to do here, in particular figuring out how to get this thing to build with a private repo, but just so you see how the MOA provider looks post refactor.

PR #394, opened 2020-05-21
[WIP] Provider plugins.

meowfaceman:
Providers can now be supplied via plugin instead of needing to sit
inside the osde2e source tree.


meowfaceman:
FYI, this isn't really viable -- I'm leaving the review up just to peruse to see if any of these concepts could/can be useful, but I'll close this in a bit.

PR #395, opened 2020-05-21
SDCICD-317: Add CRC as a cluster provider

jeefy:
This adds `CRC` as a cluster provider for OSDe2e.

It makes the assumption that you have a valid/working CRC installation, and run a specific version. In the future this should be more configurable.

Additionally, this adds the "Type" function to the SPI interface that simply outputs the Provider "Type" which is hardcoded for each provider. Doing this allows us to write some provider-specific hooks (such as in the case of CRC, skipping certain checks before running e2e tests)


meowfaceman:
/lgtm

This is so cool.

PR #396, opened 2020-05-22
SDCICD-319. Install top level addons.

meowfaceman:
We can now support top level addons.


meowfaceman:
/hold

So, here's the rub on this one. Inspecting uhc-clusters-service. This builds properly and produces a valid JSON output. However, with top level addons supplied, I get:

```
<html><body><h1>502 Bad Gateway</h1>
The server returned an invalid or incomplete response.
</body></html>
```

So I can't really say if this works or if this is an OCM bug. But this is overall pretty straightforward, so I figured I'd push it here while I track down what's going on with that 502 error.

jeefy:
It's still held, but

/lgtm

meowfaceman:
/hold cancel

This works. Looks like I was populating the wrong fields in the addon installation build object before.

PR #400, opened 2020-05-26
SDCICD-250. Condense configs, switch to Viper.

meowfaceman:
The config object has largely been ripped out and replaced with Viper.
This allows us to disregard our own custom rolled config logic and
additionally allows providers to specify and load their own configs.


jeefy:
/lgtm

This cleans up a LOT of code and I am here for this.

meowfaceman:
/hold cancel

Got most of the way through an upgrade (ran into the openshift-samples) but, but otherwise everything seems to have gotten through, so I'm going to call this good.

jeefy:
/lgtm

:pray:

PR #401, opened 2020-05-27
SDCICD-281. Introduce utility functions to OCM provider.

meowfaceman:
The OCM provider has some utility functions that will be used by
external providers that also consume the OCM provider.


jeefy:
ezpz

/lgtm

PR #403, opened 2020-05-28
SDCICD-328. Add in plugin registration function.

meowfaceman:
This adds in a mechanism for registering plugins on startup. Since
golang will only call the init() functions of packages that are
imported, it's impossible for this to be totally hands off, e.g. drop in
a provider and it "just works." To support this, osde2e can now generate
the provider list automatically by calling make generate-providers. This
is additionally checked during make check to ensure that the providers
list is indeed updated as expected.

The golang generation isn't exactly ideal, but this is at least some way
of dropping in providers and having some scripting in place to support
it.


meowfaceman:
Not really sure how I feel about this. This'll get us closer, but just due to the way Go is, this isn't really doable in the way I was hoping. Let me know what you think about this.

PR #399, opened 2020-05-26
Port osde2e CLI to cobra

mrbarge:
**About**
This PR provides a port of osde2e's CLI interface to [cobra](https://github.com/spf13/cobra).

cobra is a subcommand-driven library similar to the existing [subcommand](https://github.com/google/subcommands) library used by osde2e.

**Rationale**
Its adoption brings a number of potential benefits:
- conformance with other OSD-related tools (kubectl, OCM and openshift-tests all use cobra) both for ease-of-development purposes and for producing a common usability interface.
- easy bash completion 
- dynamic tab-completion 
I thought these might be slightly useful in making the CLI interface a little more usable, and might be a useful base for allowing any future ideas or improvements.

**Issues**
I believe this PR is _almost_ entirely a drop-in replacement for the existing osde2e (meaning that existing means of running osde2e will work with the port), with one exception: the flags require a double-dash instead of a single-dash.

That is to say:
`osde2e test -configs=e2e-suite -custom-configs=blah`
is invalid, and should instead be:
`osde2e test --configs=e2e-suite --custom-configs=blah`

I have addressed areas of the Makefile where this is a problem.

**Bash completion**
The main feature added is bash completion and dynamic flag argument completion.

_Source the generated bash completion_
`. <(osde2e completion)`

_Tab complete configs, output formats, etc_
`osde2e test --configs=<tab><tab>`
`addon-suite    int   prod     two-releases-from-prod-default ...`

PR #404, opened 2020-05-28
Add viper config integration on CLI for test subcommand

mrbarge:
This PR integrates the work of #399 and #400 to allow the ability to specify some test-related configuration parameters via the CLI. 

Parameters selected were based on personal preferences of ones I change/use frequently, these can also be viewed via `osde2e test --help`:

- cluster ID
- OCM provider environment
- kube config path
- tests to focus
- tests to skip
- skip health checks
- destroy after test

As an example:
`osde2e test --configs informing-suite --environment int --cluster-id 1ddkj9cr9j908gdlb1q5v6ld4b7ina5m --focus="namespace validating" --skip-health-check`

The README has also been updated to include an example.


meowfaceman:
/lgtm

Thanks for catching that provider bug, too!

PR #408, opened 2020-05-29
SDCICD-191. Add cluster scaling, scale before mastervertical.

meowfaceman:
Clusters will now be scaled before mastervertical is run.

Additionally, the scale-ci-tooling namespace is cleaned up between scale
runs now. This would prevent running multiple types of scale tests in
the same test suite.


meowfaceman:
This isn't the end of this PR, but I was able to trigger what *could* be an actual legit issue based on this last night, so I'm pushing this up to get feedback and potentially get more data from our scale jobs.

jeefy:
/hold

Tossing a hold on here, but also wanted to mark that things look good so far.

PR #411, opened 2020-05-29
Start test for aws-efs-operator

2uasimojo:
This won't be ready for primetime until
[all](https://github.com/openshift/aws-efs-operator/pull/4)
[the](https://github.com/openshift/release/pull/9355)
[things](https://gitlab.cee.redhat.com/service/app-interface/merge_requests/4924)
and more have merged.

For now, run this as:

```shell
$ export OCM_TOKEN=...
$ export CLUSTER_ID=...
$ osde2e test --configs stage,skip-health-checks --custom-config configs/wip-suite.yaml
```

### DIFF
--- a/configs/wip-suite.yaml
+++ b/configs/wip-suite.yaml
@@ -1,0 +1,3 @@
+tests:
+  testsToRun:
+    - '[Suite: wip]'
--- a/pkg/e2e/operators/awsefs.go
+++ b/pkg/e2e/operators/awsefs.go
@@ -1,0 +1,42 @@
+package operators
+
+import (
+	"github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/openshift/osde2e/pkg/common/helper"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+const sharedVolumeCRDName = "sharedvolumes.aws-efs.managed.openshift.io"
+
+var _ = ginkgo.Describe("[Suite: wip] [OSD] AWS EFS Operator", func() {
+	h := helper.New()
+
+	ginkgo.Context("Sniff", func() {
+		ginkgo.FIt("SharedVolume CRD exists", func() {
+			crd, err := h.Dynamic().Resource(
+				schema.GroupVersionResource{
+					Group:    "apiextensions.k8s.io",
+					Resource: "customresourcedefinitions",
+					Version:  "v1",
+				},
+			).Get(sharedVolumeCRDName, metav1.GetOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			Expect(crd.GetName()).Should(Equal(sharedVolumeCRDName))
+		})
+
+		ginkgo.FIt("No SharedVolume resources exist", func() {
+			sharedVolumes, err := h.Dynamic().Resource(
+				schema.GroupVersionResource{
+					Group:    "aws-efs.managed.openshift.io",
+					Resource: "sharedvolumes",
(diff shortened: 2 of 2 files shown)


meowfaceman:
/approve
Sorry for the delay here. Could you run:

```
go get github.com/markbates/pkger/cmd/pkger
pkger
```

at the root of the repo and check in the updates to pkged.go? Otherwise this lgtm.

configs/wip-suite.yaml
```diff
@@ -0,0 +1,3 @@
+tests:
+  testsToRun:
+    - '[Suite: wip]'
```
jeefy, line 3: This should be set to `informing` not `wip`

meowfaceman: So, I took wip to be useful as I don't think this test is actually ready to run yet due to its dependencies. @2uasimojo , is that right?

jeefy: All of the dependencies have been merged. :) I checked before doing my review.

meowfaceman: Well snap. :-) Nevermind then!

2uasimojo: I was careful to leave myself an "out" with "and more" :P

See main thread for more...

2uasimojo:
> Looks awesome so far! Few small fixes requested.

Yeah, so this operator isn't deploying into OSD yet. IIUC [this](https://gitlab.cee.redhat.com/service/saas-osd-operators/merge_requests/290) would close that gap. BUT we're not super sure when (or even if) we're going to do that. Until/unless that happens, this suite is really only useful for being run locally against a cluster where the operator has been deployed manually. That's the/a reason I've got it set up in a WIP/focus suite -- so that it doesn't get run with everything else (even `informing`), and so I can run it alone when I run it locally.

That being the case, I'll understand if this is deemed unsuitable/unready for actual merging at this time.

> Another question: You're currently testing/ensuring that the CRD exists but no EFS volumes actually exist. Would you also want to validate that an EFS volume could be provisioned by the operator? Or is that a test that would be added later?

The operator, by design, doesn't provision anything in EFS. That may or may not be coming in the future.

That said, I do intend to go deeper. With the current incarnation of the operator, I want my e2e test to create EFS artifacts ("manually"), deploy instances of the custom resource, create pods using the resulting PVCs, and check that they behave properly with respect to read- and writeability.

But so far this PR was mostly a PoC to get me used to developing in osde2e, and a save-work showing future me how to deal with resources that aren't integrated in the `helper` package.

meowfaceman:
@2uasimojo Any thoughts on this? At the very minimum we need the focus bits out, for sure.

2uasimojo:
Well.

What would it take to create a suite/bucket that requires another operator to be installed, in addition to whatever's already there with OSD?

If I were the only one who needed to do this, I wouldn't even ask, but @jharrington22 and @dustman9000 and I were just talking earlier today about another case where that would be useful.

meowfaceman:
@2uasimojo Hey, this has been sitting for a bit. Is this PR still needed?

2uasimojo:
@meowfaceman TL;DR: Nah, let's close it for now. I can always reopen later.

The aws-efs-operator has been made a [community operator](https://github.com/operator-framework/community-operators/tree/master/community-operators/aws-efs-operator/) available in OperatorHub. So without the ability in osde2e to install something from there (or "by hand") as opposed to expecting it to exist in OSD natively, we can't really expect to use this other than for manually-driven testing. And there are [simpler ways](https://github.com/2uasimojo/aws-efs-operator/pull/2) to do that.

At some point in the future we may make it an OSD operator, but there's no timeline associated with that, so no reason to leave this hanging around until then.

PR #412, opened 2020-05-31
Fix assets, introduce UPDATE_OSDE2E environment variable.

meowfaceman:
Assets were not being imported before importing the various bits and
pieces of osde2e. This was causing osde2e to fail when run outside of
the soruce directory. Additionally, an UPDATE_OSDE2E environment
variable was added that, when set, will update the osde2e binary.


meowfaceman:
All, it looks like this assets issue has broken the weather report. Once verify passes, I'm going to merge this in manually. FYI @jeefy.

PR #418, opened 2020-06-04
Introduce MOA provider.

meowfaceman:
The MOA provider has been introduced to osde2e.


jeefy:
This is pretty cut and dry with the provider interface being defined. :taco: :tada:

/lgtm

PR #419, opened 2020-06-04
SDCICD-293: Add provider metrics and test OCM metrics endpoint

jeefy:
This adds a "Metrics" method to the Provider interface and fleshes out the Metrics method for the OCM Provider.

With that method, we can flex the OCM Metrics endpoint to validate the data coming out and ensure there are no errors coming from it. 

The new test will fall under the informing suite for now.

Eventually, we should flesh out the stubbed Metrics methods for the Mock/CRC providers.


pkg/common/providers/ocmprovider/cluster.go
```diff
@@ -354,6 +354,114 @@ func (o *OCMProvider) ClusterKubeconfig(clusterID string) ([]byte, error) {
 	return []byte(resp.Body().Kubeconfig()), nil
 }
 
+// GetMetrics gathers metrics from OCM on a cluster
+func (o *OCMProvider) GetMetrics(clusterID string) (*v1.ClusterMetrics, error) {
+	clusterMetricsBuilder := v1.NewClusterMetrics()
+
+	clusterClient := o.conn.ClustersMgmt().V1().Clusters().Cluster(clusterID)
+
+	/*cluster, err := clusterClient.Get().Send()
```
meowfaceman, line 363: nit: Remove.

pkg/common/providers/ocmprovider/cluster.go
```diff
@@ -354,6 +354,114 @@ func (o *OCMProvider) ClusterKubeconfig(clusterID string) ([]byte, error) {
...
+	if err != nil {
+		log.Printf("Error retrieving cluster: %s", err.Error())
+	}
+	*/
+
+	metricsClient := clusterClient.MetricQueries()
+	alertsMetricQuery, err := metricsClient.Alerts().Get().Send()
```
meowfaceman, line 370: We should wrap this in retryer logic to make sure this doesn't explode when we hit 500s.

PR #421, opened 2020-06-08
OSD-3680 Add route monitoring during upgrades

mrbarge:
**About**

This PR presents a first cut at implementing [OSD-3680](https://issues.redhat.com/browse/OSD-3680) to perform and present 
monitoring of important routes during an OpenShift upgrade.

The PR introduces a `RouteMonitors` package which, when used, will automatically detect and monitor the following routes to ensure they continue to respond to HTTP GET requests:
- API servers
- OAuth
- Web Console
- Any OSDE2E workloads with routes

The `redmine` OSDE2E workload has been updated as part of this PR to now create a route that can hence be monitored.

Monitoring of routes is being performed using `vegeta` - this has been added as a new go module dependency.

**Configuration**

The route monitor process can be controlled through the MonitorRoutesDuringUpgrade configuration option (and `UPGRADE_MONITOR_ROUTES` environment variable)

The PR defaults to this option being enabled.

**File Output**

At the completion of a cluster upgrade, the RouteMonitors will persist the following files in the working directory (in a "route-monitors" sub-directory):
 - HDR Histograms of latency (`<route>.hist`)
 - HTML Plots of latency over time (`<route>.html`)

**Metadata**

The following metadata entries are introduced:
* `RouteLatencies` - for each route, stores the mean latency over the upgrade period (milliseconds)
* `RouteThroughputs` - for each route, stores the throughput over the upgrade period (rate of successful requests per second) 
* `RouteAvailabilities` - for each route, stores the availability over the upgrade period (ratio of successful requests)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,12 +6,14 @@ require (
 	github.com/Masterminds/semver v1.5.0
 	github.com/adamliesko/retry v0.0.0-20200123222335-86c8baac277d
 	github.com/aws/aws-sdk-go v1.29.17
+	github.com/c2h5oh/datasize v0.0.0-20200112174442-28bbd4740fee // indirect
 	github.com/code-ready/crc v1.10.0
 	github.com/emicklei/go-restful v2.9.6+incompatible
 	github.com/ghodss/yaml v1.0.0
 	github.com/google/go-github/v31 v31.0.0
 	github.com/google/uuid v1.1.1
 	github.com/hashicorp/go-multierror v1.1.0
+	github.com/influxdata/tdigest v0.0.1 // indirect
 	github.com/kylelemons/godebug v1.1.0
 	github.com/markbates/pkger v0.16.0
 	github.com/onsi/ginkgo v1.12.0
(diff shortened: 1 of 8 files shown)


mrbarge:
Hey team - I appreciate this is a big one, so there's certainly no rush needed on a review. There's probably still tweaks to make, but I just wanted to get something up as a first cut for review before I get too deep. Thanks! :)

mrbarge:
Thanks for the review @meowfaceman ! I think I'm largely done with commits now - hopefully it doesn't cause things to go up in flames when merged. I've run it through quite a few test runs now, so hopefully not.

I wasn't too sure on the metrics side if I'd done the right thing - definitely happy to take any advice on better ways to be capturing new data. :)

jeremyeder:
Is this running now?  Where can I find some metrics?

meowfaceman:
It is, but I think we'll need a follow on to expose these metrics a little more thoroughly in a way that's queryable in Prometheus.

PR #422, opened 2020-06-10
SDCICD-200. Handle secret gathering within osde2e.

meowfaceman:
Secret gathering is now handled by osde2e. This is the first step in
removing our reliance on scripting glue, which should make our job
definitions simpler.

Some refactoring was done to MOA to ensure that we'd be able to remove
the prow_run_tests.sh script.

### DIFF
--- a/configs/moa.yaml
+++ b/configs/moa.yaml
@@ -1,3 +1,1 @@
-cloudProvider:
-  providerId: moa
-  region: us-east-1
+provider: moa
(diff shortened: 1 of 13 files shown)


configs/moa.yaml
```diff
@@ -1,3 +1 @@
-cloudProvider:
-  providerId: moa
-  region: us-east-1
+provider: moa
```
jeefy, line 1: Is this supposed to be part of the PR?

jeefy: Answered my own question.

PR #423, opened 2020-06-11
SDCICD-294: Define per-ginkgo-context alerting mechanism

jeefy:
This is the foundation for fine-grained alerting on test failures (SDCICD-294). 

Currently, this only supports alerting via Slack. Emails will eventually be supported. 

Using our daemonset tests as an example, we define the alert as part of the whole test definition. If, within a 24h period, a failure threshold is met, this will alert the specified slack channel.

I moved some code around (mainly the WeatherRoundTripper) so it could be reused.

### DIFF
--- a/pkg/common/alert/alert.go
+++ b/pkg/common/alert/alert.go
@@ -2,17 +2,14 @@ package alert
 
 import (
 	"context"
-	"crypto/tls"
 	"fmt"
 	"log"
-	"net"
-	"net/http"
-	"net/url"
 	"strings"
 	"sync"
 	"time"
 
 	"github.com/openshift/osde2e/pkg/common/config"
+	osde2eProm "github.com/openshift/osde2e/pkg/common/prometheus"
 	"github.com/prometheus/client_golang/api"
 	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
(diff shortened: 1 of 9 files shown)


pkg/common/alert/alert.go
```diff
@@ -0,0 +1,165 @@
...
+		RoundTripper: osde2eProm.WeatherRoundTripper,
+	})
+
+	if err != nil {
+		return fmt.Errorf("unable to create Prometheus client: %v", err)
+	}
+
+	promAPI := v1.NewAPI(client)
+	for _, ma := range mas {
+		log.Printf("%v", ma)
```
meowfaceman, line 90: Is this intended?

pkg/common/alert/alert.go
```diff
@@ -0,0 +1,165 @@
...
+	Email string
+
+	// --- Description of Alert Triggers ---
+	// FailureThreshold is the number of failures in a rolling window
+	FailureThreshold int
+}
+
+// Notify prepares and then iterates through MetricAlerts to generate notifications
+func (mas MetricAlerts) Notify() error {
+	client, err := api.NewClient(api.Config{
```
meowfaceman, line 79: This should probably use the new CreateClient function.

PR #427, opened 2020-06-15
Safe handling of helper in routemonitor code

mrbarge:
Observed in CI job 1271955636241829888 that when a cluster fails to become ready, the routemonitors (OSD-3680) were triggering a panic as the helper they use is nil.

This commit should at least address the panic behaviour so that the execution of the remainder of the osde2e run continues normally.


jeefy:
Huh. Don't put emojis after lgtm I guess.

/lgtm

PR #434, opened 2020-06-17
SDCICD-337. Add in exports to clean up MOA script.

meowfaceman:
Add in the exports for the clean up MOA script, because of course it
needs exports.


jeefy:
Round 2

/lgtm

PR #442, opened 2020-06-19
Added velero backup check test.

aliceh:
The test is checking for the presence of weekly, daily, and hourly backups when the cluster is deployed and waits for the backups to be in the "Completed" state for 5 minutes while checking the state every 5 seconds.


go.mod
```diff
@@ -35,6 +35,7 @@ require (
 	github.com/streadway/quantile v0.0.0-20150917103942-b0c588724d25 // indirect
 	github.com/tsenart/go-tsz v0.0.0-20180814235614-0bd30b3df1c3 // indirect
 	github.com/tsenart/vegeta v12.7.0+incompatible
+	github.com/vmware-tanzu/velero v1.4.0
```
meowfaceman, line 38: We'll need to do a go mod vendor update as well while we're here and check in vendored code. I know this isn't how your team probably does this, but for now this is our process.

jeefy:
Yeah the vendor update should happen in this PR.

@aliceh Could you run a `go mod vendor` and then add that commit to this PR? :) Thanks!

pkg/e2e/operators/managedvelero.go
```diff
@@ -29,4 +36,37 @@ var _ = ginkgo.Describe("[Suite: operators] [OSD] Managed Velero Operator", func
...
+						FromUnstructured(item.UnstructuredContent(), &backup)
+					if err != nil {
+						return false, fmt.Errorf("Error casting object: %s", err.Error())
+					}
+					if backup.Status.Phase == velerov1.BackupPhaseFailed {
+						return false, fmt.Errorf("Backup failed for %s", backup.Name)
+					}
+					if backup.Status.Phase != velerov1.BackupPhaseCompleted {
+						return false, nil
+					}
```
cblecker, line 65: This loops through the backups, and if any one of them is in this state it immediately returns out a false and fails the test. Do we want to check again if something isn't complete? Either way, we should also return out which bucket is not completed if we are going to fail the test completely.

meowfaceman:
/retest
Sorry, just using this test to try out the new vendor check. :-)

pkg/e2e/operators/managedvelero.go
```diff
@@ -58,12 +59,15 @@ func checkVeleroBackups(h *helper.H) {
...
-						return false, fmt.Errorf("Backup failed for %s", backup.Name)
+						backupError = true
 					}
 					if backup.Status.Phase != velerov1.BackupPhaseCompleted {
-						return false, nil
+						backupError = true
 					}
 				}
+				if backupError == true {
+					return backupError, err
```
cblecker, line 69: isn't `true` == success?

aliceh: yes. whoops :-)

PR #444, opened 2020-06-19
SDCICD-345: Add region to metadata and metrics

jeefy:
As we're now testing across multiple regions we need to track this in our metadata and metrics :)


meowfaceman:
/lgtm

We may need to generalize this in the future so that we can get "regions" for other providers (MOA), but I think this is good for now and we can think about how to proceed here.

PR #445, opened 2020-06-19
Check if the vendor directory is up to date.

meowfaceman:
The Makefile will now (non-destructively) check to make sure that the
vendor directory is up to date.

### DIFF
--- a/scripts/check-vendor.sh
+++ b/scripts/check-vendor.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+#
+# Check if the vendor directory needs an update.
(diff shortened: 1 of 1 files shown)


meowfaceman:
/hold

The script is a bit heavyweight, but I think it'll be useful -- we've been bitten by this a few times recently.

jeefy:
I don't actually think we need this.

If there's a change to `go.mod` or `go.sum` we can safely assume `vendor` should have had an update and request any incoming PRs to update `vendor` before being merged in.

PR #446, opened 2020-06-19
Created base structure of osde2ectl with core commands

agopala95:
This is just an insight into the basic structure of osde2ectl. Once a review is done, its functionality can be fleshed out and tested.


cmd/osde2ectl/extend/cmd.go
```diff
@@ -0,0 +1,95 @@
...
+
+func init() {
+	pfs := Cmd.PersistentFlags()
+
+	pfs.StringVarP(
+		&args.clusterID,
+		"cluster-id",
+		"i",
+		"",
+		"Existing OCM cluster ID to delete.",
```
meowfaceman, line 39: Copy/paste issue.

cmd/osde2ectl/main.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/openshift/osde2e/cmd/osde2ectl/list"
+	"github.com/spf13/cobra"
+)
+
+var root = &cobra.Command{
+	Use:           "osde2ectl",
+	Long:          "Command line tool for osde2ectl.",
+	SilenceErrors: true,
+	SilenceUsage:  true,
+	// PersistentPreRun: func(cmd *cobra.Command, argv []string) {
```
meowfaceman, line 23: Delete this, but this looks like something that we may want to look into ourselves.

pkg/common/providers/ocmprovider/cluster.go
```diff
@@ -564,6 +564,10 @@ func (o *OCMProvider) ocmToSPICluster(ocmCluster *v1.Cluster) (*spi.Cluster, err
 		cluster.State(ocmStateToInternalState(state))
 	}
 
+	if properties, ok := ocmCluster.GetProperties(); ok {
```
meowfaceman, line 567: Nice, good catch here.

PR #452, opened 2020-06-25
Add gather of test project state at end of run

mrbarge:
This PR runs an `oc adm inspect` on the OSDE2E test project namespace at the completion of an `osde2e` run.

The output is saved in a file `osde2e-project.tar.gz` in the `install` directory of the `osde2e` test output.

The contents of this file can be used either in conjunction with the `must-gather` or in isolation in order to better understand the state of the `osde2e` project upon completion of the test. This can be useful for providing better context around why a test may have failed to pass when conducting follow-up analysis.

### DIFF
--- a/pkg/common/helper/helper.go
+++ b/pkg/common/helper/helper.go
@@ -318,3 +318,28 @@ func (h *H) ConvertTemplateToString(template *template.Template, data interface{
 	}
 	return cmd.String(), nil
 }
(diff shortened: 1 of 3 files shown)


meowfaceman:
/approve
/lgtm

Love this. Both approving and lgtm-ing since Jeff's on PTO.

PR #454, opened 2020-06-26
SDCICD-354. Make sure cluster health checks fire for MOA clusters.

meowfaceman:
Cluster health checks will now fire for MOA clusters.


meowfaceman:
Because this affects the signal quality of MOA, I'm going to push this through now, FYI @jeefy.

PR #456, opened 2020-06-26
Create and delete functions for osde2ectl

agopala95:
The create and delete command operations are enabled for osde2ectl. Further developments and changes will be done in future commits.

### DIFF
--- a/cmd/osde2ectl/create/cmd.go
+++ b/cmd/osde2ectl/create/cmd.go
@@ -5,12 +5,10 @@ import (
 	"log"
 	"time"
 
-	. "github.com/onsi/gomega"
 	"github.com/openshift/osde2e/cmd/osde2e/common"
 	"github.com/openshift/osde2e/cmd/osde2e/helpers"
 	"github.com/openshift/osde2e/pkg/common/cluster"
 	"github.com/openshift/osde2e/pkg/common/config"
-	"github.com/openshift/osde2e/pkg/common/events"
 	"github.com/openshift/osde2e/pkg/common/providers/ocmprovider"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
@@ -90,7 +88,7 @@ func run(cmd *cobra.Command, argv []string) error {
 	}
 
 	err := cluster.SetupCluster()
-	events.HandleErrorWithEvents(err, events.InstallSuccessful, events.InstallFailed).ShouldNot(HaveOccurred(), "failed to setup cluster")
+
 	if err != nil {
 		return fmt.Errorf("Creation of a cluster failed.")
 	}
--- a/cmd/osde2ectl/create/cmd.go
+++ b/cmd/osde2ectl/create/cmd.go
@@ -25,11 +25,12 @@ var Cmd = &cobra.Command{
 }
 
 var args struct {
-	configString string
-	customConfig string
-	clusterID    string
-	environment  string
-	kubeConfig   string
+	configString    string
+	customConfig    string
+	secretLocations string
+	clusterID       string
+	environment     string
+	kubeConfig      string
 }
 
 func init() {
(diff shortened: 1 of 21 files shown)


cmd/osde2ectl/create/cmd.go
```diff
@@ -0,0 +1,107 @@
...
+		return fmt.Errorf("Creation of a cluster failed - %s", err.Error())
+	}
+
+	if len(viper.GetString(config.Kubeconfig.Contents)) == 0 {
+		// Give the cluster some breathing room.
+		log.Println("OSD cluster installed. Sleeping for 600s.")
+		time.Sleep(600 * time.Second)
+	} else {
+		log.Printf("No kubeconfig contents found, but there should be some by now.")
+	}
```
meowfaceman, line 103: I think we can remove this section.

cmd/osde2ectl/main.go
```diff
@@ -0,0 +1,49 @@
...
+	Long: "Command line tool for osde2ectl.",
+	// SilenceErrors: true,
+	// SilenceUsage:  true,
+}
+
+func init() {
+	// Add the command line flags:
+	//pfs := root.PersistentFlags()
+	//arguments.AddDebugFlag(pfs)
+	//arguments.AddUpdateFlag(pfs)
```
meowfaceman, line 29: Do we still need these?

cmd/osde2ectl/main.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/openshift/osde2e/cmd/osde2ectl/get"
+	"github.com/openshift/osde2e/cmd/osde2ectl/list"
+	"github.com/spf13/cobra"
+)
+
+var root = &cobra.Command{
+	Use:  "osde2ectl",
+	Long: "Command line tool for osde2ectl.",
+	// SilenceErrors: true,
+	// SilenceUsage:  true,
```
meowfaceman, line 22: We should remove these.

pkg/common/providers/crc/crc.go
```diff
@@ -285,3 +285,8 @@ func getPullSecretFileContent() (string, error) {
 	}
 	return crcPullSecret, nil
 }
+
+// ExtendExpiry extends the expiration time of an existing cluster
+func (m *Provider) ExtendExpiry(clusterID string, hours uint64, minutes uint64, seconds uint64) error {
+	return nil
```
meowfaceman, line 291: Let's return an error here, fmt.Errorf("ExtendExpiry is unsupported by CRC clusters.")

pkg/common/providers/mock/mock.go
```diff
@@ -234,3 +234,8 @@ func (m *MockProvider) SetVersionList(list *spi.VersionList) {
 func (m *MockProvider) Type() string {
 	return "mock"
 }
+
+// ExtendExpiry extends the expiration time of an existing cluster
+func (m *MockProvider) ExtendExpiry(clusterID string, hours uint64, minutes uint64, seconds uint64) error {
+	return nil
```
meowfaceman, line 240: Let's return an unsupported error here too.

pkg/common/providers/moaprovider/moa.go
```diff
@@ -35,3 +35,8 @@ func New() (*MOAProvider, error) {
 func (m *MOAProvider) Type() string {
 	return "moa"
 }
+
+// ExtendExpiry extends the expiration time of an existing cluster
+func (m *MOAProvider) ExtendExpiry(clusterID string, hours uint64, minutes uint64, seconds uint64) error {
+	return nil
```
meowfaceman, line 41: You can add this to the wrapped calls gofile and just call the OCM extend expiry function. This will be useful pretty quickly.

cmd/osde2ectl/create/cmd.go
agopala95: Yeah, I just kept it just in case if the kubeconfig doesn't show up. But, I'll take it down.

cmd/osde2ectl/main.go
agopala95: I'll remove these comments.

PR #465, opened 2020-06-30
e2e tests for machinehealthchecks

yithian:
for OSD-3901


pkg/e2e/osd/machinehealthcheck.go
```diff
@@ -0,0 +1,88 @@
...
+				),
+			))
+		}
+	})
+
+	ginkgo.It("should replace unhealthy nodes", func() {
+		r := h.Runner("chroot /host -- systemctl stop kubelet")
+		r.Name = "stop-kubelet"
+		// i can't believe SecurityContext.Privileged is a pointer to a bool
+		thisIsSoDumb := true
```
jeefy, line 58: Nit: I can't believe it either, but in cases like this I typically name the variable `truePointer` ;)

yithian: that's a much more descriptive variable name that still maintains the silliness, changed :)

cblecker:
LGTM

PR #470, opened 2020-07-06
Fix Timeout comparison

enxebre:
This timeout is expected to contain the s for seconds. See https://github.com/openshift/managed-cluster-config/blob/master/deploy/osd-machine-api/011-machine-api.srep-worker-healthcheck.MachineHealthCheck.yaml#L10-L16.


enxebre:
cc @yithian

PR #472, opened 2020-07-07
SDCICD-362: Output URL of previous diff

jeefy:
We should output the URL of the previous diff. This is mainly a debugging measure but seems appropriate to keep in.


jeefy:
You're on vacation dude. :P

meowfaceman:
> You're on vacation dude. :P 

I can do a 1 line PR while I'm waiting for bloodwork. :-D

PR #475, opened 2020-07-08
Added support for get and list commands of osde2ectl

agopala95:
Added support for the stage and int environments when using the get cluster info command under osde2ectl and modified print statements for the list clusters feature.

### DIFF
--- a/cmd/osde2ectl/get/cmd.go
+++ b/cmd/osde2ectl/get/cmd.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 
+	"github.com/openshift/osde2e/cmd/osde2e/common"
 	"github.com/openshift/osde2e/pkg/common/config"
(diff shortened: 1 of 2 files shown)


cmd/osde2ectl/get/cmd.go
```diff
@@ -92,6 +106,7 @@ func run(cmd *cobra.Command, argv []string) error {
 		if err != nil {
 			return fmt.Errorf("Error getting the cluster's kubeconfig - %s", err)
 		}
+		fmt.Printf("Cluster kubeconfig info - %v", config.Kubeconfig.Contents)
```
jeefy, line 109: We shouldn't print out the kubeconfig contents to stdout/err. It should get written to a file. We don't want kubeconfig contents accidentally getting logged to a public place if this is used in CI.

I could see some merit to maybe doing it if we support setting a debug output, but at the moment it should be omitted.

PR #477, opened 2020-07-09
Disable failed user webhook testing

wanghaoran1988:


meowfaceman:
@wanghaoran1988 So while I'm okay with this, is there any indication what's going on here? I filed https://issues.redhat.com/browse/OSD-4186 to look into this. This *only* seems to fail to 4.5.

wanghaoran1988:
@meowfaceman We already have one track this issue: https://issues.redhat.com/browse/SDCICD-379, I filed a bug for this.  https://bugzilla.redhat.com/show_bug.cgi?id=1855500

PR #478, opened 2020-07-13
SDCICD-372. Sort objects by timestamp.

meowfaceman:
All objects in the metrics library are now sorted by timestamp in a
descending order. Additionally, interfaces have been added that will
allow for reverse sorting/arbitrary sorting of metrics objects.


agopala95:
\lgtm

PR #482, opened 2020-07-15
SDCICD-371. osde2ectl can create batches of clusters.

meowfaceman:
osde2ectl can now create clusters in batches, and will record the
results of each cluster provision in the report directory as well as
report on an abbreviated health check periodically during the batch
cluster creation.

Quite a bit of refactoring was necessary to support this, particularly
to excise the stateful nature of cluster creation and move the state
recordation into the e2e path. This required quite a bit of movement of
functionality:

* Moving provider creation into most clusterutil functions.
* Provisioning and health check waits can now be called separately.
* Removing most/all viper.Sets from the cluster creation path. A few
  stateful things were left, like metadata, which will not realistically
  be used by osde2ectl.
* Added clusterName to LaunchCluster in the SPI, as all providers
  depended on viper for the cluster name.

Some improvements have been made to osde2ectl commands in general to
support this development effort:

* Adding more data into osde2ectl list.
* Adding the ability to delete all clusters by owner.


jeefy:
/hold
I'm mid review as well

PR #484, opened 2020-07-15
Support for get cluster kubeconfig in osde2ectl

agopala95:
Added support for the retrieval of a cluster's kubeconfig and then download it into a file. Also added a provision to extend expiry time when retrieving the kubeconfig (By default, 30 minutes would be added to the expiry time if no input is given)


cmd/osde2ectl/get/cmd.go
```diff
@@ -65,10 +78,32 @@ func init() {
...
+		"s",
+		0,
+		"Cluster expiration extension value in seconds.",
+	)
+
 	viper.BindPFlag(config.Cluster.ID, Cmd.PersistentFlags().Lookup("cluster-id"))
-	Cmd.RegisterFlagCompletionFunc("output-format", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
-		return []string{"json", "prom"}, cobra.ShellCompDirectiveDefault
-	})
+	viper.BindPFlag(config.Kubeconfig.Path, Cmd.PersistentFlags().Lookup("kube-config-path"))
```
meowfaceman, line 106: Thinking on this, why don't we get rid of the viper.Binds entirely? We should be able to do most of osde2ectl without using viper.

cmd/osde2ectl/get/cmd.go
```diff
@@ -77,8 +112,6 @@ func run(cmd *cobra.Command, argv []string) error {
 		return fmt.Errorf("error loading initial state: %v", err)
 	}
 
-	viper.BindPFlag(config.Cluster.ID, cmd.PersistentFlags().Lookup("cluster-id"))
-
 	kubeconfigStatus, err := cmd.PersistentFlags().GetBool("kube-config")
 
 	if err != nil {
```
meowfaceman, line 117: Let's get rid of the viper.GetStrings below and just use the args struct.

cmd/osde2ectl/get/cmd.go
```diff
@@ -99,28 +132,67 @@ func run(cmd *cobra.Command, argv []string) error {
...
+		content, err := setKubeconfig(clusterID)
 		if err != nil {
 			return fmt.Errorf("Error getting the cluster's kubeconfig - %s", err)
 		}
+
+		filename := cluster.Name() + "-kubeconfig.txt"
+
+		var filePath string
+		if viper.GetString(config.Kubeconfig.Path) != "" {
+			_, err := os.Stat(viper.GetString(config.Kubeconfig.Path))
```
meowfaceman, line 147: I think we should be more explicit here. This assumes that any error trying to stat the path will be because the directory doesn't exist. Check using https://golang.org/pkg/os/#IsNotExist for this.

Also, let's use the args struct directly instead of viper here.

cmd/osde2ectl/get/cmd.go
```diff
@@ -99,28 +132,67 @@ func run(cmd *cobra.Command, argv []string) error {
...
+			filePath = filepath.Join(viper.GetString(dir), filename)
+		}
+		err = ioutil.WriteFile(filePath, content, os.ModePerm)
+
+		if err != nil {
+			return fmt.Errorf("could not write KubeConfig into a file: %v", err)
+		}
+		fmt.Println("Successfully downloaded the kubeconfig into -", filePath, ". Run the command \"export TEST_KUBECONFIG=", filePath, "\"")
+		if args.hours == 0 && args.minutes == 0 && args.seconds == 0 {
+			if err = provider.ExtendExpiry(clusterID, 0, 30, 0); err != nil {
```
meowfaceman, line 171: Does this effectively state that if we try to get the kubeconfig, we should extend the expiry by 30 minutes? I like the idea, but let's make a check to only do this if the cluster is within 30 minutes of being terminated.

cmd/osde2ectl/get/cmd.go
```diff
@@ -65,10 +78,32 @@ func init() {
...
+
+	pfs.Uint64VarP(
+		&args.minutes,
+		"minutes",
+		"m",
+		0,
+		"Cluster expiration extension value in minutes.",
+	)
+
+	pfs.Uint64VarP(
```
jeefy, line 97: I don't think we need second-level resolution. This should always just default to zero.

cmd/osde2ectl/get/cmd.go
```diff
@@ -99,28 +132,67 @@ func run(cmd *cobra.Command, argv []string) error {
...
+		fmt.Println("Extended cluster expiry time by :", args.hours, "h ", args.minutes, "m ", args.seconds, "s")
 	}
 
 	return nil
 }
 
-func setKubeconfig(clusterID string) (err error) {
-	if provider, err = providers.ClusterProvider(); err != nil {
-		return fmt.Errorf("could not setup cluster provider: %v", err)
+func setKubeconfig(clusterID string) ([]byte, error) {
```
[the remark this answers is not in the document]
jeefy: +1

PR #494, opened 2020-07-20
SDCICD-352: Check machine state during cluster readiness check

jeefy:
This adds a check during cluster readiness that lists all machines in a cluster and validates that they are in the "Running" state.


pkg/common/cluster/healthchecks/clusterversionoperator.go
```diff
@@ -37,3 +42,29 @@ func CheckCVOReadiness(configClient configclient.ConfigV1Interface, logger *log.
...
+
+// CheckMachinesObjectState lists all openshift machines and validates that they are "Running"
+func CheckMachinesObjectState(dynamicClient dynamic.Interface, logger *log.Logger) (bool, error) {
+	mc := dynamicClient.Resource(schema.GroupVersionResource{Group: "machine.openshift.io", Resource: "machines", Version: "v1beta1"})
+	obj, err := mc.List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return false, err
+	}
+	var runningPhase string
+	runningPhase = "Running"
```
meowfaceman, line 54: Can we one line this? `var runningPhase string = "Running"`?

jeefy: Derp yeah.

PR #505, opened 2020-07-25
Extend the alert pod retry.

meowfaceman:
It looks like the pod alert retry is not long enough, as the alert pod
often fails but, logging onto the cluster later, the alerts test works
as expected. This will extend the alerts pod test out to about 10
minutes of waiting, which should hopefulyl cover the delta where it's
unable to get the alerts from the alertmanager.


meowfaceman:
Because this is a minor change, I'm going to push this through.

PR #506, opened 2020-07-25
Fix batch cluster allocation panic.

meowfaceman:
Batch cluster allocation panic has been fixed when using longish batch
wait times.


meowfaceman:
This supports phase 0 scale testing, so I'm going to push this through once tests pass.

PR #507, opened 2020-07-25
Fix batch wait.

meowfaceman:
Batch wait will now work as expected and indicate the number of
successful clusters.


meowfaceman:
Because this supports phase 0 scale testing, I'm going to merge this in manually.

PR #508, opened 2020-07-26
Fix batch wait 2.

meowfaceman:
Batch wait no longer relies on channels, which were a questionable use
for what I was attempting to do. This should be more stable.


meowfaceman:
To support phase 0, I'm pushing this in as well.

PR #513, opened 2020-07-29
Remove channel close from batch cluster wait.

meowfaceman:
Looks like, unbeknownst to me, closing channels isn't strictly necessary
in go. If nothing is waiting on or listening to the cluster any more, it
looks like these will be garbage collected. I'm removing the channel
close logic because I think it's an avenue of potential panics, and
hopefully this fixes any of the latent batch provisioning failures I've
seen.


jeefy:
Huh. TIL.

/lgtm

PR #515, opened 2020-07-30
SDCICD-370: Add a 'cleanup' action to addons

jeefy:
This allows addon authors to have a separate "cleanup" action in their test harness container. It expects the container to support either no argument, or `cleanup` passed to it. Since we're calling the addon-test-harness yet again, I moved the addon-runner code into `helper` for reuse later.

Along the way I identified an issue with the helper object losing state during e2e cleanup.


pkg/e2e/e2e.go
```diff
@@ -485,11 +486,16 @@ func cleanupAfterE2E(h *helper.H) (errors []error) {
 		log.Print("No cluster ID set. Skipping OCM Queries.")
 	}
 
-	// We need to clean up our helper tests manually.
-	if !viper.GetBool(config.DryRun) {
```
jeefy, line 489: We check for a DryRun before e2eCleanup is invoked, so this was redundant.

pkg/common/helper/helper.go
```diff
@@ -339,7 +339,4 @@ func (h *H) InspectState() {
 	}
 	err = h.inspect(h.CurrentProject())
 	Expect(err).ShouldNot(HaveOccurred(), "could not inspect project '%s'", h.proj)
-
-	h.restConfig = nil
```
jeefy, line 343: The offending lines of code causing issues during cleanup

Makefile
```diff
@@ -19,10 +19,10 @@ ifndef $(GOPATH)
 endif
 
 check: diffproviders.txt
-	CGO_ENABLED=0 go test -v $(PKG)/cmd/... $(PKG)/pkg/...
-	
 	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(shell go env GOPATH)/bin v1.23.8
```
jeefy, line 22: This is functionally a minor change but it allows syntax errors to be caught and surfaced before the go tests. 

It's helpful when debugging. We shouldn't run `go test` if there is a lint failure.

PR #523, opened 2020-08-04
SDCICD-296: Refactor alert registration, add alerting script

jeefy:
This PR accomplishes 3 things:
- Add in alert script for Jenkins to call
- Fix Prometheus query logic to _actually work_
- Refactor alert registration (hindsight is 20/20)

### DIFF
--- a/pkg/common/alert/alert.go
+++ b/pkg/common/alert/alert.go
@@ -1,7 +1,6 @@
 package alert
 
 import (
-	"context"
 	"fmt"
 	"log"
 	"strings"
@@ -9,9 +8,7 @@ import (
 	"time"
 
 	"github.com/openshift/osde2e/pkg/common/config"
-	"github.com/openshift/osde2e/pkg/common/prometheus"
-	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
-	"github.com/prometheus/common/model"
+	"github.com/openshift/osde2e/pkg/metrics"
 	"github.com/slack-go/slack"
 	"github.com/spf13/viper"
 )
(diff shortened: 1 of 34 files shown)


pkg/common/alert/alert.go
```diff
@@ -82,6 +82,7 @@ func (mas MetricAlerts) Notify() error {
 
 	promAPI := v1.NewAPI(client)
 	for _, ma := range mas {
+		log.Printf("Checking %s", ma.Name)
```
meowfaceman, line 85: Won't hold up the PR on this, but how easy would it be to move this to the metrics client and add the helper functions you need to support it?

scripts/granular-alerting.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+#
+# alert will look back at the previous job failures and generate granular alerts per-test-group
+#
+
+set -e
+
+#docker pull quay.io/app-sre/osde2e
```
meowfaceman, line 8: Uncomment this for the checkin.

pkg/common/alert/alert.go
jeefy: Agree. :)

pkg/common/alert/alert.go
```diff
@@ -75,14 +72,19 @@ type MetricAlert struct {
 
 // Notify prepares and then iterates through MetricAlerts to generate notifications
 func (mas MetricAlerts) Notify() error {
-	client, err := prometheus.CreateClient()
+	/*client, err := prometheus.CreateClient()
+	if err != nil {
+		return fmt.Errorf("unable to create Prometheus client: %v", err)
+	}*/
```
meowfaceman, line 78: nit: Remove. Otherwise lgtm.

jeefy: good catch

PR #525, opened 2020-08-04
SDCICD-409: Retry checking for cluster alerts

jeefy:
There is a chance the cluster is actually healthy, but the alerts haven't cleared. Therefore we should wrap the alert-check in an `Eventually` block.


jeefy:
/hold cancel

Saw something wrong and fixed it. :)

PR #529, opened 2020-08-05
Support for cluster version retrieval in osde2ectl

agopala95:
Added support in osde2ectl for cluster version retrieval when 'get' or 'list' commands are used.


pkg/common/providers/ocmprovider/cluster.go
```diff
@@ -628,3 +638,72 @@ func (o *OCMProvider) ExtendExpiry(clusterID string, hours uint64, minutes uint6
...
+	ocmCluster, err := o.getOCMCluster(clusterID)
+
+	if err != nil {
+		return err
+	}
+
+	cluster, err := o.ocmToSPICluster(ocmCluster)
+	if err != nil {
+		return err
+	}
```
meowfaceman, line 656: We can probably get rid of this -- I don't think we need to do this conversion here.

pkg/e2e/e2e.go
```diff
@@ -334,6 +334,10 @@ func runGinkgoTests() error {
 			}
 			events.RecordEvent(events.UpgradeSuccessful)
 
+			if err = provider.AddProperty(clusterID, "UpgradeVersion", viper.GetString(config.Upgrade.ReleaseName)); err != nil {
+				log.Printf("Error while adding upgrade version property to cluster via OCM: %v", err)
+			}
+
```
meowfaceman, line 340: I'm wondering if we should add this upgrade version at the start of time if it's set. Maybe we can do that as a follow on -- what do you thin @agopala95, @jeefy?

agopala95: @meowfaceman Did you mean adding the property as soon as the upgrade has just started to run/install? I put it here so that I ensure the property is only added if the upgrade has finished installing in the cluster. I also assumed that osde2e might run upgrades on existing clusters and test them.

pkg/common/providers/ocmprovider/cluster.go
agopala95: Yeah I guess this can be removed without affecting functionality.

pkg/e2e/e2e.go
jeefy: I'm of two minds on this:
1) The upgrade version should be set at the start of the **whole** test run
OR
2) The upgrade version should be set just before the upgrade is attempted.

The first shows the intent of the whole job run at a cluster level.
The second shows the intended state of a cluster regardless of success.

If I had to unilaterally decide, I'd opt for number one.

PR #532, opened 2020-08-06
Fix summary failure rate calculation.

meowfaceman:


meowfaceman:
This is a small fix and it'll make the OSDe2e reports look better, so I'm going to push this through once the checks pass.

PR #542, opened 2020-08-11
SDCICD-416. Generalize reporting.

meowfaceman:
Reporting now supports generic reporters and generic formats.


jeefy:
lgtm as well

Will this change require updates to the Jenkins job?

meowfaceman:
Actually no, shouldn't require any changes -- the jobs use scripts which will be updated as part of this PR.

PR #552, opened 2020-08-19
OSD-3356 Cluster upgrades use managed-upgrade-operator

mrbarge:
### Overview

This PR changes the behaviour of OSDE2E to use the [managed-upgrade-operator](https://github.com/openshift/managed-upgrade-operator) for performing OSD cluster upgrades.

Additionally, tests have been added to verify the managed-upgrade-operator's presence on OSD clusters, as well as verify its behaviour when provided with different types of input.

### Configuration changes

The PR introduces three new configuration items:

- Upgrade.ManagedUpgrade (env var `UPGRADE_MANAGED`)

If TRUE, this instructs OSDE2E to perform cluster upgrades using the managed-upgrade-operator. If FALSE, cluster upgrades take place using the existing approach of patching CVO. Defaults to TRUE.

(It is hoped that after a successful trial of using the managed-upgrade-operator for E2E tests, the existing method of upgrading can be phased out and this config item eliminated)

- Upgrade.ManagedUpgradeTestPodDisruptionBudgets (env var `UPGRADE_MANAGED_TEST_PDBS`)

If TRUE, OSDE2E will create a workload prior to upgrading that shall test the managed-upgrade-operator's ability to handle disruptive workloads protected by Pod Disruption Budgets. If FALSE, this workload is not created prior to upgrades. (Default to FALSE initially until managed-upgrade-operator handles PDBs)

- Upgrade.WaitForWorkersToManagedUpgrade (env var `UPGRADE_WAIT_FOR_WORKERS`)

If TRUE, OSDE2E will wait for all worker nodes to upgrade before considering the upgrade complete. If FALSE, OSDE2E will consider an upgrade complete as soon as the Control Plane is done. Defaults to TRUE.

### Other changes/fixes

Some other small changes have been made to OSDE2E in support of this PR:
- OSDE2E workloads can now support creation of Pod Disruption Budgets.
- A new workload "pdb" has been added which creates pods and pod disruption budgets (PDBs) for those pods. 
- Supporting functions have been added to `pkg/common/prometheus` to return a Prometheus client connected to the in-cluster Prometheus. This can be used by OSDE2E tests to query the state of the in-cluster Prometheus if needed. An example of this is demonstrated in the new tests introduced in `pkg/e2e/operators/managedupgrade.go` to verify that the operator is setting correct metrics.
- Changes have been made in `pkg/e2e/operators/operators.go` to support the ability to verify the presence of Cluster Roles and Cluster Role Bindings which have 'random' suffixes. The Velero, Configure Alertmanager, and Managed Upgrade Operator operators benefit from this change and their tests have been updated accordingly to now verify the presence of Cluster Roles/Cluster Role Bindings.
- A small fix to the "Writing Tests" documentation to correct the arguments of `pkger`.

### Anticipated side-effects

- The managed-upgrade-operator performs a scale-out of 1 worker node prior to upgrading. This worker node is removed upon completion of the upgrade. This may impact regions where AWS/GCP quota is already stretched thin.
- The time taken to complete an upgrade will extend by approximately 45-60 minutes. This is due to time taken to scale a worker (~10m), time to verify/action PDB draining (~20m), and general time to upgrade all workers (~20m).

### Testing

This PR has been tested with the following scenarios
- Testing/upgrading an existing `staging` cluster, Z-stream-only 
- Testing/upgrading an existing `staging` cluster, Y-stream jump (4.3 to 4.4)
- Installing/testing/upgrading a new cluster in `staging`.


mrbarge:
As this is a big change that could break things, am adding a hold whilst it's assessed. cc @dofinn FYI.

pkg/common/upgrade/managed.go
```diff
@@ -0,0 +1,273 @@
...
+
+	image := viper.GetString(config.Upgrade.Image)
+	releaseName := viper.GetString(config.Upgrade.ReleaseName)
+
+	// determine requested upgrade targets
+	if image != "" {
+		return cVersion, fmt.Errorf("image-based managed upgrades are unsupported")
+	}
+
+	upgradeVersion := strings.Replace(releaseName, "openshift-v", "", -1)
```
meowfaceman, line 81: Try using [our utility function](https://godoc.org/github.com/openshift/osde2e/pkg/common/util#OpenshiftVersionToSemver) for this instead.

pkg/common/upgrade/managed.go
```diff
@@ -0,0 +1,273 @@
...
+	upgradeVersionParsed := semver.MustParse(upgradeVersion)
+
+	targetChannel, err := VersionToChannel(upgradeVersionParsed)
+	if err != nil {
+		return cVersion, fmt.Errorf("unable to channel from version: %v", err)
+	}
+
+	// Create Pod Disruption Budget test workloads if desired
+	if 	viper.GetBool(config.Upgrade.ManagedUpgradeTestPodDisruptionBudgets) {
```
meowfaceman, line 90: nit: Looks like there's an errant tab here.

pkg/common/upgrade/managed.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+
+	// If success, add the workload to the list of installed workloads
+	h.AddWorkload(pdbWorkloadName, h.CurrentProject())
+
+	return nil
+}
+
+// IsManagedUpgradeDone returns with done true when a managed upgrade is complete.
+func IsManagedUpgradeDone(h *helper.H, desired *configv1.Update) (done bool, msg string, err error) {
```
meowfaceman, line 232: Does this need to be exported since it's all in the same package?

meowfaceman:
Also, the vendor check failure just means you should run `go mod vendor` and check in go.mod/go.sum/vendor. :-)

pkg/common/upgrade/managed.go
```diff
@@ -0,0 +1,273 @@
...
+upgradeWindow:
+  timeOut: 60
+nodeDrain:
+  timeOut: 10
+healthCheck:
+  ignoredCriticals:
+  - DNSErrors05MinSRE
+  - MetricsClientSendFailingSRE
+  - UpgradeNodeScalingFailedSRE
+`
```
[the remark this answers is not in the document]
mrbarge: Great idea, thank you - and will massively help if we ever want to templatize the values in there. I've done this in commit 2fbe0d2

pkg/common/upgrade/managed.go
```diff
@@ -0,0 +1,273 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"github.com/openshift/osde2e/pkg/common/config"
+	"github.com/openshift/osde2e/pkg/common/helper"
+)
+
+var (
```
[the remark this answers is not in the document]
mrbarge: None at all - fixed in 2fbe0d2

pkg/common/upgrade/managed.go
mrbarge: Perfect - done in 2fbe0d2

pkg/common/upgrade/managed.go
mrbarge: Fixed!

pkg/common/upgrade/managed.go
mrbarge: Not at all - fixed in 2fbe0d2

mrbarge:
Thanks for the :eyes: @meowfaceman - I've addressed all the comments and your tip has sorted out the vendor issue as well.

pkg/common/upgrade/managed.go
```diff
@@ -0,0 +1,268 @@
...
+	_, err := helper.ApplyYamlInFolder(pdbWorkloadDir, h.CurrentProject(), h.Kube())
+	if err != nil {
+		return fmt.Errorf("can't create upgrade PDB workload: %v", err)
+	}
+
+	// Give the cluster a second to churn before checking
+	time.Sleep(3 * time.Second)
+
+	// Wait for all pods to come up healthy
+	err = wait.PollImmediate(15*time.Second, 10*time.Minute, func() (bool, error) {
```
jeefy, line 210: Could this timeout be lessened? I'm starting to get concerned with the amount of 10min timeouts that we're accumulating. :) 

Also, should this be MaxDuration instead of a static 10min?

pkg/e2e/operators/managedupgrade.go
```diff
@@ -0,0 +1,304 @@
...
+	"github.com/prometheus/common/model"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"time"
+)
+
+var managedUpgradeOperatorTestName string = "[Suite: operators] [OSD] Managed Upgrade Operator"
```
jeefy, line 23: This should be put in `[Suite: informing]` since it's a new test.

pkg/common/upgrade/managed.go
mrbarge: Oh absolutely, sorry about that - that's me being lazy and copy/pasting an existing block without putting any thought into the timings. I've tightened it up quite a bit in the latest push.

pkg/e2e/operators/managedupgrade.go
mrbarge: Thanks - done!

PR #553, opened 2020-08-19
SDCICD-437: Support separate addon cleanup harnesses

jeefy:
This adds the `ADDON_CLEANUP_HARNESSES` config option and updates the docs to support it.

### DIFF
--- a/docs/Addons.md
+++ b/docs/Addons.md
@@ -20,8 +20,6 @@ Regarding addon testing, OSDe2e has three primary config options: `ADDON_IDS`, `
 
 `ADDON_TEST_USER` will specify the in-cluster user that the test harness containers will run as. It allows for a single wildcard (`%s`) which will automatically be evaluated as the OpenShift Project the test harness is created under.
 
-If your addon test creates or affects anything outside of the OSD cluster lifecycle, a separate cleanup action is required. If `ADDON_RUN_CLEANUP` is set to `true`, OSDe2e will run the test harness a **second time** passing the argument `cleanup` to the container.
-
 ```
 env:
 - name: ADDON_IDS
@@ -31,6 +29,17 @@ env:
 - name: ADDON_TEST_USER
   value: system:serviceaccount:%s:cluster-admin
 ```
+
+### Addon Cleanup
+
+If your addon test creates or affects anything outside of the OSD cluster lifecycle, a separate cleanup action is required. If `ADDON_RUN_CLEANUP` is set to `true`, OSDe2e will run the test harness a **second time** passing the argument `cleanup` to the container.
+
+There may be a case where a separate cleanup container/harness is required. That may be configured using the `ADDON_CLEANUP_HARNESSES` config option. It is formatted in the same way as `ADDON_TEST_HARNESSES`. This however, may cause some confusion as to what is run when:
+
+`ADDON_RUN_CLEANUP` is true, and `ADDON_CLEANUP_HARNESSES` is not set, OSDe2e will only run `ADDON_TEST_HARNESSES` again, passing the `cleanup` argument.
+
+`ADDON_RUN_CLEANUP` is true, and `ADDON_CLEANUP_HARNESSES` is set, OSDe2e will only run the `ADDON_CLEANUP_HARNESSES`, passing no arguments.
+
 
 ### **Getting an OCM refresh token for your tests**
 
--- a/pkg/common/config/config.go
+++ b/pkg/common/config/config.go
@@ -250,12 +250,17 @@ var Addons = struct {
 	// RunCleanup is a boolean to specify whether the testHarnesses should have a separate
 	// cleanup phase. This phase would run at the end of all e2e testing
 	RunCleanup string
+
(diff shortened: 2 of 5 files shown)


pkg/common/config/config.go
```diff
@@ -250,12 +250,17 @@ var Addons = struct {
...
-	IDs:           "ids",
-	TestHarnesses: "testHarnesses",
-	TestUser:      "addons.testUser",
-	RunCleanup:    "addons.runCleanup",
+	IDsAtCreation:    "addons.idsAtCreation",
+	IDs:              "ids",
+	TestHarnesses:    "testHarnesses",
+	TestUser:         "addons.testUser",
+	RunCleanup:       "addons.runCleanup",
+	CleanupHarnesses: "cleanupHarnesses",
```
meowfaceman, line 263: Huh, odd. I have a nit to expand all of these to addons.*, so:

```
addons.ids
addons.testHarnesses
addons.cleanupHarnesses
```

I must have forgotten this when I set this up. I realize that ids/testHarnesses isn't part of your change, but mind fixing them while we're here?

pkg/e2e/e2e.go
```diff
@@ -503,8 +503,16 @@ func cleanupAfterE2E(h *helper.H) (errors []error) {
...
 	log.Printf("Addon cleanup: %v", viper.GetBool(config.Addons.RunCleanup))
 	if viper.GetBool(config.Addons.RunCleanup) {
+		// By default, use the existing test harnesses for cleanup
+		harnesses := strings.Split(viper.GetString(config.Addons.TestHarnesses), ",")
+
+		// Check if cleanup harnesses exist and if so, use those instead
+		cleanupHarnesses := viper.GetString(config.Addons.CleanupHarnesses)
+		if len(cleanupHarnesses) > 0 {
+			harnesses = strings.Split(viper.GetString(config.Addons.CleanupHarnesses), ",")
+		}
```
meowfaceman, line 513: Think we need an else here.

jeefy: I don't think we do. I default the cleanup harnesses to be the previously-set test harness, and overwrite `harnesses` only if it's present/valid.

meowfaceman: Oh, I'm sorry. I overlooked the assign -- yes, please disregard.

docs/Addons.md
```diff
@@ -32,6 +30,17 @@ env:
...
 
+### Addon Cleanup
+
+If your addon test creates or affects anything outside of the OSD cluster lifecycle, a separate cleanup action is required. If `ADDON_RUN_CLEANUP` is set to `true`, OSDe2e will run the test harness a **second time** passing the argument `cleanup` to the container.
+
+There may be a case where a separate cleanup container/harness is required. That may be configured using the `ADDON_CLEANUP_HARNESSES` config option. It is formatted in the same way as `ADDON_TEST_HARNESSES`. This however, may cause some confusion as to what is run when:
+
+`ADDON_RUN_CLEANUP` is true, and `ADDON_CLEANUP_HARNESSES` is not set, OSDe2e will only run `ADDON_TEST_HARNESSES` again, passing the `cleanup` argument.
+
+`ADDON_RUN_CLEANUP` is true, and `ADDON_CLEANUP_HARNESSES` is set, OSDe2e will only run the `ADDON_CLEANUP_HARNESSES`, passing the `cleanup` argument.
```
Boomatang, line 41: As the `ADDON_CLEANUP_HARNESSES` can only be run when the `ADDON_RUN_CLEANUP` flag is set, I don't see the need to pass in the `cleanup` argument.

jeefy: That's a fair point. :) I'll update the docs and logic quick.

pkg/e2e/e2e.go
Boomatang: On the if/else statement, I feel h.RunAddonTests() should possible be move in to so not to have to pass the "cleanup" flag to the clean up harnesses

pkg/e2e/e2e.go
```diff
@@ -503,8 +503,18 @@ func cleanupAfterE2E(h *helper.H) (errors []error) {
...
 	log.Printf("Addon cleanup: %v", viper.GetBool(config.Addons.RunCleanup))
 	if viper.GetBool(config.Addons.RunCleanup) {
+		// By default, use the existing test harnesses for cleanup
+		harnesses := strings.Split(viper.GetString(config.Addons.TestHarnesses), ",")
+		arguments := []string{"cleanup"}
+
+		// Check if cleanup harnesses exist and if so, use those instead
+		cleanupHarnesses := viper.GetString(config.Addons.CleanupHarnesses)
+		if len(cleanupHarnesses) > 0 {
+			harnesses = strings.Split(viper.GetString(config.Addons.CleanupHarnesses), ",")
```
meowfaceman, line 513: nit: One last nit I missed -- just reuse `cleanupHarnesses` here instead of doing the second `viper.GetString`.

jeefy: oh duh

PR #554, opened 2020-08-19
SDCICD-436: Refactor incluster alert check, output all active alerts

jeefy:
Checking cluster alert state is causing grief either not running OR coming back with errors, but there's not enough debugging info.

This removes the Eventually block and will both output additional failure info and also all active alerts when the cluster check occurs.

### DIFF
--- a/pkg/e2e/state/alerts.go
+++ b/pkg/e2e/state/alerts.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"text/template"
-	"time"
 
 	"log"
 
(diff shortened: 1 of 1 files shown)


pkg/e2e/state/alerts.go
```diff
@@ -68,42 +68,32 @@ var _ = ginkgo.Describe(clusterStateTestName, func() {
...
-			stopCh := make(chan struct{})
-			// run tests
-			if err = r.Run(alertsTimeoutInSeconds, stopCh); err != nil {
-				log.Printf("Error running command on pod: %s", err.Error())
-				return false
-			}
+		stopCh := make(chan struct{})
+		// run tests
+		err = r.Run(alertsTimeoutInSeconds, stopCh)
+		Expect(err).NotTo(BeNil(), "error running alert gatherer")
```
meowfaceman, line 74: Should these be `NotTo` or `To`? We want all these errors to be nil, right?

jeefy:
Ding

PR #559, opened 2020-08-20
SDCICD-420: Update docs removing makefile references, add more config context

jeefy:
This PR removes all references to the old `make test` usage and updates the docs with the current style of how to run `osde2e` locally. 

As config options are bountiful, it also updates the config package with inline-comments of their environment variables. This should clear up most questions when writing a local script that uses environment variables, and also show up in the godoc (eventually)

cc @cblecker


meowfaceman:
This looks good to me, but I'll wait for cblecker's feedback.

cblecker:
LGTM too. Thank you!

PR #566, opened 2020-09-03
Move quota check to launch method, remove need for viper.set

jeefy:
Viper apparently isn't entirely thread safe. So I removed the need for it entirely. 

It required a minor refactor of our quotaCheck method, which now expects a `flavour` argument.


jharrington22 (approved):
looks good to me

PR #568, opened 2020-09-08
Added a cluster uninstall check for cluster deletion

agopala95:
Added a check to see if clusters being uninstalled/deleted are completely gone before confirming with the requesting user.


pkg/common/providers/ocmprovider/cluster.go
```diff
@@ -236,7 +236,18 @@ func (o *OCMProvider) DeleteCluster(clusterID string) error {
 	if err != nil {
 		return fmt.Errorf("couldn't delete cluster '%s': %v", clusterID, err)
 	}
-	return nil
+
+	time.Sleep(300 * time.Second)
```
jeefy, line 240: We shouldn't block here. What would be more appropriate would be to use a `wait.PollImmediate`. See https://github.com/openshift/osde2e/blob/5c18d4991183b47c7e34e8755789947fd4e5f8b7/pkg/e2e/verify/namespace_webhook.go#L280-L284 for an example

pkg/common/providers/ocmprovider/cluster.go
```diff
@@ -236,7 +236,18 @@ func (o *OCMProvider) DeleteCluster(clusterID string) error {
...
+	_, err = o.GetCluster(clusterID)
+	if err != nil {
+		if strings.Contains(err.Error(), "identifier is '404', code is 'CLUSTERS-MGMT-404'") {
+			log.Printf("Cluster %s has been cleanly removed and no longer exists", clusterID)
+			return nil
+		}
+		return fmt.Errorf("Cluster %s still exists with cluster information retrieval error: %v", clusterID, err)
```
jeefy, line 247: Within the wait I would just output when there's an unexpected error returned from `o.GetCluster`

pkg/common/providers/ocmprovider/cluster.go
jeefy: Also I'd poll ever 15s over 5min

jeefy (changes requested):
Hopefully a quick change, looks good mostly!

PR #576, opened 2020-09-15
SDCICD-440: Force updateNamespace to wait for changes to occur

jeefy:
This ensures that `updatedNamespace` doesn't complete until the change to a namespace object has been reflected in the API. If the namespace change hasn't been reflected within 2min, it should timeout (and thus error the job)

### DIFF
--- a/pkg/e2e/verify/namespace_webhook.go
+++ b/pkg/e2e/verify/namespace_webhook.go
@@ -210,8 +210,10 @@ func deleteGroup(groupName string, h *helper.H) error {
 }
 
 func updateNamespace(namespace string, asUser string, userGroup string, h *helper.H) error {
+	// reset impersonation upon return
+	defer h.Impersonate(rest.ImpersonationConfig{})
 
-	// reset impersonation
+	// reset impersonation at the beginning just-in-case
 	h.Impersonate(rest.ImpersonationConfig{})
 
 	// Verify the namespace already exists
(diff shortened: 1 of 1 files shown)


pkg/e2e/verify/namespace_webhook.go
```diff
@@ -231,7 +231,16 @@ func updateNamespace(namespace string, asUser string, userGroup string, h *helpe
 		UserName: asUser,
 		Groups:   userGroups,
 	})
-	_, err = h.Kube().CoreV1().Namespaces().Update(context.TODO(), ns, metav1.UpdateOptions{})
+	updatedNamespace, err := h.Kube().CoreV1().Namespaces().Update(context.TODO(), ns, metav1.UpdateOptions{})
+
+	err = wait.PollImmediate(2*time.Second, 1*time.Minute, func() (bool, error) {
+		ns, err = h.Kube().CoreV1().Namespaces().Get(context.TODO(), namespace, metav1.GetOptions{})
+		if err != nil {
+			return false, fmt.Errorf("failed to find namespace to update '%s': %v", namespace, err)
```
cblecker, line 239: nit: this error doesn't seem right. you've already updated the ns, right?

pkg/e2e/verify/namespace_webhook.go
```diff
@@ -231,7 +231,16 @@ func updateNamespace(namespace string, asUser string, userGroup string, h *helpe
 		UserName: asUser,
 		Groups:   userGroups,
 	})
-	_, err = h.Kube().CoreV1().Namespaces().Update(context.TODO(), ns, metav1.UpdateOptions{})
+	updatedNamespace, err := h.Kube().CoreV1().Namespaces().Update(context.TODO(), ns, metav1.UpdateOptions{})
```
cblecker, line 234: should this error be checked, and short circuit the get below?

pkg/e2e/verify/namespace_webhook.go
jeefy: Yep! Updated.

pkg/e2e/verify/namespace_webhook.go
jeefy: Absolutely! Updated.

PR #577, opened 2020-09-15
SimpleHTTPServer has been merged into http.server in python3

yithian:
python3 is the version used in rhel8 images

for SDCICD-484


yithian:
/hold

i think this will work but i still need to test it against rhel7 and rhel8 images

yithian:
i've verified this against rhel7

PR #580, opened 2020-09-18
Add missing PDB workload for managed upgrades

mrbarge:
This adds some Pod Disruption Budget workloads intended to be part of PR #552 but were missing.

These workloads are controlled via the `UPGRADE_MANAGED_TEST_PDBS` environment variable which, by default currently, is set to `False` (which is why we didn't notice this!). This default will not change as part of this PR.

Once a future release of the `managed-upgrade-operator` rolls out which will be able to handle PodDisruptionBudget workloads during cluster upgrades, we'll propose enabling this workload during upgrades.

This PR has been tested with an OSDE2E cluster upgrade with `UPGRADE_MANAGED_TEST_PDBS` enabled and the creation of the resulting workloads verified (noting that currently such an upgrade will ultimately fail because the PodDisruptionBudgets will stall the upgrade).


jeefy:
Will we need to add `UPGRADE_MANAGED_TEST_PDBS` as a config option to some of the pipelines in openshift/release (or the existing configs) to have these run?

mrbarge:
> Will we need to add `UPGRADE_MANAGED_TEST_PDBS` as a config option to some of the pipelines in openshift/release (or the existing configs) to have these run?

Once we've got the upgrade operator successfully handling PDBs (sometime this sprint), I'm hoping that what we can do is submit a subsequent PR to set the default value of `UPGRADE_MANAGED_TEST_PDBS` to `True` (currently it's False in `config.go`) in order for it to be created by default during e2e upgrades. Though I'm definitely happy to err on the side of whatever you'd prefer if there's other ways we could go about it.

PR #581, opened 2020-09-18
SDCICD-447: Add marketplace AMI flag to moa provider

jeefy:
This should force all existing MOA pipelines to use the Marketplace AMIs by passing the flag to OCM.


pkg/common/providers/moaprovider/cluster.go
```diff
@@ -56,6 +56,7 @@ func (m *MOAProvider) LaunchCluster(clusterName string) (string, error) {
 	}
 
 	clusterProperties, err := m.ocmProvider.GenerateProperties()
+	clusterProperties["moa_use_marketplace_ami"] = "true"
```
vkareh, line 59: Since you import `moactl`, you could import `"github.com/openshift/moactl/pkg/ocm/properties"` and do
```go
clusterProperties[properties.UseMarketplaceAMI] = "true"
```
Not a blocker, though

jeefy: We populate other cluster properties that we want going into OCM (mainly MadeByOSDe2e and which type of job it is) so we should keep this how it is :)

vkareh: sounds good

PR #584, opened 2020-09-22
Adding node drain test workload deployment for managed-upgrade-operator

ravitri:
This PR regards https://issues.redhat.com/browse/OSD-4934 Jira which regards adding node drain test workload to osde2e flow. Based on the functionality provided by PR https://github.com/openshift/managed-upgrade-operator/pull/133, the managed-upgrade-operator would automatically handle the deletion of pods stuck in Terminating state.

Remarks:

1. Node drain deployment added at /assets/workloads/e2e/drain/node-drain-test-deployment.yaml
2. The command `pkger -include $(pwd)/assets` was run to include this asset change
3.  Env variable UPGRADE_MANAGED_TEST_DRAIN can be provided to osde2e to mention if the node drain workload needs to be created or not (Defaults to false).
4. The flow was tested on staging cluster against PR https://github.com/openshift/managed-upgrade-operator/pull/133 and it went well. 
5. There was one issue faced during the upgrades which regards https://issues.redhat.com/browse/OSD-5292 and will be handled separately.


ravitri:
/hold

Will hold this PR till https://github.com/openshift/managed-upgrade-operator/pull/133 is merged. Once done, will rebase and unhold.

pkg/common/upgrade/managed.go
```diff
@@ -223,6 +238,49 @@ func createPodDisruptionBudgetWorkloads(h *helper.H) error {
 	return nil
 }
 
+// createNodeDrainWorkload creates test workload to test Node Drain functionality
+// which the managed-upgrade-operator is expected to be able to handle
+func createNodeDrainWorkload(h *helper.H) error {
```
mrbarge, line 243: To reduce code duplication between this and `createPodDisruptionBudgetWorkloads` which are very similar, it's probably nicer to refactor them both into something generic (eg. `createManagedUpgradeWorkload`) which can accept the workload that need creating as an parameter.

ravitri: @mrbarge - Agreed.. That's a nice suggestion.. Will work on it tomorrow mostly since today will be occupied in hackday ahead.. Will update soon on it.. Thanks

ravitri: Added changes.. I think it would be better to separate out these workload constants to https://github.com/openshift/osde2e/blob/main/pkg/common/config/config.go which can be the sole place to add/remove workload vars eventually along with other vars and we can still have the same functionality.. For now, I kept it simple.. Let me know if you prefer having it any other way..

ravitri:
The `pkged.go` file in commits is causing the conflict. I'll resolve it and rebase when will unhold this PR after https://github.com/openshift/managed-upgrade-operator/pull/133 is merged and more tests are run..

pkg/common/upgrade/managed.go
```diff
@@ -75,12 +82,25 @@ func TriggerManagedUpgrade(h *helper.H) (*configv1.ClusterVersion, error) {
...
+	// Create Node Drain test workloads if desired
+	if viper.GetBool(config.Upgrade.ManagedUpgradeTestNodeDrain) {
+		err = createManagedUpgradeWorkload(drainWorkloadName, drainWorkloadDir, h)
+		if err != nil {
+			return cVersion, fmt.Errorf("unable to setup node drain test workload for upgrade: %v", err)
+		}
+
+		// Delete node drain workload so that removal of pod stuck in terminating state will be tested by managed-upgrade-operator
+		log.Printf("Deleting workload %s to test node drain by upgrade", drainWorkloadName)
+		h.Kube().AppsV1().Deployments(h.CurrentProject()).Delete(context.TODO(), drainWorkloadName, metav1.DeleteOptions{})
```
mrbarge, line 100: IMO we don't need this to occur. This will and should happen naturally as a consequence of the cluster upgrade, rather than be initiated by e2e. (much as it would for normal workloads)

ravitri: Thanks Matt !! Agreed.. Have removed the delete workload stuff to let MUO handle it.. Have rebased and add your change..

mrbarge:
@jeefy we're happy that this is good to go now if you're able to give it some :eyes: - thanks again!

assets/workloads/e2e/drain/node-drain-test-deployment.yaml
```diff
@@ -0,0 +1,30 @@
...
+  replicas: 1
+  selector:
+    matchLabels:
+      app: node-drain-test
+  template:
+    metadata:
+      labels:
+        app: node-drain-test
+      finalizers:
+      - ingress.openshift.io/canttouchthis
```
jeefy, line 23: Stop. Hammer time.

jeefy (approved):
/approve

What is the intended method for running this test? It's currently a subset of the MUO tests, but requires setting an additional config to invoke the test which none of our pipelines currently use. 

Overall looks good, just holding off lgtm till I know the intent. :)

ravitri:
Thanks @jeefy ! 

The intended method to run this test is to create a upgrade scenario where a problematic workload on a node causes the node drain to get stuck during upgrades which is eventually detected, handled and thus, tested by the managed-upgrade-operator itself.

So, the faulty finalizer in the workload introduced by this test will result in pod to be stuck in Terminating state which would be resolved by the managed-upgrade-operator by removing the finalizer itself and allowing the node drain to proceed.

We would add more tests/scenarios as required. It also seemed better to not always enable this flag by default for now and have a boolean available to have such a workload or not. Can be changed ahead as well if that suits better to have it specifically for managed upgrade pipeline and no other e2e pipeline.

Let us know if you find this ok or have any other suggestions regarding same.

@mrbarge - Please add or clarify any points if am missing out on anything else.

mrbarge:
@ravitri I think you've covered it pretty well.

Once we have this in place, our next planned step is to enable the config flags for these workloads to default to "true" so that they're created pre-upgrade, with the end goal being our ability to test the managed-upgrade-operator's abilities to automate the handling of problematic user workloads during upgrades that (currently) require SRE to manually fix. 

The "defaulting to true" step isn't done yet because once we have this PR merged, I'm reasonably sure we'll need to tweak the timings of the upgrade operator's configuration so that we can properly test handling these workloads without it taking so long that it would blow out the time it takes E2E to upgrade from ~70-90 minutes to multiple hours. But that'll be the next PR we do. :)

PR #588, opened 2020-09-29
OSD-5292 always be overriding managed-upgrade config

mrbarge:
This PR corrects an issue whereby the `managed-upgrade-operator`'s E2E-tailored configmap may get overridden by a Hive sync mid-upgrade. 

The configmap is now checked and (only if needed) corrected during OSDE2E's repeated "is the upgrade finished" checks, so that even if it is replaced by a Hive sync it will be quickly corrected with the overriding values.

### DIFF
--- a/pkg/common/upgrade/upgrade.go
+++ b/pkg/common/upgrade/upgrade.go
@@ -78,6 +78,12 @@ func RunUpgrade() error {
 	done = false
 	if err = wait.PollImmediate(10*time.Second, MaxDuration, func() (bool, error) {
 		if viper.GetBool(config.Upgrade.ManagedUpgrade) && viper.GetBool(config.Upgrade.WaitForWorkersToManagedUpgrade) {
+			// Keep the managed upgrade's configuration overrides in place, in case Hive has replaced them
+			err = overrideOperatorConfig(h)
+			// Log if it errored, but don't cancel the upgrade because of it
+			if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/common/upgrade/upgrade.go
```diff
@@ -78,6 +78,12 @@ func RunUpgrade() error {
 	done = false
 	if err = wait.PollImmediate(10*time.Second, MaxDuration, func() (bool, error) {
 		if viper.GetBool(config.Upgrade.ManagedUpgrade) && viper.GetBool(config.Upgrade.WaitForWorkersToManagedUpgrade) {
+			// Keep the managed upgrade's configuration overrides in place, in case Hive has replaced them
+			err = overrideOperatorConfig(h)
```
ravitri, line 82: As I understand it, the Hive sync would happen every two hours be default and the upgrade would take about an hour or so on an average. This change will certainly reduce the number of failures on the upgrade start itself due to sync override but is it also possible that the last Hive sync could be done 1.5 hours before an upgrade starts and within half on hour of upgrade the Hive sync takes place again to cause the same issue again?

mrbarge: This syncs the configmap every 10 seconds. If hive has synced it, the change will be replaced in under that time period.

ravitri: Oh yes, ok understood. Thanks Matt !

PR #589, opened 2020-10-02
Support for pkger.go PR check and readme file updates

agopala95:
Added a script to check if pkged.go has updated itself or not. Also updated README.md with details on osde2ectl and included a new readme file under /docs which indicates a list of available environment variables or CLI flags that can be plugged in while running osde2e or osde2ectl.


README.md
```diff
@@ -1,176 +1,216 @@
...
 ```
-
+ 
+A list of commonly used CLI flags are included in the file /docs/Config.md.
+ 
 ##### Full custom YAML config example
 ```
 dryRun: false
 cluster:
- name: jsica-test
```
jeefy, line 117: name and multiAZ are config options under `cluster` and should remain indented.

README.md
```diff
@@ -1,176 +1,216 @@
...
- multiAZ: true
+name: jsica-test
+multiAZ: true
 ocm:
- debug: false
```
jeefy, line 120: like above, debug/token/env should remain indented under `ocm`

README.md
```diff
@@ -1,176 +1,216 @@
...
- token: [Redacted]
- env: stage
+debug: false
+token: [Redacted]
+env: stage
 tests:
- testsToRun:
```
jeefy, line 124: testsToRun et. all should also remain indented, as above.

README.md
```diff
@@ -201,3 +241,50 @@ To write your own test, see [Writing Tests].
...
+ 
+## osde2ectl
+ 
+The `osde2ectl` command is a cluster provisioning and management tool that is complementary to osde2e. The sub-commands can be viewed using the --help parameter.
+ 
+The ENV variables and config files used for osde2e can be used for osde2ectl as well as it shares the same codebase. The OCM token will also have to be exported before running osde2ectl.
+ 
+Key features include:-
+ 
+1. Cluster creation:
```
jeefy, line 254: I'm not a fan of how this gets rendered with the numbered list.

Instead, maybe `###` subheadings for each of the features.

Instead of doing `Example -->` and the code block, I'd also do something like:

**Example:** 
`./osde2ectl create --configs prod --number-of-clusters 4 --batch-size 2 --seconds-between-batches 40`

docs/Config.md
```diff
@@ -0,0 +1,277 @@
+# **Environment variable configuration**
+
+Some environment variables commonly used for pipelines under osde2e are indicated below. These flags or variables are shared by both osde2e and osde2ectl.
+
+
+## Common Environment Variables
+
+```
+Cluster related:-
```
jeefy, line 9: Could each of these subheadings be broken out rather than one large code block? Kind of like 

`### Cluster related:` and then start the new block of config options

docs/Config.md
```diff
@@ -0,0 +1,277 @@
...
+PROMETHEUS_BEARER_TOKEN:  Token needed for communicating with Prometheus.
+ 
+```
+
+## Command Line Flags for osde2e
+
+CLI flags that are commonly used include:
+ 
+For the test sub-command:
+--cluster-id: Existing OCM cluster ID to run tests against.
```
jeefy, line 181: I'd wrap these flags and the "query sub-command" flags in a code block

docs/Config.md
```diff
@@ -0,0 +1,277 @@
...
+ 
+For the query sub-command:
+--output-format:  Output format for query results (json|prom). Defaults to json. (default "-")
+
+ 
+## Common config flag values
+
+The following are the values that can be plugged in for the --configs flag when running osde2e/osde2ectl. The values correspond to existing YAML files in the /configs folder:-
+
+1. OSD environment values:
```
jeefy, line 198: For each of these options, could you do a markdown table? I think it'll make the options easier to visually parse

jeefy (changes requested):
pkger part looks great!

Doc-wise I have some formatting changes I'd like to see. Thanks!

PR #591, opened 2020-10-03
OSD-4933 OSD-4934 Enable PDB and Drain workloads for MUO

mrbarge:
This PR enables the `ManagedUpgradeTestPodDisruptionBudgets` and `ManagedUpgradeTestNodeDrain` configurations by default when performing an OSD managed upgrade.

These configurations will enable the workloads introduced in #585 and #580 to be deployed prior to upgrade commencement. These workloads create application deployments on the cluster which contain conditions that are disruptive to upgrades and node draining:
- Overly strict Pod Disruption Budgets 
- Finalizers that will never finalize.

The `managed-upgrade-operator` which drives the OSDE2E upgrade process contains the ability to automically handle these type of disruptive workloads. 

Should an upgrade test time out at the `AllWorkerNodesUpgraded` phase of an upgrade, it is a potential signal that the `manage-upgrade-operator` is having issues with these types of workloads, and the `managed-upgrade-operator` owners should be notified. [0]

Enabling these configurations will cause upgrades to take longer than they presently do. A handful of OSDE2E upgrades with these configurations enabled were performed to understand average duration of upgrade:

- 94 minutes
- 113 minutes
- 92 minutes
- 98 minutes

In all cases OSDE2E did not time out the upgrade/consider the upgrade failed. 

Refs:
- OSD-4933
- OSD-4934

[0] https://github.com/openshift/managed-upgrade-operator


mrbarge:
FYI @ravitri

jeefy:
From start to finish (including cluster-provisioning) do these tests take >
3h45m? Not 100% clear on whether those numbers are just the test or if it
also includes the whole cluster lifecycle. Our jobs currently have a
hard-cut at 3h45m so I want to make sure we stay under Prow's limit. :)

mrbarge:
> From start to finish (including cluster-provisioning) do these tests take > 3h45m? Not 100% clear on whether those numbers are just the test or if it also includes the whole cluster lifecycle. Our jobs currently have a hard-cut at 3h45m so I want to make sure we stay under Prow's limit. :)

Indeed, that's what I was worried about as well! So I ran four complete tests today in staging to see how they would go. All four finished up on time, though admittedly they're creeping a bit close..

- 3h30m
- 3h35m
- 3h31m
- 3h20m

..so I've added a further tweak to the behaviour of the upgrade operator in this PR to trim its timings down a bit, which might shave a tiny bit more off those timings.

PR #596, opened 2020-10-11
Rename managed-upgrade CR name

mrbarge:
`managed-upgrade-operator` now standardizes on a `UpgradeConfig` CustomResource name of `osd-upgrade-config`. When this change was made, it caused E2E upgades to no longer occur because E2E was creating a CR name of `osde2e-upgrade-config`. 

This PR rectifies this so that E2E will now create a CR with the expected name which will allow upgrades to occur.


mrbarge:
cc @dofinn

PR #598, opened 2020-10-13
Added tests for publishingstrategies CR

aliceh:
In this commit, I added a test to verify that dedicated-admin can not create a publishingstrategies CR, and that cluster-admin can.


cblecker:
lgtm
/assign @jeefy

PR #602, opened 2020-10-15
OSD-5479 E2E managed upgrades driven through provider

mrbarge:
This PR changes the way that managed upgrades via the `managed-upgrade-operator` are initiated to reflect the changing way in which OSD upgrades are being handled.

OSD managed upgrades are now initiated through the cloud provider (in this case OCM). To that end, this PR introduces a new `Upgrade` function to the `Provider` interface which makes a request to the provider to initiate a cluster upgrade. The MOA and OSD providers implement this function, the other providers have a no-op stub function.

Because of this change, **managed** E2E upgrades will now error if the E2E cloud provider is neither MOA or OCM.

Similarly, the `managed-upgrade-operator` config template has now been parameterized and its config values now injected by the code in `managed.go`.

As part of this change, the version of the OCM SDK was (necessarily) updated to v0.1.140, so there are some vendor changes included.

This change has been successfully tested in five separate full E2E runs (cluster provision, full suite, upgrade, suite, tear-down).

Refs: [OSD-5479](https://issues.redhat.com/browse/OSD-5479)


mrbarge:
Placing a hold as this can't go live until we promote the next release of `managed-upgrade-operator` (next week), I'm just getting this ready in advance. It's good for review, though.

/hold

mrbarge:
cc @dofinn FYI

mrbarge:
We have now promoted the `managed-upgrade-operator` so this PR can go ahead once it has passed review.
/hold cancel

PR #605, opened 2020-10-21
SDCICD-502: Support AWS CCS clusters and support supplying a list of AWS accounts for multi-cluster testing

jeefy:
- Adds AWS account configs to the OCM Provider
- Creates `aws-accounts-file` argument to `osde2ectl create` to supply multiple accounts during cluster creation
  - Mutex is used for modifying the OCM AWS account configs between cluster creation to prevent `viper.Set` concurrent read/write error

### DIFF
--- a/cmd/osde2ectl/create/cmd.go
+++ b/cmd/osde2ectl/create/cmd.go
@@ -1,6 +1,7 @@
 package create
 
 import (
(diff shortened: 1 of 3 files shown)


jharrington22 (approved):
/lgtm

PR #606, opened 2020-10-22
Expand e2e test coverage for user validation webhook OSD-5372

csheremeta:


pkg/e2e/verify/user_webhook.go
```diff
@@ -27,15 +27,390 @@ var _ = ginkgo.Describe(userWebhookTestName, func() {
...
+			h.Impersonate(rest.ImpersonationConfig{
+				UserName: "system:admin",
+				Groups: []string{
+					"system:authenticated",
+					"system:authenticated:oauth",
+				},
+			})
+			user, err := createUser(userName, identities, groups, h)
+			defer func() {
+				h.Impersonate(rest.ImpersonationConfig{})
```
fahlmant, line 50: I'm unfamiliar with the defaults for an ImpersonationConfig, but why reset it here? Is it guaranteed that using this, the user can be deleted?

pkg/e2e/verify/user_webhook.go
```diff
@@ -27,15 +27,390 @@ var _ = ginkgo.Describe(userWebhookTestName, func() {
...
+				deleteUser(user.Name, h)
+			}()
+			Expect(err).NotTo(HaveOccurred())
+			err = deleteUser(user.Name, h)
+			Expect(err).NotTo(HaveOccurred())
+		}, float64(viper.GetFloat64(config.Tests.PollingTimeout)))
+
+		// not testing osd-sre-admins because RBAC prevents them from managing
+		// users, even though webhook allows them to
```
fahlmant, line 214: I know this is a webhook test, so maybe not here, but should we have a test for this case and the below `oauth-openshift` to ensure that the RBAC does that? Again, probably not this PR but if we're calling it out, we could add a reference to the test that proves that in the comments here if/when they exist.

pkg/e2e/verify/user_webhook.go
```diff
@@ -27,15 +27,390 @@ var _ = ginkgo.Describe(userWebhookTestName, func() {
...
+			identities := []string{"OpenShift_SRE:" + util.RandomStr(5)}
+			h.Impersonate(rest.ImpersonationConfig{
+				UserName: "test-user@redhat.com",
+				Groups: []string{
+					"osd-sre-cluster-admins",
+					"system:authenticated",
+					"system:authenticated:oauth",
+				},
+			})
+			user, err := createUser(userName, identities, []string{}, h)
```
fahlmant, line 262: Message also mentions update as well. Thoughts on adding it into this test?

fahlmant: Also, can osd-sre-cluster-admins delete these users? Just not create or update?

pkg/e2e/verify/user_webhook.go
```diff
@@ -27,15 +27,390 @@ var _ = ginkgo.Describe(userWebhookTestName, func() {
...
+			addUserToGroup(userName, groups[0], h)
+			h.Impersonate(rest.ImpersonationConfig{
+				UserName: "test-user@redhat.com",
+				Groups: []string{
+					"osd-sre-cluster-admins",
+					"system:authenticated",
+					"system:authenticated:oauth",
+				},
+			})
+			user, err := createUser(userName, identities, groups, h)
```
fahlmant, line 289: Similar questions from the above test

pkg/e2e/verify/user_webhook.go
```diff
@@ -27,15 +27,390 @@ var _ = ginkgo.Describe(userWebhookTestName, func() {
...
+				h.Impersonate(rest.ImpersonationConfig{})
+				deleteUser(user.Name, h)
+			}()
+			Expect(err).NotTo(HaveOccurred())
+			err = deleteUser(user.Name, h)
+			Expect(err).NotTo(HaveOccurred())
+		}, float64(viper.GetFloat64(config.Tests.PollingTimeout)))
+
+		// osd-sre-cluster-admins can manage customer users
+		ginkgo.It("osd-sre-cluster-admins can manage customer users with other IDP and no group", func() {
```
fahlmant, line 321: Can they also manage customer users with random groups? Just thinking of another case, not sure how important it is to add

pkg/e2e/verify/user_webhook.go
fahlmant: Ah I see now, delete is further down

fahlmant (commented):
Overall looks good. I think a clarification of what "manage" specifically means would be helpful (i.e. create/update but delete is considered seperate?)

pkg/e2e/verify/user_webhook.go
csheremeta: yes, exactly - at the end of this test, we want to clean up our mess regardless of how the test went

jeefy:
/approve

I'll leave it to @fahlmant to LGTM post-review.

pkg/e2e/verify/user_webhook.go
csheremeta: right. I'm not sure where such a test should go. I don't want to put it here, because we would have to say that we expected a failure even though the webhook would allow this behavior - which would mean that if we changed the RBAC to also allow this behavior, the webhook e2e tests would break due to a change in a completely different/irrelevant repository. @lisa may have more ideas on this?

pkg/e2e/verify/user_webhook.go
csheremeta: I didn't add any updates to the test for simplicity. the code is the same, regardless of whether the user is trying to create or update a user (it only differs in the case of deleting the user), and testing update in addition to create adds a level of complexity to the tests (would need to create a `updateUser` function and attempt to determine _what_ to update in the user object that wouldn't cause unnecessary testing weirdness. considering the code is the same, I don't think the overhead is worth it

pkg/e2e/verify/user_webhook.go
csheremeta: similar answers :)

pkg/e2e/verify/user_webhook.go
csheremeta: yes, absolutely. the important bits are whether the user is in the SRE groups or not. I can add a test case to prove this

pkg/e2e/verify/user_webhook.go
fahlmant: Yep, that's totally fair. Maybe leave a comment to the effect of "only testing create for simplicity"?

pkg/e2e/verify/user_webhook.go
fahlmant: Let's take this into slack, and we can create a card if needed there.

pkg/e2e/verify/user_webhook.go
fahlmant: I'm meh on it. Won't block merging and the card even says this doesn't need to cover every case. It was just the first thing that came to mind when reading this message. Up to you if you want to add it

csheremeta:
in regards to what "manage" means - in all cases where we expect the same results for creating, updating, or deleting users, I have simply used the word "manage". this differs from the cases where the expected results are different for "create/update" vs "delete". here's a brief (I hope) explanation of the only case when we would expect different results for "create/update" vs "delete", and why:
- we want users who are in the RH groups ("osd-devaccess", "osd-sre-admins", "layered-cs-sre-admins") to ONLY use the OpenShift_SRE IdP
- we want users who use the OpenShift_SRE IdP to ONLY be in the RH groups ("osd-devaccess", "osd-sre-admins", "layered-cs-sre-admins")
- only members of the osd-sre-cluster-admins group are allowed to manage (create/update/delete) users in the RH groups or users who use the OpenShift_SRE IdP (nobody else gets to do anything with these users)

THEREFORE, osd-sre-cluster-admins should only be allowed to create or update users in the RH groups if those users are also using the OpenShift_SRE IdP, and osd-cluster-admins should only be allowed to create or update users who use the OpenShift_SRE IdP if those users are also in the RH groups. If (RH_group XOR SRE_IdP) is true for a user, that user should not exist, so osd-sre-cluster-admins should not be able to create or update any such user.

HOWEVER, it is entirely possible that such a user (where (RH_group XOR SRE_IdP) is true) existed before the webhook was put in place, and therefore still exists after the creation of the webhook. osd-sre-cluster-admins SHOULD be able to delete such a user, because we don't want those users existing in the first place. in fact, we have a card open to go through all clusters and delete such users - https://issues.redhat.com/browse/OSD-4885

fahlmant:
@csheremeta That makes sense. A small comment stating manage means all 3 at the top of the changes would be a nice to have just to be explicit. I'm also being really pedantic here. In the context of a webhook test, I personally want to know exactly what the expected permissions are, but I do see all three are implied already by the word manage.

fahlmant (approved):
/lgtm

PR #609, opened 2020-10-27
Increase PDB timeout duration for upgrades

mrbarge:
This PR increases the Pod Disruption Budget timeout for upgrades from 5 minutes to 15, as '15' is now the minimum value enforced by OCM when scheduling an upgrade. The maximum node drain time was increased to 15 accordingly.

This should correct job errors such as [this](https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/osde2e-stage-aws-e2e-upgrade-default-next/1320878116121350144).

Fixes [OSD-5674](https://issues.redhat.com/browse/OSD-5674)

### DIFF
--- a/pkg/common/providers/crc/crc.go
+++ b/pkg/common/providers/crc/crc.go
@@ -299,7 +299,7 @@ func (m *Provider) AddProperty(cluster *spi.Cluster, tag string, value string) e
 }
 
 // Upgrade CRCs initiates a cluster upgrade to the given version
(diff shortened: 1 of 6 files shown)


mrbarge:
cc @dofinn

mrbarge:
I'm not wholly sure what impact this will have to the overall upgrade time in E2E tests. I haven't tested a full end to end with it yet - but without this fix, upgrades won't commence at all, so I think it's probably critical to at least get this in initially and then observe results after.

PR #612, opened 2020-10-28
Test to verify dedicated admin can not create subjectpermissions CR

aliceh:
This commit adds tests to verify that dedicated-admins are not allowed to create subjectpermissions CRs and cluster admins are.


pkg/e2e/operators/rbac.go
```diff
@@ -50,6 +52,9 @@ var _ = ginkgo.Describe(rbacOperatorBlocking, func() {
 var _ = ginkgo.Describe(subjectPermissionsTestName, func() {
 	h := helper.New()
 	checkSubjectPermissions(h, "dedicated-admins")
+	checkSubjectPermissions(h, "sre-admins")
```
jeefy, line 55: Are `checkSubjectpernission` and `daCreateSubjectpernission` Spelling errors? I see that it's a separately-defined function, but it's so similarly named that I did a double-take. Could this be made clearer?

pkg/e2e/operators/rbac.go
```diff
@@ -76,6 +81,68 @@ func checkSubjectPermissions(h *helper.H, spName string) {
 	})
 }
 
+func makeSubjectpermission(name string) api.SubjectPermission {
```
jeefy, line 84: This should probably be `makeSubjectPermission`

pkg/e2e/operators/rbac.go
```diff
@@ -50,6 +52,9 @@ var _ = ginkgo.Describe(rbacOperatorBlocking, func() {
 var _ = ginkgo.Describe(subjectPermissionsTestName, func() {
 	h := helper.New()
 	checkSubjectPermissions(h, "dedicated-admins")
+	checkSubjectPermissions(h, "sre-admins")
+	daCreateSubjectpernission(h)
```
jeefy, line 56: `daCreateSubjectpernission` and `checkSubjectpernission` aren't reused. I would put their ginkgo Context blocks back under the Describe block, there's no benefit to breaking these out into separate functions.

jeefy (changes requested):
Couple questions/changes, overall looks good!

aliceh:
@jeefy, thanks for reviewing! I updated the code accordingly.

clcollins (approved):
/lgtm

PR #614, opened 2020-10-29
Cluster alerts check using golang instead of a runner pod

agopala95:
Modified the existing cluster alerts check to use Prometheus queries directly rather than using a runner pod. 
A few things to note:
1.The test currently supports the query only if environment variables are provided for Prometheus credentials. 
2. Mirrored the number of retries as seen in the alerts.template file in the assets/state/ folder.


pkg/e2e/state/alerts.go
```diff
@@ -49,60 +36,50 @@ func init() {
...
+		// Query for alerts with a retry count of 40 and timeout of 20 minutes
+		err = wait.PollImmediate(30*time.Second, 20*time.Minute, func() (bool, error) {
+			query := fmt.Sprintf("ALERTS{alertstate!=\"pending\",alertname!=\"Watchdog\"}")
+			context, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
+			defer cancel()
+			result, _, err := promAPI.Query(context, query, time.Now())
+			if err != nil {
+				return false, fmt.Errorf("Unable to query prom API")
+			}
+			queryresult, err = json.MarshalIndent(result, "", "  ")
```
jeefy, line 60: This seems a bit of a hack to interpret results. Instead, we should use Prometheus' native libraries/objects.

https://github.com/openshift/osde2e/blob/main/pkg/metrics/client.go#L62-L82 is a good example of retrieving a list of results and how to handle them. In this case, we could potentially see multiple alerts and, if we do, we should log all the alerts we've seen. 

This will impact the changes to `findCriticalAlerts` below, since you _should_ theoretically pass it an array of results (even if it's an empty array)

jeefy (changes requested):
Hopefully these changes won't be too big, but they're needed to fall in line with how we're handling other Prometheus queries.

If you want to work this out together let's schedule something early next week!

PR #615, opened 2020-10-29
Support for cluster version check query

agopala95:
Added support to validate cluster install and upgrade versions through a prometheus query. Users can include the cloudprovider, install version, upgrade version and environment. The count of passed/skipped results are extracted against the total results count and displayed.


jeefy (changes requested):
This is a good start, but needs a little more work done.

Ideally, we would compute for the user a pass percentage for a given install (and, if set, upgrade) path, and return to them a simple number.

For example, `./osdee2 query --version-check --install-version="openshift-v4.5.11"` should return something like `0.9755555`. Maybe we could output it as JSON and Prom as well, but the idea here is a high-level pass rate of a version in an environment. 

Let me know if I explained this well or if you have questions on it! :)

cmd/osde2e/query/cmd.go
```diff
@@ -110,5 +162,14 @@ func run(cmd *cobra.Command, argv []string) error {
 		return fmt.Errorf("error writing output: %v", err)
 	}
 
+	encoded := []rune(string(data))
+	data = []byte(string(encoded[1:(len(string(data)) - 1)]))
```
jeefy, line 166: This line is causing an error when I try to run this locally (with the necessary env vars set):

`osde2e query --install-version=openshift-v4.5.11 --version-check`

Can you double-check this @agopala95 ? :)

agopala95: Yup. Had a minor bug. Fixed it now.

PR #617, opened 2020-10-31
Fixes SDCICD-513 duplicate clusterName upon create

dofinn:
Checks if proposed cluster name exists before attempting to launch
cluster.

Will fail after 10 attempts to validate name. Can fail due to a
combination of 10 failure of duplicate names and/or OCM errors upstream
Ideally when this actually fails, it indicates we have much larger
issue that needs addressing.

### DIFF
--- a/pkg/common/providers/moaprovider/cluster.go
+++ b/pkg/common/providers/moaprovider/cluster.go
@@ -27,9 +27,9 @@ func (m *MOAProvider) IsValidClusterName(clusterName string) (bool, error) {
 
 	// Retrieve the list of clusters using pages of ten items, till we get a page that has less
 	// items than requests, as that marks the end of the collection:
-	size := 10
+	size := 50
 	page := 1
-	searchPhrase := fmt.Sprintf("name like '%s'", clusterName)
+	searchPhrase := fmt.Sprintf("name = '%s'", clusterName)
 	for {
 		// Retrieve the page:
 		response, err := collection.List().
--- a/pkg/common/providers/moaprovider/cluster.go
+++ b/pkg/common/providers/moaprovider/cluster.go
@@ -1,6 +1,7 @@
 package moaprovider
 
 import (
+	"context"
 	"fmt"
 	"log"
 	"net"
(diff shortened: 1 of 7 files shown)


pkg/common/providers/moaprovider/cluster.go
```diff
@@ -16,6 +17,47 @@ import (
...
+	// Create a context:
+	ctx := context.Background()
+
+	collection := m.ocmProvider.GetConnection().ClustersMgmt().V1().Clusters()
+
+	// Retrieve the list of clusters using pages of ten items, till we get a page that has less
+	// items than requests, as that marks the end of the collection:
+	size := 10
+	page := 1
+	searchPhrase := fmt.Sprintf("name like '%s'", clusterName)
```
jeefy, line 32: Is there a better way to look for an existing name? Instead of a `like` style lookup could we look for an exact match?

Doing that would remove the need for a loop and any ambiguity; if there's a result, the name's taken.

pkg/common/providers/ocmprovider/cluster.go
```diff
@@ -19,6 +19,46 @@ import (
...
+	// Create a context:
+	ctx := context.Background()
+
+	collection := o.conn.ClustersMgmt().V1().Clusters()
+
+	// Retrieve the list of clusters using pages of ten items, till we get a page that has less
+	// items than requests, as that marks the end of the collection:
+	size := 10
+	page := 1
+	searchPhrase := fmt.Sprintf("name = '%s'", clusterName)
```
jeefy, line 34: Why is this different from the MOA query above? Is there a difference between the two?

This query would lead me to think it should only ever return a single result. As far as I know, when someone logs in to OCM they're part of an Org, and no Org can have a cluster name collision (CS won't allow it.)

So theoretically you should be able to remove the for loop and then validate if there are zero or not-zero results.

jeefy (changes requested):
Questions/request about the actual OCM query, otherwise lgtm!

pkg/common/providers/ocmprovider/cluster.go
dofinn: > Why is this different from the MOA query above? Is there a difference between the two?
Please see previous comment

The loop accounts for the size and page requests. I used the size/page ints from the examples in the OCM sdk. We could probabaly enlarge the size to avoid more loops. However the for loop should stay IMO as when production fleet is at a large number, we will need to iterate through the pages. e819298f

pkg/common/providers/moaprovider/cluster.go
dofinn: This was a mistake, copied from testing ocmprovider before I fine tuned the syntax to an exact. Resolved in: e819298f

PR #618, opened 2020-11-02
OSD-5252 fix occasional operator upgrade test errors

mrbarge:
This PR attempts to fix several issues around sporadic failures of the operator "should upgrade from previous version" test. This affected the `configure-alertmanager-operator`, `splunk-forwarder-operator` and `rbac-permissions-operator`.

Some issues were identified:

- Updating the `subscription` could occasionally return `object has been modified`, so it's now surrounded in a `RetryOnConflict` block.
- Instead of incrementally approving `InstallPlan`s, the test was occasionally approving multiple (a race condition dependent on a delay when approving the first), which caused it to then get stuck in a holding pattern looking for a `ClusterServiceVersion` that didn't exist.

Refs [OSD-5252](https://issues.redhat.com/browse/OSD-5252)

### DIFF
--- a/pkg/e2e/operators/operators.go
+++ b/pkg/e2e/operators/operators.go
@@ -12,8 +12,12 @@ import (
 	"github.com/openshift/osde2e/pkg/common/runner"
 	"github.com/openshift/osde2e/pkg/common/templates"
 
+	appsv1 "k8s.io/api/apps/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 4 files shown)


mrbarge:
I've been running the RBAC/CAM/Splunk tests in a loop for the last couple of hours with this branch, and it's no longer hitting the errors we've been seeing in Prow. I'm hopefully this is the last time we've seen issues we see around this test - and if that's not the case, I'll recommend we take a fresh look at how to do this.

jeefy:
/approve
/lgtm
:taco:

PR #620, opened 2020-11-03
SDCICD-524: Refactor upgrade version selectors

jeefy:
Our upgrade version selectors behaved differently from the install selectors. Additionally, they expected different providers such as CIncinnati or the Release Controller. Since we expect _spi.Providers_ to have a Versions endpoint, all of that logic is moot.

This PR aims to unify the logic between install and upgrade selectors. It expects that we get valid versions, including upgrade paths, from the cluster provider, and simplifies upgrade selectors to things like "latest", "next y", "latest z". It updates the docs to reflect the change as well.

Of note: a follow-up PR into openshift/release will be forthcoming updating the pipelines to have specific upgrade paths. Until that is present, upgrade pipelines will simply be the equivalent of an install-test.


jeefy:
/hold
Holding till the openshift/release PR is staged as well.

pkg/common/config/config.go
```diff
@@ -67,21 +67,17 @@ var keyToSecretMappingMutex = sync.Mutex{}
...
-	// Env: UPGRADE_TO_CIS_IF_POSSIBLE
-	UpgradeToCISIfPossible string
+	// UpgradeToLatest will look for the newest-possible version and select that
+	// Env: UPGRADE_TO_LATEST
+	UpgradeToLatest string
 
-	// OnlyUpgradeToZReleases will restrict upgrades to selecting Z releases on stage/prod.
-	// Env: ONLY_UPGRADE_TO_Z_RELEASES
-	OnlyUpgradeToZReleases string
+	// UpgradeToNextY will look for the next Y version for the cluster and select that
```
dofinn, line 74: `UpgradeToNextY` and `upgrade-to-next-y` definitions are confusing.. Will this upgrade to the latest .z of the latest .y? If so this should be title `UpgradeToLatestY` and defined as "will look for the latest Y version and update to its latest Z"

jeefy: Here was my thought behind it....
"Latest Y" could mean, if (In Integration) we installed 4.5, and a 4.7 nightly is available, it would imply it could upgrade to 4.7. Next implies Y+1

The reality is, it's only ever looking at the available upgrade edges so that case is _extremely_ unlikely to happen and it would only ever do a Y+1

All that said, if you think it should be s/next/latest/ I can crank that out quick. :)

pkg/e2e/version.go
```diff
@@ -37,6 +37,7 @@ func ChooseVersions() (err error) {
 		err = setupUpgradeVersion(clusterVersion, versionList)
 
 		if err != nil {
+			log.Printf("error while selecting upgrade version: %v", err)
```
jharrington22, line 40: Do you need the extra log message here? Create should fail and report the error returned.

jharrington22 (approved):
Just one comment otherwise looks good.

pkg/e2e/version.go
jeefy: ACK, will fix. (Need to rebase anyway)

jharrington22 (approved):
/lgtm

PR #624, opened 2020-11-05
SDCICD-539: Fail create if invalid aws-accounts-file provided

jeefy:
There was a flaw in the logic checking whether an aws-accounts-file was provided and no explicit check occurred. This explicitly checks to make sure, if aws-accounts-file is provided, it is a valid file or it fails.


cmd/osde2ectl/create/cmd.go
```diff
@@ -150,20 +150,25 @@ func run(cmd *cobra.Command, argv []string) error {
 		}
 	}
 
-	if _, err := os.Stat(args.awsAccountsFile); err == nil {
-		csvfile, err := os.Open(args.awsAccountsFile)
-		if err != nil {
-			log.Fatalln("Couldn't open the csv file", err)
-		}
+	if args.awsAccountsFile != "" {
```
jharrington22, line 153: Should we have a dedicated flag for CCS clusters? If you want CCS but forget to pass the file here we'll create OSD clusters.

jharrington22 (commented):
Left a comment but otherwise the changes are good and will now catch a file if it doesn't exist.

cmd/osde2ectl/create/cmd.go
jeefy: Good idea, make it a two-key system.

pkg/common/providers/ocmprovider/cluster.go
```diff
@@ -104,17 +104,21 @@ func (o *OCMProvider) LaunchCluster(clusterName string) (string, error) {
...
-	awsAccessKey := viper.GetString(AWSAccessKey)
-	awsSecretKey := viper.GetString(AWSSecretKey)
-
-	if awsAccount != "" && awsAccessKey != "" && awsSecretKey != "" {
-		newCluster.CCS(v1.NewCCS().Enabled(true)).AWS(
-			v1.NewAWS().
-				AccountID(awsAccount).
-				AccessKeyID(awsAccessKey).
-				SecretAccessKey(awsSecretKey))
+	if viper.GetBool(CCS) {
```
jharrington22, line 107: :+1:

jharrington22 (approved):
/lgtm

PR #626, opened 2020-11-10
Added filters to the osde2ectl list command

agopala95:
Brought in the new filters listed below when using the osde2ectl list command to list clusters:-
1. Owner
2. Installed Version
3. Upgrade Version
4. Cluster Status

These filters can be applied along with the existing environment variables.


pkg/common/providers/ocmprovider/cluster.go
```diff
@@ -392,7 +392,7 @@ func (o *OCMProvider) ListClusters(query string) ([]*spi.Cluster, error) {
 	emptyPage := false
 	page := 1
 
-	for len(clusters) < totalItems || emptyPage {
```
agopala95, line 395: I've just modified this code snippet to stop looping if there are no results for the corresponding search string or when there are no more results to bring in. As of now, the variable emptyPage seems to serve this operation. I can revert this change if this impacts overall operation.

PR #631, opened 2020-11-16
Pod webhook

aliceh:
In this commit, we test the pod webhook and verify that the dedicated-admin user can not create pods in unprivileged namespaces on master and infra nodes. We also verify that RBAC does not allow ordinary users to create pods.

### DIFF
--- a/pkg/e2e/verify/pod_webhook.go
+++ b/pkg/e2e/verify/pod_webhook.go
@@ -16,10 +16,10 @@ import (
 	"k8s.io/client-go/rest"
 )
 
-var podWebhookTestName string = "[Suite: e2e] [OSD] pod validating webhook"
+var podWebhookTestName string = "[Suite: informing] [OSD] pod validating webhook"
 
 func init() {
-	alert.RegisterGinkgoAlert(podWebhookTestName, "SD-SREP", "Alice Hubenko", "sd-cicd-alerts", "sd-cicd@redhat.com", 4)
+	alert.RegisterGinkgoAlert(podWebhookTestName, "SD-SREP", "Matt Bargenquast", "sd-cicd-alerts", "sd-cicd@redhat.com", 4)
 }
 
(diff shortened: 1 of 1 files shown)


pkg/e2e/verify/pod_webhook.go
```diff
@@ -0,0 +1,209 @@
...
+	"github.com/openshift/osde2e/pkg/common/alert"
+	"github.com/openshift/osde2e/pkg/common/helper"
+	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/rest"
+)
+
+var podWebhookTestName string = "[Suite: e2e] [OSD] pod validating webhook"
```
jeefy, line 19: This should go in the informing suite first :)

pkg/e2e/verify/pod_webhook.go
```diff
@@ -0,0 +1,209 @@
...
+
+func init() {
+	alert.RegisterGinkgoAlert(podWebhookTestName, "SD-SREP", "Alice Hubenko", "sd-cicd-alerts", "sd-cicd@redhat.com", 4)
+}
+
+var _ = ginkgo.Describe(podWebhookTestName, func() {
+
+	h := helper.New()
+
+	ginkgo.Context("pod webhook", func() {
```
jeefy, line 29: Is there a reason you specify multiple ginkgo Contexts? I think you should only use one and put all the `It` blocks underneath it

jeefy (changes requested):
/approve

Couple minor changes before I can lgtm :)

pkg/e2e/verify/pod_webhook.go
aliceh: I used it for running individual tests. I updated so that there's a single Context.

PR #635, opened 2020-11-17
Test to verify dedicated admin can not manage SplunkForwarders

aliceh:
In this commit, we verify that dedicated-admin can not manage SplunkForwarders and the cluster admin can.


aliceh:
@cblecker please review. I included a cleanup block. 
I have to note that however, the test runs correctly I can not pass the correct namespace even with a new local variable which is very odd.

```
2020/11/16 17:28:19 Get splunkforwarder osde2e-dedicated-admin-splunkforwarder-x in namespace openshift-rbac-permissions; Error:(<nil>)
2020/11/16 17:28:19 Delete splunkforwarder osde2e-dedicated-admin-splunkforwarder-x in namespace openshift-rbac-permissions; Error:(<nil>)
```

PR #637, opened 2020-11-17
SDCICD-387: Add script/test to test ocm adoption of an OCP cluster

jeefy:
This is a script to validate that a vanilla OCP cluster shows up in OCM.

We test this by using a CI-created cluster and looking up its external cluster ID in OCM. To validate the workflow in Stage, we need to patch the cluster's Telemetry endpoint then wait until the telemetry operator cycles and registers the cluster. 

There will be a subsequent PR in openshift/release with the Prow Job that uses this


jeefy:
/hold cancel
Doing the thing :)

PR #641, opened 2020-11-20
Added function to clean up apischeme-cr-test

luis-falcon:
Added a clean up functionality to the test. 
after the test for creation of the apischeme is ran the resource will be deleted allowing for extra executions
without failures or having to manually delete the apischeme from the cluster.

https://issues.redhat.com/browse/OSD-5881

### DIFF
--- a/pkg/e2e/operators/cloudingress/apischeme_cr.go
+++ b/pkg/e2e/operators/cloudingress/apischeme_cr.go
@@ -74,6 +74,9 @@ func testCRapischemes(h *helper.H) {
 			err := addApischeme(h, as)
 			Expect(err).NotTo(HaveOccurred())
 
+			err = apiSchemeCleanup(h, as.Name)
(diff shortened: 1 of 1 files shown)


pkg/e2e/operators/cloudingress/apischeme_cr.go
```diff
@@ -75,6 +81,9 @@ func testCRapischemes(h *helper.H, operatorNamespace string) {
 			err := addApischeme(h, as, operatorNamespace)
 			Expect(err).NotTo(HaveOccurred())
 
+			err = apiSchemeCleanup(h, as.Name)
```
cblecker, line 84: Should we just add this as a defer instead?

luis-falcon: I don't mind if it is.

cblecker:
@luis-falcon can you clean up the title/body of the PR, with a link to the OSD card?

PR #643, opened 2020-11-23
OSD-5902 add sleep for rhapi_endpoint

luis-falcon:
https://issues.redhat.com/browse/OSD-5902


jeefy:
When in doubt, add another `time.Sleep()`

/approve

PR #644, opened 2020-11-24
Add optional pod prefix to pod health check

mrbarge:
This PR adds a variadic argument section to `CheckPodHealth`, allowing for more refined checking of pod status by comparing the list of retrieved pods against a supplied slice of pod name prefixes.

This allows tests which (re)use `CheckPodHealth` to not be impacted if another unrelated pod in the cluster happens to be in a Pending state.

The `guestbook` and `redmine` tests, and the managed upgrade code, have been updated to make use of the additional arguments. Features such as the cluster health check will continue to scan through all pods.

### DIFF
--- a/pkg/common/cluster/clusterutil.go
+++ b/pkg/common/cluster/clusterutil.go
@@ -272,7 +272,7 @@ func PollClusterHealth(clusterID string, logger *log.Logger) (status bool, failu
 			clusterHealthy = false
 		}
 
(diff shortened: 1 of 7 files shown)


mrbarge:
Need to fix a broken test case.
/hold

jeefy:
This is awesome. Fixing the tests shouldn't be too difficult either, so I can't wait to get this in.

/approve

mrbarge:
Thanks @jeefy - I've done a bunch more tidying-up and fixed what was breaking the tests, so this should be good now.
/hold cancel

PR #645, opened 2020-11-25
Make TEST_KUBECONFIG use the mock provider

mrbarge:
Since OCM provider-driven testing is having issues for SRE with fetching the kubeconfig, this PR attempts to get the `TEST_KUBECONFIG` method of local `kubeconfig`-based testing working again, by setting it up with the mock provider.

I found that with the existing `osde2e`, trying to use `TEST_KUBECONFIG` caused code panics because it assumes that a provider is created, but one never is. So this change attempts to address that problem in a pretty crude way via the mock provider, so that there is at least some avenue for SRE to continue using `osde2e` for testing whilst this whole issue is figured out.


jeefy:
Good catch!
/approve
/lgtm

PR #646, opened 2020-11-25
add upgrade test for reschedule

bmeng:
Adding the upgrade test for reschedule the upgrade policy before the upgrade triggered.

Haven't test this yet, since the kubeconfig issue on stage env.

### DIFF
--- a/pkg/common/spi/provider.go
+++ b/pkg/common/spi/provider.go
@@ -115,8 +115,8 @@ type Provider interface {
 	// Upgrade requests the provider initiate a cluster upgrade to the given version
 	Upgrade(clusterID string, version string, t time.Time) error
 
-	// GetUpgradePolicy gets the first upgrade policy from the top
-	GetUpgradePolicy(clusterID string) (string, error)
+	// GetUpgradePolicyID gets the first upgrade policy from the top
+	GetUpgradePolicyID(clusterID string) (string, error)
 
 	// UpdateSchedule updates the existing upgrade policy for re-scheduling
 	UpdateSchedule(clusterID string, version string, t time.Time, policyID string) error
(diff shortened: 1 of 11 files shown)


bmeng:
cc @mrbarge Please help take a look

pkg/common/spi/provider.go
```diff
@@ -114,4 +114,10 @@ type Provider interface {
 
 	// Upgrade requests the provider initiate a cluster upgrade to the given version
 	Upgrade(clusterID string, version string, t time.Time) error
+
+	// GetUpgradePolicy gets the first upgrade policy from the top
+	GetUpgradePolicy(clusterID string) (string, error)
```
mrbarge, line 119: Very minor nit, but maybe this (and its implementations) might be better named `GetUpgradePolicyID` since it's returning that specific element.. at least for the OCM provider.

mrbarge:
@bmeng - on the whole I like this addition!

I haven't tested it personally due to the same OCM permission issue, so am just purely looking at code.

The main thing I would suggest is that perhaps this test might be better suited being moved into MUO's dedicated functional test area instead of being part of the code handling the e2e cluster upgrade:
https://github.com/openshift/osde2e/blob/main/pkg/e2e/operators/managedupgrade.go

That way it's nicely isolated from the code that performs the E2E upgrade, and can actually be reported as an individual passing/failing test result.

jeefy:
/approve
Will wait on lgtm from @mrbarge

mrbarge:
Just putting a hold on until I can check with @bmeng on his plans (he's away this week) based on my feedback. I think he has been attempting to relocate the work to the MUO specific operator tests.
/hold

bmeng:
@mrbarge I just made some change to the testing.
Adding the test config for the reschedule testing.
Skip the post-upgrade testing if a reschedule defined.
Could you help take a look?

PR #647, opened 2020-11-26
OSD-3985 remove image-pruner cronjob test

mrbarge:
As part of [OSD-3985](https://issues.redhat.com/browse/OSD-3985) we will be removing the SRE image pruner cronjob on 4.6.x clusters in favour of the existing cluster image pruner.

This PR adjusts the `Prune Jobs` test so that it will not attempt to test the ability to create jobs from the SRE image pruner cronjob, as this cronjob will not exist on 4.6 clusters when [OSD-3985](https://issues.redhat.com/browse/OSD-3985) is promoted.

cc @wanghaoran1988 as original test author.


jeefy:
/approve
/lgtm
/hold
Hold for @wanghaoran1988

PR #648, opened 2020-11-26
e2e install verification test for MGO

arjunrn:
This is a simple test which verifies that the latest build of must-gather-operator can be sucessfully installed in the cluster.

### DIFF
--- a/pkg/e2e/operators/mustgather.go
+++ b/pkg/e2e/operators/mustgather.go
@@ -1,0 +1,35 @@
+package operators
+
+import (
+	"github.com/onsi/ginkgo"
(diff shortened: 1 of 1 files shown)


arjunrn:
@csheremeta I've added this e2e test for the MGO so that the e2e verify that the operator catalog is sane and it can be installed correctly. This test is part of the e2e tests for all operators and when you decide to add more tests for MGO you can add them here.

csheremeta:
nice, thanks @arjunrn! :) cc @jeefy in case you have thoughts on what MGO e2e might look like

jeefy:
Thanks!
/approve

whereswaldon (approved):

PR #650, opened 2020-12-02
osde2ectl delete should error when no id is passed

agopala95:
Modified the osde2ectl delete command to return an error if no cluster ID or owner arguments are passed.

### DIFF
--- a/cmd/osde2ectl/delete/cmd.go
+++ b/cmd/osde2ectl/delete/cmd.go
@@ -132,6 +132,8 @@ func run(cmd *cobra.Command, argv []string) error {
 			}
 		}
 		return allErrors.ErrorOrNil()
+	} else {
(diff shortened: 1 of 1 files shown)


cmd/osde2ectl/delete/cmd.go
```diff
@@ -132,6 +132,8 @@ func run(cmd *cobra.Command, argv []string) error {
 			}
 		}
 		return allErrors.ErrorOrNil()
+	} else {
```
jeefy, line 135: This won't catch it correctly. You should if/else the cluster_id check explicitly, and change the `if owner` block to be its own individual if-statement. :)

jeefy (changes requested):
Quick nit! :)

cmd/osde2ectl/delete/cmd.go
jeefy: i am dumb need more coffee

PR #660, opened 2020-12-16
Allow non-managed image-based upgrades

mrbarge:
This tiny little PR resumes allowing OSDE2E to perform image-based upgrades. Previously this was being blocked outright, regardless of whether managed upgrades were enabled or disabled in the viper config.

I've tested this successfully with an image-based upgrade in `osde2e`. Should have no impact to our existing `e2e` runs which use managed upgrades.


jeefy:
/approve
/lgtm

tiny pr <3

PR #663, opened 2020-12-16
Fixed the failing splunkforwarder test

aliceh:


pkg/e2e/operators/splunkforwarder.go
```diff
@@ -163,7 +164,12 @@ func dedicatedAaddSplunkforwarder(SplunkForwarder sfv1alpha1.SplunkForwarder, na
 		return err
 	}
 	unstructuredObj := unstructured.Unstructured{obj}
-	h.SetServiceAccount("system:serviceaccount:%s:dedicated-admin-project")
+	h.Impersonate(rest.ImpersonationConfig{
+		UserName: "test-user@redhat.com",
+		Groups: []string{
+			"dedicated-admins",
+		},
+	})
```
mrbarge, line 172: I can't see anywhere where it's resetting the impersonation when the test finishes. There should likely be something like the following, in keeping with the convention that other tests do.

```
      defer func() {
        h.Impersonate(rest.ImpersonationConfig{})                                                                                                                                                               
      }()
```

aliceh: Added the reset step. Thanks for reviewing, @mrbarge !

mrbarge:
Verified test succeeded on a staging cluster.
/lgtm
/approve

PR #665, opened 2020-12-22
OSD-6104 Fix for the failing upgradeconfig test

aliceh:


mrbarge:
Gave this a test - first verifying that `main` branch fails the test, then verifying that this PR passes it. 
Code looks fine to me also. Thanks for fixing it!
/lgtm
/approve

PR #666, opened 2020-12-23
SDCICD-519 only create route monitors when necessary

mrbarge:
This PR addresses [SDCICD-519](https://issues.redhat.com/browse/SDCICD-519) whereby the following concerning error messages were being logged:
```
Error creating route monitors: Unable to generate helper outside ginkgo
```
which, to a newcomer to reviewing CI logs, may look like a problem.

In reality this was occuring because the route monitors were being created regardless of whether or not the cluster had successfully been able to set up its kubeconfig credentials (needed to create a `helper`), and also being created unnecessarily if no upgrade needed to occur.

This PR adjusts creation of the route monitors so that they only get created if:
- E2E has been able to obtain the cluster's kubeconfig
- E2E is running a test job to upgrade a cluster

So the aforementioned error won't show up when it shouldn't.

Has been successfully tested on an E2E upgrade to verify route monitors still get created when they should.

### DIFF
--- a/pkg/e2e/e2e.go
+++ b/pkg/e2e/e2e.go
@@ -338,39 +338,44 @@ func runGinkgoTests() error {
 	getLogs()
 	upgradeTestsPassed := true
 
-	var routeMonitorChan chan struct{}
-	closeMonitorChan := make(chan struct{})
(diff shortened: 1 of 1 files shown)


mrbarge:
Heck yes PR #666 :smiling_imp:

jeefy:
Cursed PR. 
/approve
/lgtm

PR #670, opened 2021-01-19
Added filtering of Before Suite errors into separate Junit Results 

agopala95:
Added a config file to keep track of Before Suite errors that crop up during osde2e testing and populate them all into a separate Junit file. More error checks can be added in the future if needed.


mrbarge:
Gave this a test-run and it seemed to work okay to me!

PR #671, opened 2021-01-26
Validate no nodes have the no schedule taint

jharrington22:
This change attempts to validate that no nodes in a cluster have the "NoSchedule" effect.


mrbarge:
Gave this a test. 
- Verified that on a cluster with healthy nodes it didn't trigger a healthcheck failure state. 
- Verified that on a cluster with a cordoned node it triggered a healthcheck failure state.

/lgtm

PR #673, opened 2021-01-27
add notification channel config to addon harness

whereswaldon:
This will simply list the harnesses that failed in slack. @jeefy mentioned perhaps providing links to review the test output (which would be cool), but I'm putting this up in its current state to get feedback about the high-level approach.

### DIFF
--- a/pkg/e2e/addons/addon_test_harness.go
+++ b/pkg/e2e/addons/addon_test_harness.go
@@ -1,12 +1,14 @@
 package addons
 
 import (
+	"fmt"
 	"log"
 	"strings"
 
 	"github.com/onsi/ginkgo"
 	"github.com/spf13/viper"
 
+	"github.com/openshift/osde2e/pkg/common/alert"
 	"github.com/openshift/osde2e/pkg/common/config"
 	"github.com/openshift/osde2e/pkg/common/helper"
 )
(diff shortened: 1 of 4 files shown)


pkg/e2e/addons/addon_test_harness.go
```diff
@@ -24,6 +26,11 @@ var _ = ginkgo.Describe("[Suite: addons] Addon Test Harness", func() {
 	}
 	ginkgo.It("should run until completion", func() {
 		h.SetServiceAccount(viper.GetString(config.Addons.TestUser))
-		h.RunAddonTests("addon-tests", int(addonTimeoutInSeconds), strings.Split(viper.GetString(config.Addons.TestHarnesses), ","), []string{})
+		harnesses := strings.Split(viper.GetString(config.Addons.TestHarnesses), ",")
+		failed := h.RunAddonTests("addon-tests", int(addonTimeoutInSeconds), harnesses, []string{})
+		if len(failed) > 0 {
+			// tests failed, notify
+			alert.SendSlackMessage(viper.GetString(config.Addons.SlackChannel), fmt.Sprintf("Addon tests failed: %v", failed))
```
jeefy, line 33: Can you either wrap this in a check to see if the Slack API token is set (or do it as part of SendSlackMessage)?

Currently for normal addon jobs we don't set it so once this is in we'll have to cut a PR against openshift/release to add the Slack token into it. :)

jeefy (changes requested):
Looks great! One minor change/thought.

PR #674, opened 2021-01-27
SDCICD-571: Reduce credentials and logging calls to OCM

jeefy:
Credentials calls 157 -> 1
Logging Calls 1497 -> 45

Seems good right? :)

### DIFF
--- a/pkg/common/providers/ocmprovider/cluster.go
+++ b/pkg/common/providers/ocmprovider/cluster.go
@@ -518,7 +518,13 @@ func (o *OCMProvider) ClusterKubeconfig(clusterID string) ([]byte, error) {
 		return getLocalKubeConfig(localKubeConfig)
 	}
 
+	existingKubeConfig := viper.GetString(config.Kubeconfig.Contents)
+	if existingKubeConfig != "" {
+		return []byte(existingKubeConfig), nil
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/common/providers/ocmprovider/cluster.go
```diff
@@ -518,7 +518,13 @@ func (o *OCMProvider) ClusterKubeconfig(clusterID string) ([]byte, error) {
 		return getLocalKubeConfig(localKubeConfig)
 	}
 
+	existingKubeConfig := viper.GetString(config.Kubeconfig.Contents)
+	if existingKubeConfig != "" {
+		return []byte(existingKubeConfig), nil
+	}
+
 	if creds, ok := o.credentialCache[clusterID]; ok {
+		viper.Set(config.Kubeconfig.Contents, creds)
```
jharrington22, line 527: How does this reduce calls to one? do you store the kubeconfig for every cluster?

jeefy:
We were supposed to cache the Kubeconfig for a cluster we had already
grabbed.

We had this credentialCache object that we used but there were other calls
that checked if a Kubeconfig existed using Viper.

This area needs more clean up but this functionally does what we need it to
and it's a pretty minor change. :)

PR #677, opened 2021-02-01
Before Suite Errors into Junit Results

agopala95:
Added more before suite errors into the config file to be filtered into junit results.


whereswaldon (approved):
Thanks for adding these! With the context I have, this looks great.

agopala95:
> Thanks for adding these! With the context I have, this looks great.

Sure. No problem :)

jeefy:
Hype!
/approve
/lgtm

PR #681, opened 2021-02-02
Enable osd-cluster-ready in-cluster healthcheck

2uasimojo:
This commit enables real health checks to be run from the [osd-cluster-ready Job](https://github.com/iamkirkbater/osd-cluster-ready-job) by:
- Tweaking `PollClusterHealth` to use in-cluster REST config if the `clusterID` is not specified (empty string).
- Ignoring the `osd-cluster-ready` pod itself, as it will be Pending while it's running the health check.


2uasimojo:
Required for https://github.com/iamkirkbater/osd-cluster-ready-job/pull/1

jharrington22 (commented):
/lgtm

PR #684, opened 2021-02-02
Test cases for PublishingStrategy's management of public/private ingress

blrm:
This implements two tests for the cloud-ingress-operator's management of the PublishingStrategy CR. It updates the default OSD PublishingStrategy to make the default application ingress private and waits until it can confirm that the correct service LoadBalancer object has been created. It then tests the same process for toggling back from private to public. 

The associated JIRA ticket: https://issues.redhat.com/browse/OSD-5337

### DIFF
--- a/pkg/e2e/operators/cloudingress/public_private.go
+++ b/pkg/e2e/operators/cloudingress/public_private.go
@@ -2,23 +2,20 @@ package cloudingress
 
 import (
 	"context"
-	"fmt"
 	"log"
-	"strings"
 	"time"
 
 	"github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	cloudingressv1alpha1 "github.com/openshift/cloud-ingress-operator/pkg/apis/cloudingress/v1alpha1"
-	"github.com/spf13/viper"
 
-	"github.com/openshift/osde2e/pkg/common/config"
 	"github.com/openshift/osde2e/pkg/common/constants"
 	"github.com/openshift/osde2e/pkg/common/helper"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 // tests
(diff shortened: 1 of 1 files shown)


pkg/e2e/operators/cloudingress/public_private.go
```diff
@@ -2,35 +2,128 @@ package cloudingress
...
+
+	ps.Object, err = runtime.DefaultUnstructuredConverter.ToUnstructured(&PublishingStrategyInstance)
+	Expect(err).NotTo(HaveOccurred())
+
+	// Update the publishingstrategy
+	ps, err = h.Dynamic().Resource(schema.GroupVersionResource{Group: "cloudingress.managed.openshift.io", Version: "v1alpha1", Resource: "publishingstrategies"}).Namespace(OperatorNamespace).Update(context.TODO(), ps, metav1.UpdateOptions{})
+	Expect(err).NotTo(HaveOccurred())
+}
+
+func pollAppIngressService(h *helper.H, serviceName string, svcNamespace string, lbscheme string) error {
```
cblecker, line 79: When doing polls, there are helper funcs for that:
https://github.com/openshift/osde2e/blob/d9e539bbd336ed12dc0e74d03f22b700fdf14e79/pkg/e2e/operators/cloudingress/apischeme_cr.go#L26-L34
https://pkg.go.dev/k8s.io/apimachinery@v0.20.2/pkg/util/wait#PollImmediate

cblecker (changes requested):
/hold

pkg/e2e/operators/cloudingress/public_private.go
blrm: Great! Didn't know about those. I've updated the tests to use them and removed the `pollAppIngressService` function.

pkg/e2e/operators/cloudingress/public_private.go
```diff
@@ -2,35 +2,100 @@ package cloudingress
...
-			tests := []struct {
-				Name string
-			}{}
+			updateApplicationIngress(h, "internal")
 
-			for _, test := range tests {
-				Expect(test.Name).To(Equal(""))
-			}
+			//wait for router-default service loadbalancer to have an annotation indicating its scheme is internal
+			err := wait.PollImmediate(10*time.Second, 5*time.Minute, func() (bool, error) {
```
cblecker, line 31: Is 5 mins the right deadline for this? What are the changes of this being racy with the 5 min deadline?

cblecker (commented):
one question, otherwise lgtm

pkg/e2e/operators/cloudingress/public_private.go
blrm: While testing this, I never saw it take anywhere near that long, generally less than a minute. Most of that time is spent waiting for the ingress pods to be terminated/recreated. I think its unlikely to be racy.

blrm:
@cblecker thanks! Can you remove the hold?

PR #687, opened 2021-02-04
Scc test to verify new Scc does not break pods

aliceh:
In this commit, I added a test that creates a new permissive Scc and then restarts the Prometheus and ssh pods. Next, the test verifies that the new pods came up as expected. This test was created to verify the fix for Bug 1868976 https://bugzilla.redhat.com/show_bug.cgi?id=1868976

### DIFF
--- a/pkg/e2e/verify/sccs.go
+++ b/pkg/e2e/verify/sccs.go
@@ -2,12 +2,19 @@ package verify
 
 import (
 	"context"
+	"errors"
+	"fmt"
+	"log"
+	"time"
+
 	"github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	v1 "github.com/openshift/api/security/v1"
 	"github.com/openshift/osde2e/pkg/common/alert"
 	"github.com/openshift/osde2e/pkg/common/helper"
+	apiv1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
-	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 var dedicatedAdminSccTestName = "[Suite: e2e] [OSD] RBAC Dedicated Admins SCC permissions"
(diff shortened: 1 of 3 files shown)


pkg/e2e/verify/sccs.go
```diff
@@ -68,3 +129,145 @@ func checkSccPermissions(h *helper.H, clusterRole string, scc string) {
...
+//Extracts pod names from a filtered list and counts how many are in running state
+func getPodNames(namespace string, list *apiv1.PodList, h *helper.H) ([]string, int) {
+	var notReady []apiv1.Pod
+	var ready []apiv1.Pod
+	var podNames []string
+	var total int
+	var numReady int
+	notReady = nil
+	ready = nil
+	podNames = nil
```
whereswaldon, line 200:
```suggestion
```
The zero value for a slice is already nil, so this is a no-op.

pkg/e2e/verify/sccs.go
```diff
@@ -68,3 +129,145 @@ func checkSccPermissions(h *helper.H, clusterRole string, scc string) {
...
+		phase := pod.Status.Phase
+		if phase != apiv1.PodRunning && phase != apiv1.PodSucceeded {
+			notReady = append(notReady, pod)
+		} else {
+			ready = append(ready, pod)
+
+		}
+		total = len(list.Items)
+		numReady = (total - len(notReady))
+	}
```
whereswaldon, line 214:
```suggestion
	}
	total = len(list.Items)
	numReady = (total - len(notReady))
```

I can't see any reason to recompute these values each loop iteration. Am I missing something?

pkg/e2e/verify/sccs.go
```diff
@@ -68,3 +129,145 @@ func checkSccPermissions(h *helper.H, clusterRole string, scc string) {
...
+		if err != nil {
+			return false, nil
+		}
+
+		// scale up
+		s, err = h.Kube().AppsV1().Deployments(ns).GetScale(context.TODO(), operator, metav1.GetOptions{})
+		if err != nil {
+			return false, nil
+		}
+		sc = *s
```
whereswaldon, line 245:
```suggestion
```
I have the sense that I'm probably just missing some k8s context here, but why do we need to fetch the scale again? We already have the version of it with zero replicas in the local variable `sc`. Can't we just update that one and apply it again?

whereswaldon (changes requested):
I have a few questions about the implementation, but it looks good at a high level. Nice work! What a gnarly bug.

aliceh:
Thanks for the useful suggestions, @whereswaldon !

whereswaldon (approved):

whereswaldon:
/approve

You're welcome! Looks good.

PR #690, opened 2021-02-05
Disable addon list on scale tests

agopala95:
Inserted a config flag check to disable listing of addons in clusters for a scale test during health check polling.

### DIFF
--- a/pkg/common/config/config.go
+++ b/pkg/common/config/config.go
@@ -357,7 +357,7 @@ var Addons = struct {
 	// Env: ADDON_SLACK_CHANNEL
 	SlackChannel string
 
-	// SkipAddonList is a boolean to indicate whether the listing of addons has to be disbaled or not.
+	// SkipAddonList is a boolean to indicate whether the listing of addons has to be disabled or not.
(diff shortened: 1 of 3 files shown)


jeefy (changes requested):
There's another instance of listing addons that was missed: https://github.com/openshift/osde2e/blob/main/pkg/common/providers/ocmprovider/cluster.go#L753-L771

This is on me, but I think a better approach now would be to add a completely separate config option/env var (`SKIP_ADDON_LIST`? I'm open to somethign different) and then check the config option instead of introspecting the cluster properties.

agopala95:
> There's another instance of listing addons that was missed: https://github.com/openshift/osde2e/blob/main/pkg/common/providers/ocmprovider/cluster.go#L753-L771
> 
> This is on me, but I think a better approach now would be to add a completely separate config option/env var (`SKIP_ADDON_LIST`? I'm open to somethign different) and then check the config option instead of introspecting the cluster properties.

I've made use of the env var/config option. Thanks for the tip.

whereswaldon (approved):

PR #691, opened 2021-02-05
SDCICD-574: Ensure supplied versions retain their channel suffix so CIS matches

jeefy:
Previously we had to strip the channel names from versions to match installed CISs.

Now CISs have the channel as a suffix and we support multiple channels. :)

### DIFF
--- a/pkg/common/versions/version_selector.go
+++ b/pkg/common/versions/version_selector.go
@@ -4,12 +4,15 @@ import (
 	"fmt"
 	"log"
 	"math"
+	"strings"
 
 	"github.com/Masterminds/semver"
+	"github.com/openshift/osde2e/pkg/common/config"
 	"github.com/openshift/osde2e/pkg/common/spi"
 	"github.com/openshift/osde2e/pkg/common/util"
 	"github.com/openshift/osde2e/pkg/common/versions/installselectors"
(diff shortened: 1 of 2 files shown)


pkg/common/versions/version_selector.go
```diff
@@ -33,6 +36,11 @@ func GetVersionForInstall(versionList *spi.VersionList) (*semver.Version, string
 
 	v, selector, err := selectedVersionSelector.SelectVersion(versionList)
 
+	channel := viper.GetString(config.Cluster.Channel)
+	if channel != "stable" && !strings.Contains(v.Original(), channel) {
+		v = semver.MustParse(fmt.Sprintf("%s-%s", v.Original(), channel))
```
RiRa12621, line 41: Doesn't that allow arbitrary input as long as it's semver formatted?

jharrington22 (approved):
/lgtm

pkg/common/versions/version_selector.go
jeefy: Arbitrary input for supplying the channel? Yes. But that goes for almost all of our config options.

RiRa12621: Was just curious if I understood correctly. Sorry if the comment was misleading

whereswaldon (approved):

PR #693, opened 2021-02-09
fix: use junit data to determine openshift-suite pass/fail status

whereswaldon:
This commit does several things:
- Fixes all of the tests to be configured as cluster admin
- Stops overriding the user account in use away from cluster admin
- Changes the way that we determine failure from exit status to
  parsing the junit XML. Unparsable xml, lack of xml, or failures
  in XML will cause the test to fail.

### DIFF
--- a/pkg/e2e/openshift/images.go
+++ b/pkg/e2e/openshift/images.go
@@ -23,6 +23,7 @@ var _ = ginkgo.Describe(imageRegistryTestName, func() {
 
 	e2eTimeoutInSeconds := 3600
 	ginkgo.It("should run until completion", func() {
+		h.SetServiceAccount("system:serviceaccount:%s:cluster-admin")
 		// configure tests
 		cfg := DefaultE2EConfig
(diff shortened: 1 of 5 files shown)


pkg/e2e/openshift/images.go
```diff
@@ -42,6 +78,11 @@ var _ = ginkgo.Describe(imageRegistryTestName, func() {
 		results, err := r.RetrieveResults()
```
jeefy, line 78: What are your thoughts on including both calling `ensurePassingXML` as well as checking that it's valid and passing as part of `r.RetrieveResults()` ?

I feel like that makes tests less redundant but I could see value in being explicit in each case.

whereswaldon: I think that's a good idea *if and only if* there is no usage of the helper that doesn't produce JUnit result data. If we always expect this, then we should absolutely do that.

jeefy:
/lgtm
What a pain

PR #704, opened 2021-02-13
to detect when hiding as an ENV VAR

dofinn:


pkg/e2e/e2e.go
```diff
@@ -743,7 +743,7 @@ func runTestsInPhase(phase string, description string) bool {
 		}
 		clusterState = cluster.State()
 	}
-	if !viper.GetBool(config.DryRun) && clusterState == spi.ClusterStateReady {
```
mrbarge, line 746: Any motivation for this being replaced in favour if it being an additional argument to the function? It could be left as it is and still behave the same.

dofinn: there are multiple sources of truth for this var. We add it to the ginkgo config explicity so we should refer to that here and elsewhere.

mrbarge: Fair enough, that makes sense

dofinn:
@mrbarge anything outstanding on this PR ?

PR #706, opened 2021-02-15
add informing to run suites that are marked as informing locally

boranx:
that might enable running informing suites locally with `GINKGO_FOCUS` flag


boranx:
ignore as there's already a _informing-suite.yaml_

PR #707, opened 2021-02-15
SDCICD-583: update latest filter to follow latest-y and latest-z filters

jeefy:
I didn't update the basic "latest" filter with the same logic I did y/z. :upside_down_face:

### DIFF
--- a/pkg/common/versions/upgradeselectors/latest_version_selector.go
+++ b/pkg/common/versions/upgradeselectors/latest_version_selector.go
@@ -30,15 +30,19 @@ func (l latestVersion) SelectVersion(installVersion *spi.Version, versionList *s
 
 	for _, v := range versionList.FindVersion(installVersion.Version().Original()) {
 		for upgradeVersion := range v.AvailableUpgrades() {
-			if upgradeVersion.Prerelease() != "" {
-				if strings.Contains(upgradeVersion.Prerelease(), "nightly") && upgradeVersion.GreaterThan(newestVersion.Version()) {
+			if strings.Contains(upgradeVersion.Original(), "nightly") && !strings.Contains(newestVersion.Version().Original(), "nightly") {
+				newestVersion = spi.NewVersionBuilder().Version(upgradeVersion).Build()
+				continue
+			}
+
+			// Catch the rest
+			if strings.Contains(upgradeVersion.Original(), "nightly") && strings.Contains(newestVersion.Version().Original(), "nightly") {
+				if upgradeVersion.Original() > newestVersion.Version().Original() {
 					newestVersion = spi.NewVersionBuilder().Version(upgradeVersion).Build()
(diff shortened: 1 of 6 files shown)


pkg/common/versions/upgradeselectors/latest_version_selector.go
```diff
@@ -30,15 +30,19 @@ func (l latestVersion) SelectVersion(installVersion *spi.Version, versionList *s
...
-			if upgradeVersion.Prerelease() != "" {
-				if strings.Contains(upgradeVersion.Prerelease(), "nightly") && upgradeVersion.GreaterThan(newestVersion.Version()) {
+			if strings.Contains(upgradeVersion.Original(), "nightly") && !strings.Contains(newestVersion.Version().Original(), "nightly") {
+				newestVersion = spi.NewVersionBuilder().Version(upgradeVersion).Build()
+				continue
+			}
+
+			// Catch the rest
+			if strings.Contains(upgradeVersion.Original(), "nightly") && strings.Contains(newestVersion.Version().Original(), "nightly") {
+				if upgradeVersion.Original() > newestVersion.Version().Original() {
```
whereswaldon, line 40: @jeefy, you taught me a new piece of the go programming language! I had no idea that strings supported the raw `<` and friends comparison operators.

That being said, I think this will do the wrong thing for comparisons like `4.5.9 < 4.5.11`, since `1 < 9`, I think this concludes that `4.5.9` is the newer version. Example: https://play.golang.org/p/4pIEji75qkQ

We should change the comparison to something like this:
```suggestion
				if upgradeVersion.GreaterThan(newestVersion.Version()) {
```

I believe this should actually respect the semver properly: https://pkg.go.dev/github.com/Masterminds/semver/v3#Version.GreaterThan


Actually, we do exactly the same thing in this if block as in the else block below it. I think we can combine them.

whereswaldon: I think we can change the body of the inner loop to:

```go
upgradeIsNightly := strings.Contains(upgradeVersion.Original(), "nightly")
newestIsNightly := strings.Contains(newestVersion.Version().Original(), "nightly")
if (upgradeIsNightly && !newestIsNightly) || upgradeVersion.GreaterThan(newestVersion.Version()) {
	newestVersion = spi.NewVersionBuilder().Version(upgradeVersion).Build()
}
```

whereswaldon (changes requested):
I'm glad you found the issue! I think we can probably simplify this some though.

pkg/common/versions/upgradeselectors/latest_version_selector.go
jeefy: This is so awesome, thank you!

whereswaldon (approved):

PR #710, opened 2021-02-16
[OSD-5910] Adding e2e test for HTTP strict transport security setting verification on osd managed routes

treymer:


clcollins (approved):
/lgtm

PR #719, opened 2021-02-19
Support for log alert thresholds and config per platform

agopala95:
Introduced config files for AWS and GCP to have platform specific log alerts.


pkg/common/config/log_metrics.go
```diff
@@ -23,14 +23,22 @@ var beforeSuiteMetrics = LogMetrics{}
 // GetLogMetrics will return the log metrics.
 func GetLogMetrics() LogMetrics {
 	once.Do(func() {
-		viper.UnmarshalKey("logMetrics", &logMetrics)
+		if CloudProvider.CloudProviderID == "gcp" {
```
jeefy, line 26: What if we removed the if statement and instead dynamically attempted to do UnmarshalKey? Like...

`err := viper.UnmarshalKey(fmt.Sprintf("%slogMetrics", CloudProvider.CloudProviderID)` and then if there's an error, log it and bubble it up?

jeefy (changes requested):
Values look good and initial work looks good! One quick thought/change :)

pkg/common/config/log_metrics.go
agopala95: Yup, this makes more sense. I've made the changes as suggested. Thanks!

jeefy:
Thanks for the changes!
/approve
/assign @whereswaldon

PR #723, opened 2021-02-19
SDCICD-580: support addon parameters

whereswaldon:


pkg/e2e/e2e.go
```diff
@@ -185,7 +185,12 @@ func writeLogs(m map[string][]byte) {
 // installAddons installs addons onto the cluster
 func installAddons() (err error) {
 	clusterID := viper.GetString(config.Cluster.ID)
-	num, err := provider.InstallAddons(clusterID, strings.Split(viper.GetString(config.Addons.IDs), ","))
+	params := make(map[string]map[string]string)
+	strParams := viper.GetString(config.Addons.Parameters)
+	if err := json.Unmarshal([]byte(strParams), &params); err != nil {
```
jeefy, line 190: I don't recall this off the top of my head. If we unmarshal an empty string does that me params stays empty? Or would that cause an error?

whereswaldon: So, you are right: https://play.golang.org/p/5R1d6kb725H

It causes an error if we unmarshal the empty string. That being said, the default value for this field is set to `{}`. Is that enough of a mitigation, or do you want me to add logic here as well?

jeefy: Nope that should be good!

PR #724, opened 2021-02-19
SDCICD-591: skip certain cluster checks in e2e when using the mock provider

jeefy:
This allows addon authors the ability to use/specify a Kubeconfig externally and test against any valid openshift cluster while using the `mock` provider. This isn't an ideal solution and, in the future, we need to create a "local" provider that expects a Kubeconfig and lets us remove all these conditional statements within the main e2e code. :)

### DIFF
--- a/pkg/e2e/e2e.go
+++ b/pkg/e2e/e2e.go
@@ -105,7 +105,7 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
 		viper.Set(config.CloudProvider.Region, cluster.Region())
 		log.Printf("CLOUD_PROVIDER_REGION set to %s from OCM.", viper.GetString(config.CloudProvider.Region))
 
-		if !viper.GetBool(config.Addons.SkipAddonList) {
+		if !viper.GetBool(config.Addons.SkipAddonList) || viper.GetString(config.Provider) != "mock" {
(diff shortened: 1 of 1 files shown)


pkg/e2e/e2e.go
```diff
@@ -105,7 +105,7 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
 		viper.Set(config.CloudProvider.Region, cluster.Region())
 		log.Printf("CLOUD_PROVIDER_REGION set to %s from OCM.", viper.GetString(config.CloudProvider.Region))
 
-		if !viper.GetBool(config.Addons.SkipAddonList) {
+		if !viper.GetBool(config.Addons.SkipAddonList) || viper.GetString(config.Provider) == "mock" {
```
whereswaldon, line 108: Why do this when the provider *is* the mock provider? It seems to me that you'd want the opposite

whereswaldon (changes requested):

pkg/e2e/e2e.go
jeefy: You are _so_ so right.

jeefy: Fixed! PTAL

whereswaldon (approved):

PR #725, opened 2021-02-22
SDCICD-387: Break out create/delete to separate scripts

jeefy:
This adds ocp create and delete helper scripts to be used in Prow and with the `osde2e-credentials` secret.

This also cleans up the adoption test logic. :tada:

### DIFF
--- a/ci/prow-ocm-adoption.sh
+++ b/ci/prow-ocm-adoption.sh
@@ -1,70 +1,66 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -o pipefail
 
-mkdir -p installer
+OCM_URL="https://api.stage.openshift.com"
 
-export VERSION=latest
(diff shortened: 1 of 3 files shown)


ci/prow-ocm-adoption.sh
```diff
@@ -1,77 +1,77 @@
...
-./openshift-install create cluster --dir=$(pwd)/installer/ --log-level info
-
-export OCM_CONFIG=$(pwd)/.ocm.json
-export KUBECONFIG=$(pwd)/installer/auth/kubeconfig
+chmod +x oc
 
-{
-
-go get -u github.com/openshift-online/ocm-cli/cmd/ocm
+go get -u github.com/openshift-online/ocm-cli/cmd/ocm@master
```
whereswaldon, line 45:
```suggestion
GO111MODULE=on go get github.com/openshift-online/ocm-cli/cmd/ocm@master
```

The `-u` flag (especially when used in non-module mode) is dangerous. It can install incorrect dependency versions, as it means "upgrade all dependencies of the target package to their highest possible version".

This formulation of the command should reliably install the expected code.

whereswaldon (changes requested):
I'd like to make shellcheck happy and to update some `go` CLI usage, but this otherwise looks great!

PR #727, opened 2021-02-23
Open low urgency pagerduty alerts when periodic builds fail

whereswaldon:


jeefy:
/lgtm
/hold
Holding till we have a PD token in place. :)

pkg/common/pagerduty/pagerduty.go
```diff
@@ -0,0 +1,34 @@
...
+
+// FireIncident attempts to create an alert indicating a failure in the provided
+// pipeline.
+func (pdc Config) FireAlert(pipeline, details string) error {
+	event := pd.V2Event{
+		RoutingKey: pdc.IntegrationKey,
+		Action:     "trigger",
+		Payload: &pd.V2Payload{
+			Summary:  pipeline + " failed",
+			Severity: "info",
```
jeefy, line 22: Excellent.gif

jeefy:
/lgtm
/hold cancel

We can merge this in without the token, it'll just log a message saying it can't hit PD.

PR #734, opened 2021-03-01
deps: stop vendoring dependencies

whereswaldon:
Vendoring just causes headache when someone forgets to do it, and
it doesn't buy us much. Go has excellent support for package integrity,
so the only thing vendoring accomplishes is giving us an offline copy
of our dependencies. We don't need that, so we can afford to stop
maintaining it.


whereswaldon:
I think prow is failing because it still thinks we're vendoring (it's using the scripts that I changed or removed during this PR). @jeefy can you sanity-check me?

jeefy:
Yeah merging this in fixes the tests. BIG RED BUTTON.

PR #740, opened 2021-03-05
Added timeout to pod webhook Test 1

aliceh:
Added timeout to make sure the cluster is ready for the test, in case selector sync sets needed more time.

### DIFF
--- a/pkg/e2e/verify/pod_webhook.go
+++ b/pkg/e2e/verify/pod_webhook.go
@@ -45,6 +45,7 @@ var _ = ginkgo.Describe(podWebhookTestName, func() {
 					"system:authenticated",
 				},
 			})
+			time.Sleep(5 * time.Second)
 			defer func() {
 				h.Impersonate(rest.ImpersonationConfig{})
 			}()


whereswaldon:
@aliceh Can you walk me through the rationale for this change?

pkg/e2e/verify/pod_webhook.go
```diff
@@ -45,6 +45,7 @@ var _ = ginkgo.Describe(podWebhookTestName, func() {
 					"system:authenticated",
 				},
 			})
+			time.Sleep(5 * time.Second)
```
jeefy, line 48: Is there a more deterministic way to prevent this failure than a sleep? :thinking:

aliceh:
@whereswaldon I had a suspicion that maybe a selector sync set needs more time for the sync and that's why the webhook was not blocking the dedicated admin user's actions correctly.

pkg/e2e/verify/pod_webhook.go
whereswaldon: @aliceh If your hypothesis RE: selector sync sets is true, could we instead directly query whether that sync operation is complete? Does the sync set have a status field that could tell us?

PR #742, opened 2021-03-09
SDCICD-603: Better handle retrieving and writing test artifacts

jeefy:
We should write results before evaluating errors, don't fail writing results when there's an error downloading from a directory

Also adjusting conformance timeouts to be in line with one another.


whereswaldon (approved):
/lgtm

PR #743, opened 2021-03-10
Use healthcheck cron job to determine cluster readiness

whereswaldon:
This is always failing with an error that indicates it can't find the cron job, but I can see it if I look manually. I'm pretty confused.

Here's an example of the error:

```
 cluster failed health check
  Unexpected error:
      <*fmt.wrapError | 0xc000565140>: {
          msg: "failed looking up job: the server could not find the requested resource (get jobs)",
          err: {
              ErrStatus: {
                  TypeMeta: {Kind: "", APIVersion: ""},
                  ListMeta: {
                      SelfLink: "",
                      ResourceVersion: "",
                      Continue: "",
                      RemainingItemCount: nil,
                  },
                  Status: "Failure",
                  Message: "the server could not find the requested resource (get jobs)",
                  Reason: "NotFound",
                  Details: {
                      Name: "",
                      Group: "",
                      Kind: "jobs",
                      UID: "",
                      Causes: [
                          {
                              Type: "UnexpectedServerResponse",
                              Message: "unknown",
                              Field: "",
                          },
                      ],
                      RetryAfterSeconds: 0,
                  },
                  Code: 404,
              },
          },
      }
      failed looking up job: the server could not find the requested resource (get jobs)
  occurred

```

### DIFF
--- a/pkg/common/cluster/clusterutil.go
+++ b/pkg/common/cluster/clusterutil.go
@@ -207,6 +207,10 @@ func waitForClusterReadyWithOverrideAndExpectedNumberOfNodes(clusterID string, l
 	return nil
 }
 
+// ClusterConfig returns the rest API config for a given cluster as well as the provider it
+// inferred to discover the config.
+// param clusterID: If specified, Provider will be discovered through OCM. If the empty string,
+// 		assume we are running in a cluster and use in-cluster REST config instead.
 func ClusterConfig(clusterID string) (restConfig *rest.Config, providerType string, err error) {
 	if clusterID == "" {
 		if restConfig, err = rest.InClusterConfig(); err != nil {
--- a/pkg/common/cluster/clusterutil.go
+++ b/pkg/common/cluster/clusterutil.go
@@ -207,6 +207,34 @@ func waitForClusterReadyWithOverrideAndExpectedNumberOfNodes(clusterID string, l
 	return nil
 }
 
+func ClusterConfig(clusterID string) (restConfig *rest.Config, providerType string, err error) {
+	if clusterID == "" {
+		if restConfig, err = rest.InClusterConfig(); err != nil {
+			return nil, "", fmt.Errorf("error getting in-cluster rest config: %w", err)
+		}
+
+		// FIXME: Is there a way to discover this from within the cluster?
+		// For now, ocm and rosa behave the same, so hardcode either.
+		providerType = "ocm"
+		return
+
+	}
+	provider, err := providers.ClusterProvider()
+
+	if err != nil {
+		return nil, "", fmt.Errorf("error getting cluster provisioning client: %w", err)
+	}
+	providerType = provider.Type()
+
+	restConfig, err = getRestConfig(provider, clusterID)
+	if err != nil {
+
+		return nil, "", fmt.Errorf("error generating rest config: %w", err)
+	}
+
+	return
+}
+
 // PollClusterHealth looks at CVO data to determine if a cluster is alive/healthy or not
 // param clusterID: If specified, Provider will be discovered through OCM. If the empty string,
 // 		assume we are running in a cluster and use in-cluster REST config instead.
@@ -215,33 +243,10 @@ func PollClusterHealth(clusterID string, logger *log.Logger) (status bool, failu
 
 	logger.Print("Polling Cluster Health...\n")
 
-	var restConfig *rest.Config
-	var providerType string
-
-	if clusterID == "" {
-		if restConfig, err = rest.InClusterConfig(); err != nil {
-			logger.Printf("Error getting in-cluster REST config: %v\n", err)
-			return false, nil, nil
-		}
-
-		// FIXME: Is there a way to discover this from within the cluster?
-		// For now, ocm and rosa behave the same, so hardcode either.
-		providerType = "ocm"
-
-	} else {
-		provider, err := providers.ClusterProvider()
-
-		if err != nil {
-			return false, nil, fmt.Errorf("error getting cluster provisioning client: %v", err)
-		}
-
-		restConfig, err = getRestConfig(provider, clusterID)
-		if err != nil {
-			logger.Printf("Error generating Rest Config: %v\n", err)
-			return false, nil, nil
-		}
-
-		providerType = provider.Type()
+	restConfig, providerType, err := ClusterConfig(clusterID)
+	if err != nil {
+		logger.Printf("Error getting cluster config: %v\n", err)
+		return false, nil, nil
 	}
 
 	kubeClient, err := kubernetes.NewForConfig(restConfig)
--- a/pkg/common/cluster/healthchecks/healthcheckjob.go
+++ b/pkg/common/cluster/healthchecks/healthcheckjob.go
@@ -12,8 +12,9 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
-// CheckHealthcheckJob uses the `osd-cluster-ready` healthcheck job to determine cluster readiness.
-func CheckHealthcheckJob(k8sClient *kubernetes.Clientset, ctx context.Context, logger *log.Logger) (bool, error) {
+// CheckHealthcheckJob uses the `osd-cluster-ready` healthcheck job to determine cluster readiness. If the cluster
+// is not ready, it will return an error.
+func CheckHealthcheckJob(k8sClient *kubernetes.Clientset, ctx context.Context, logger *log.Logger) error {
 	logger = logging.CreateNewStdLoggerOrUseExistingLogger(logger)
 
 	logger.Print("Checking whether cluster is healthy before proceeding...")
@@ -23,7 +24,7 @@ func CheckHealthcheckJob(k8sClient *kubernetes.Clientset, ctx context.Context, l
 	name := "osd-cluster-ready"
 	jobs, err := bv1C.Jobs(namespace).List(ctx, metav1.ListOptions{})
 	if err != nil {
-		return false, fmt.Errorf("failed listing jobs: %w", err)
+		return fmt.Errorf("failed listing jobs: %w", err)
 	}
 	for _, job := range jobs.Items {
 		if job.Name != name {
@@ -31,7 +32,7 @@ func CheckHealthcheckJob(k8sClient *kubernetes.Clientset, ctx context.Context, l
 		}
 		if job.Status.Succeeded > 0 {
 			log.Println("Healthcheck job has already succeeded")
-			return true, nil
+			return nil
 		}
 		log.Println("Healthcheck job has not yet succeeded, watching...")
 	}
@@ -40,7 +41,7 @@ func CheckHealthcheckJob(k8sClient *kubernetes.Clientset, ctx context.Context, l
 		FieldSelector:   "metadata.name=osd-cluster-ready",
 	})
 	if err != nil {
-		return false, fmt.Errorf("failed watching job: %w", err)
+		return fmt.Errorf("failed watching job: %w", err)
 	}
 	for {
 		select {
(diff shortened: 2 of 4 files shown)


whereswaldon:
I think this is correct, but I still want to test the negative case manually (make sure it blocks).
/hold

pkg/common/cluster/healthchecks/healthcheckjob.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	for {
+		select {
+		case event := <-watcher.ResultChan():
+			switch event.Type {
+			case watch.Added:
+				fallthrough
+			case watch.Modified:
+				job := event.Object.(*batchv1.Job)
+				if job.Status.Succeeded > 0 {
```
2uasimojo, line 53: I think we need to handle "completed but failed" as well.

...But how we handle it is different before vs after https://github.com/openshift/configure-alertmanager-operator/pull/143
- Before: Failed Job means you fail here.
- After: Failed Job will be deleted and reinstated (see below).

Need to think through this some more.

2uasimojo (commented):
Ugh, this has made me realize that https://github.com/openshift/configure-alertmanager-operator/pull/143 isn't going to be compatible with this solution. As currently written, it has code paths where it can declare the cluster ready due to age even if the osd-cluster-ready Job doesn't exist, or is still running, or is failed. Of those, I think "doesn't exist" is probably the only one e2e cares about.

I need to think through this some more. Worst case, we may end up deciding to leave this code as it was before :(

pkg/common/cluster/clusterutil.go
```diff
@@ -207,41 +207,46 @@ func waitForClusterReadyWithOverrideAndExpectedNumberOfNodes(clusterID string, l
 	return nil
 }
 
-// PollClusterHealth looks at CVO data to determine if a cluster is alive/healthy or not
-// param clusterID: If specified, Provider will be discovered through OCM. If the empty string,
-// 		assume we are running in a cluster and use in-cluster REST config instead.
```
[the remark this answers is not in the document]
whereswaldon: Sure, it wasn't removed intentionally, the original is just in the next function down in the file. It is appropriate both places though.

pkg/common/cluster/healthchecks/healthcheckjob.go
```diff
@@ -0,0 +1,67 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/kubernetes"
+)
+
+// CheckHealthcheckJob uses the `openshift-cluster-ready-*` healthcheck job to determine cluster readiness.
+func CheckHealthcheckJob(k8sClient *kubernetes.Clientset, ctx context.Context, logger *log.Logger) (bool, error) {
+	logger = logging.CreateNewStdLoggerOrUseExistingLogger(logger)
+
+	logger.Print("Checking that all Nodes are running or completed...")
```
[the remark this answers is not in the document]
whereswaldon: Yeah... Now that you mention it, this doesn't make a ton of sense. I'll reword it.

pkg/common/cluster/healthchecks/healthcheckjob.go
```diff
@@ -0,0 +1,67 @@
...
+
+	bv1C := k8sClient.BatchV1()
+	namespace := "openshift-monitoring"
+	name := "osd-cluster-ready"
+	jobs, err := bv1C.Jobs(namespace).List(ctx, metav1.ListOptions{})
```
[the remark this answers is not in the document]
whereswaldon: Sigh... It does. I just did it a dumb way because I too am not very familiar with client-go.

pkg/common/cluster/healthchecks/healthcheckjob.go
```diff
@@ -0,0 +1,67 @@
...
+			switch event.Type {
+			case watch.Added:
+				fallthrough
+			case watch.Modified:
+				job := event.Object.(*batchv1.Job)
+				if job.Status.Succeeded > 0 {
+					return true, nil
+				}
+			case watch.Deleted:
+				return false, fmt.Errorf("cluster readiness job deleted before becoming ready")
```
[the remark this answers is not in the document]
whereswaldon: If something went that wrong, I doubt that we'd want to proceed with testing, so I think that this behavior is probably okay.

pkg/common/cluster/healthchecks/healthcheckjob.go
whereswaldon: I'll wait to hear your updated thoughts on how this should work before altering this logic.

pkg/common/cluster/healthchecks/healthcheckjob.go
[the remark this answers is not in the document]
whereswaldon: Wait! On second thought, it *does* need to be a list. I need the resourceVersion field *of the list itself* in order to initiate the watch later. This allows me to detect the creation and deletion of jobs matching my criteria since I performed the list.

pkg/common/cluster/healthchecks/healthcheckjob.go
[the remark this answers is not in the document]
2uasimojo: Sorry, I wasn't very clear: a failing Job is not unusual early in the cluster's life. A common thing we used to see was prometheus not being up yet. Evictions can also be a cause. The point of having the Job owned by some kind of controller was so we could mitigate this kind of thing.

That said, we could keep quite a bit of that level of retry logic in the Job itself via `backoffLimit`/`activeDeadlineSeconds` rather than having the controller react to failure. That would allow both sides to consider a failed Job "fatal" in whatever sense is appropriate.

pkg/common/cluster/healthchecks/healthcheckjob.go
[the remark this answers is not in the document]
2uasimojo: Ah, I understand.

If you use a zero resourceVersion, do you even need the initial get/list?

pkg/common/cluster/healthchecks/healthcheckjob.go
2uasimojo: Talked this through with @jharrington22 just now. We're going to rework the readiness side so the picture is clear. Specific to this case, I think for osde2e's purposes "completed but failed" will translate to "cluster not ready, ain't ever gonna be".

But I'm also pretty sure we're going to use prometheus to carry that state... which would mean substantial changes in this PR. Stay tuned -- I hope to have something written up by tomorrow.

2uasimojo: Keep an eye on OSD-6646 starting [here](https://issues.redhat.com/browse/OSD-6646?focusedCommentId=15960917&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15960917)

2uasimojo: "Completed but failed" should be treated as "cluster not ready, ain't ever gonna be". I'll post more info in a top-level comment.

2uasimojo:
I think we can move forward with this, with changes as noted inline (and please let me go through the PR top-to-bottom one more time before we land it).

**Code:**
- We landed https://github.com/openshift/osd-cluster-ready/pull/7. With this, the osd-cluster-ready Job *only* does health checks.
- Of note: with the above PR, osd-cluster-ready no longer tries to talk to prometheus or alertmanager. These were two of the main causes of Pod failures (leading to restarts by the Job). As a result, the probability of the Job only needing to create one Pod is much higher than it was.
- It is still [deployed from managed-cluster-config](https://github.com/openshift/managed-cluster-config/tree/02178fa589313127ef975bd7f0b2c3e2c0aae5c6/deploy/osd-cluster-ready), and we're still relying on the Job spec -- specifically [`backoffLimit`](https://github.com/openshift/managed-cluster-config/blob/02178fa589313127ef975bd7f0b2c3e2c0aae5c6/deploy/osd-cluster-ready/60-osd-ready.Job.yaml#L7-L11) -- to make *sure* the thing eventually runs to completion.

**Logic:**
- If you see *no* osd-cluster-ready Job, you should wait for it to appear (subject to timeout -- see below).
- If you see the Job, and then it disappears, I want to know about it. This should never happen with the code as written. So I would say you should consider it a failure.
- If the Job is running, wait for it to finish (subject to timeout).
- If the Job fails, we're done trying; it will never succeed. You should consider this a failure.
- If the Job succeeds, we consider the cluster ready. This is your success case.
- Timeout: You should define a maximum amount of time before declaring the cluster dead. In c-am-o we're using 2h, configurable via an env var.

We landed https://github.com/openshift/configure-alertmanager-operator/pull/147 which uses the above logic, and ran it during a scale test this morning. It is working well.

pkg/common/cluster/healthchecks/healthcheckjob.go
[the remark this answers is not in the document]
2uasimojo: Update on this: For now, a deletion should never happen (except manually). So please consider it an immediate failure (as if the cluster will never be ready). Ideally, please ping me and let me have a poke at the cluster as well.

whereswaldon:
@2uasimojo I've updated the job handling logic to match your description. The timeout aspect is managed via the provided context, rather than being explicitly within the healthcheck code itself. If the context is cancelled (via a timeout), we shut the loop down.

pkg/e2e/e2e.go
```diff
@@ -121,12 +124,26 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
...
-		err = clusterutil.WaitForClusterReady(cluster.ID(), nil)
+		clusterConfig, _, err := clusterutil.ClusterConfig(cluster.ID())
+		if err != nil {
+			log.Printf("Failed looking up cluster config for healthcheck: %v", err)
+		}
+		kubeClient, err := kubernetes.NewForConfig(clusterConfig)
+		if err != nil {
+			log.Printf("Error generating Kube Clientset: %v\n", err)
+		}
+		ctx, cancel := context.WithTimeout(context.Background(), time.Hour*2)
```
2uasimojo, line 135: Did you want to make this timeout configurable? It is in [c-am-o](https://github.com/openshift/configure-alertmanager-operator/blob/bb029f42f2209a8bde284c7c81c53827b543d801/pkg/readiness/cluster_ready.go#L59-L62) FWIW. (This could be a separate PR.)

2uasimojo (commented):
Thanks @whereswaldon, this looks great.

Holding for other eyes if needed, feel free to unhold whenever.

/lgtm
/hold

whereswaldon:
/assign @jeefy
Can you take a look before we merge this?

pkg/e2e/e2e.go
whereswaldon: I thought about it, but I don't think we have a real use-case for configuring it. YAGNI?

2uasimojo: IIRC there's a zillion other config knobs around this, including the clean count, success sleep, and failure sleep. But I'm sure you're right -- are the other knobs ever used?

whereswaldon: @2uasimojo I've added a config knob for this and tested it.

whereswaldon:
The test failure seems to be a fluke. It didn't even start running the build as far as I can tell.

/retest

pkg/e2e/e2e.go
```diff
@@ -132,13 +132,18 @@ func beforeSuite() bool {
 		if err != nil {
 			log.Printf("Error generating Kube Clientset: %v\n", err)
 		}
-		ctx, cancel := context.WithTimeout(context.Background(), time.Hour*2)
+		duration, err := time.ParseDuration(viper.GetString(config.Tests.ClusterHealthChecksTimeout))
+		if err != nil {
+			log.Printf("Failed parsing health check timeout, using 2 hours: %v", err)
+			duration = time.Hour * 2
```
2uasimojo, line 138: Re-defaulting this will mask certain configuration and coding errors. IMHO it would be better to just blow up.

whereswaldon: Fair enough, It is done.

pkg/e2e/e2e.go
```diff
@@ -134,8 +134,8 @@ func beforeSuite() bool {
 		}
 		duration, err := time.ParseDuration(viper.GetString(config.Tests.ClusterHealthChecksTimeout))
 		if err != nil {
-			log.Printf("Failed parsing health check timeout, using 2 hours: %v", err)
-			duration = time.Hour * 2
+			log.Printf("Failed parsing health check timeout: %v", err)
+			return false
```
2uasimojo: Cool.

Latent, but it looks like the other two conditions above might be missing `return false`?

2uasimojo (commented):
/lgtm

pkg/e2e/e2e.go
whereswaldon: Yup, I am dumb. I had tunnel vision while making this change.

2uasimojo (commented):
/lgtm

2uasimojo:
Do y'all like to squash commits once a PR is ready?

whereswaldon:
I don't think we have a policy around that. I'm not going to worry about it for now.

/hold cancel

PR #744, opened 2021-03-11
feat: add tests for must gather operator (OSD-5853)

boranx:
Implements: https://issues.redhat.com/browse/OSD-5853


csheremeta:
/lgtm
tests pass

go.mod
```diff
@@ -54,8 +55,8 @@ require (
 	golang.org/x/net v0.0.0-20200707034311-ab3426394381
 	golang.org/x/sys v0.0.0-20201015000850-e3ed0017c211 // indirect
 	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776
-	k8s.io/api v0.19.2
-	k8s.io/apimachinery v0.19.2
+	k8s.io/api v0.20.0
+	k8s.io/apimachinery v0.20.0
```
whereswaldon, line 59: Did you intentionally bump this version? I don't think it will actually take effect, as the `replace` directive further down locks all of the k8s dependencies to v0.19.2.

whereswaldon (commented):
This looks good other than the dependency update. I get that may have been automated, but I'd still prefer to keep the version declared in the top of go.mod aligned with the one chosen by the `replace`s.

go.mod
boranx: hi @whereswaldon, yes, it was intentionally ended up after running `go mod tidy` since I'm using https://pkg.go.dev/github.com/redhat-cop/must-gather-operator/api/v1alpha1 to be able to create mustGatherCRs which has a dependency to _0.20.0_: https://github.com/redhat-cop/must-gather-operator/blob/v1.0.0/go.mod. We can still create the CR by directly applying a string as a whole which could be a bit unsafe but can prevent us from editing versions like this. wdyt?

boranx:
```
go mod tidy
go: finding module for package github.com/redhat-cop/must-gather-operator/api/v1alpha1
go: found github.com/redhat-cop/must-gather-operator/api/v1alpha1 in github.com/redhat-cop/must-gather-operator v1.0.0
```

whereswaldon: Hmm. Okay, well I did a local build to confirm, and it is indeed using `v0.19.2`. However, your code compiling and running, so I guess it's not a big deal. I don't think you need to change anything.

PR #748, opened 2021-03-15
docs: update addon testing onboarding documentation

whereswaldon:

### DIFF
--- a/docs/Addons.md
+++ b/docs/Addons.md
@@ -1,49 +1,22 @@
 # **Add-On Testing**
 
-## **Test Requirements**
+This document describes the requirements to configure E2E testing of an Addon within `osde2e`. This is only one part of the overall process of onboarding an addon to OSD. The full process is outlined in the documentation [available here](https://gitlab.cee.redhat.com/service/managed-tenants/-/tree/master).
 
-How an add-on is tested can vary between groups and projects. In light of this, there are a few requirements for an add-on test harness to be integrated into OSDe2e. The test should:
+## **The Structure of an Addon Test**
 
-*   Assume it is executing in a pod within an OpenShift cluster. This means once the test code is written, it needs to be packaged into a container image.
+How an add-on is tested can vary between groups and projects. In light of this, we've tried to create a very flexible and unopinionated framework for your testing. Your test harness should take the form of an OCI (docker) container that does the following:
+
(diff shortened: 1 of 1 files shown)


docs/Addons.md
```diff
@@ -1,71 +1,184 @@
...
 
+### CCS Cluster Testing
 
-### **Getting an OCM refresh token for your tests**
-
-You will need to request an OCM refresh token in order to run your tests. The easiest way to do this is to visit [https://cloud.redhat.com/openshift/token] and copy the OFFLINE_REFRESH_TOKEN. If you do not have an account or quota, please see [Managing Organization Quota]
+If you want to test a CCS (bring your own AWS account) cluster, you'll need to provide some additional details about your AWS account in a secret. In particular, you'll need to provide values for these environment variables (from your secret):
 
-Your account will need the following permissions:
+```
```
jeefy, line 168: These would be added in their Prow Secret, named like `ocm-aws-access-key` -- they aren't expected to set it as an Environment Variable for the job :)

docs/Addons.md
```diff
@@ -1,71 +1,184 @@
...
+      - /osde2e
+      env:
+      - name: ADDON_IDS
+        value: prow-operator
+      - name: ADDON_TEST_HARNESSES
+        value: quay.io/miwilson/prow-operator-test-harness
+      - name: CONFIGS
+        value: aws,stage,addon-suite
+      - name: SECRET_LOCATIONS
+        value: /usr/local/osde2e-common,/usr/local/osde2e-credentials
```
jeefy, line 120: We should explicitly call out that they are expected to replace `osde2e-credentials` since that is the SD-CICD account/quota.

docs/Addons.md
```diff
@@ -1,71 +1,184 @@
...
-- name: ADDON_TEST_USER
-  value: system:serviceaccount:%s:cluster-admin
-- name: ADDON_PARAMETERS
-  value: '{"prow-operator": {"foo": "bar"}}'
+An example prow job that configures the "prow" operator in the stage environment:
+
+```yaml
+- agent: kubernetes
+  cluster: build02
+  cron: 0 */12 * * *
```
jeefy, line 93: We should also call out that they are allowed/able to change the cron schedule :)

jeefy (changes requested):
Awesome work so far. Couple changes please :)

PR #750, opened 2021-03-16
SDCICD-603: Update conformance tests, improve error handling when downloading files

jeefy:
This changes the error handling to not stall any downloading when there's an error grabbing any files (I missed a spot from the other PR)

Additionally, this updates our conformance tests by adding similar logic to the upstream ci-step-registry.


whereswaldon (commented):
/lgtm

PR #751, opened 2021-03-16
SDCICD-595: Add additional logging into BeforeSuite, check for nil during version selection

jeefy:
Adding some verbose logging to conditions that may cause BeforeSuite to fail without any logging as to why.

Additionally, adding a check to validate that the OpenShift version string isn't nil, hopefully preventing other error conditions observed.

:tada:

### DIFF
--- a/pkg/common/events/events.go
+++ b/pkg/common/events/events.go
@@ -1,6 +1,8 @@
 package events
 
 import (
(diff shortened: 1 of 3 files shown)


whereswaldon (commented):
/lgtm

PR #755, opened 2021-03-19
[samesite] adding cluster version verification for samesite cookie support

treymer:


pkg/e2e/verify/samesite_cookie.go
```diff
@@ -24,13 +27,30 @@ var _ = ginkgo.Describe(samesiteTestName, func() {
...
+
 		ginkgo.It("should be set for openshift-monitoring OSD managed routes", func() {
+			clusterVersion, majMinVersion, err := getClusterVersion(h)
+			Expect(err).NotTo(HaveOccurred(), "failed getting cluster version")
+			Expect(clusterVersion).NotTo(BeNil())
+
+			if majMinVersion < supportVersion {
+				ginkgo.Skip("skipping due to unsupported cluster version. Must be >=4.6.0")
+			}
+
```
whereswaldon, line 39: Since this exact same code block is repeated twice, would it be possible to refactor it into a single function. Might work well as a closure defined above the `ginkgo.It` calls...

pkg/e2e/verify/samesite_cookie.go
```diff
@@ -52,3 +72,18 @@ func managedRoutes(h *helper.H, namespace string) (bool, error) {
 	}
 	return samesiteExists, nil
 }
+
+func getClusterVersion(h *helper.H) (*v1.ClusterVersion, int, error) {
```
whereswaldon, line 76: Would you consider adding this as a method to `helper.H` instead and exporting it? I could see other `osde2e` users wanting to be able to do this.

whereswaldon (changes requested):
Looks solid, well done! I think there are some small opportunities to DRY it up and make other testers' lives easier that I'd like to talk about before we call it done.

pkg/e2e/verify/samesite_cookie.go
treymer: Yeah, I originally attempted to create a function and ginkgo was skipping the second test for some reason. Closures are new to me, I will check that out and test.

pkg/e2e/verify/samesite_cookie.go
treymer: Absolutely, that is a great idea. I will explore this.

jeefy:
/approve
Strong +1 to this. Super excited to also see adding support for other tests to consume it as well!

treymer:
Put this test back into informing for now since it is a major functional change to this test I want to ensure it bakes there for a couple days and then I will promote back to e2e.

treymer:
@whereswaldon @jeefy Added GetClusterVersion to helper and DRY'd up my code w/ a closure function.

pkg/common/helper/helper.go
```diff
@@ -340,3 +341,14 @@ func (h *H) InspectState() {
...
+
+// GetClusterVersion returns the Cluster Version object and current cluster version ex. 4.7.1 as string
+func (h *H) GetClusterVersion() (*configv1.ClusterVersion, string, error) {
+	cfgClient := h.Cfg()
+	getOpts := metav1.GetOptions{}
+	clusterVersionObj, err := cfgClient.ConfigV1().ClusterVersions().Get(context.TODO(), "version", getOpts)
+	if err != nil {
+		return nil, "", fmt.Errorf("couldn't get current ClusterVersion '%s': %v", "version", err)
+	}
+	return clusterVersionObj, clusterVersionObj.Status.Desired.Version, nil
```
whereswaldon, line 353: I think this might be kinda pedantic and awful of me, but I don't think I understand why it's worth adding another return value here when you're already returning the struct containing that value as the first return value. I can kinda see that the string of nested field accesses is long and it saves you from doing that in other code, but I'm not sure that I think the tradeoff is worth it.

I won't block on this, but I'd recommend dropping back to two return values.

pkg/e2e/verify/samesite_cookie.go
```diff
@@ -21,43 +20,54 @@ const (
...
+func verifyVersion(h *helper.H) func() bool {
+	return func() bool {
+		unsupportedVersion := false
+		clusterVersionObj, currentVersion, err := h.GetClusterVersion()
+		Expect(err).NotTo(HaveOccurred(), "failed getting cluster version")
+		Expect(clusterVersionObj).NotTo(BeNil())
+
+		// Get the cluster version and slice it, then convert the major/minor version to int Ex. majMinVersion := 46
+		splitVersion := strings.Split(currentVersion, ".")
+		majMinVersion, err := strconv.Atoi(splitVersion[0] + splitVersion[1])
```
whereswaldon, line 61: It's always been this way, so I guess I really should have paid more attention last time, but this can panic if the current version isn't what we expect (a semver). Is there any realistic chance of that happening? Feels like we should defensively do something like:
```suggestion
		if len(splitVersion) < 2 {
			return true // not supported
		}
		majMinVersion, err := strconv.Atoi(splitVersion[0] + splitVersion[1])
```

whereswaldon (commented):
Thanks for that refactor! I have two more thoughts on this codebase. You don't need to change anything if you're happy with how things stand, but I'd at least like to hear what you think before giving my approval.

pkg/common/helper/helper.go
treymer: Fair enough, keeping the business logic down stream makes sense due to unintended increases in time complexity. I will adjust :)

pkg/e2e/verify/samesite_cookie.go
treymer: While I think it would be unlikely, I am new to the team in general so I am unsure if a non semver could make its way into the API. Either way, likely worth the change to add a catch. :+1:

treymer:
> Thanks for that refactor! I have two more thoughts on this codebase. You don't need to change anything if you're happy with how things stand, but I'd at least like to hear what you think before giving my approval.

I agree with the suggested changes. I will make the changes and push a new commit here shortly :smile:

treymer:
@whereswaldon changes have been made and tested. Let me know what you think! :smile:

pkg/e2e/verify/samesite_cookie.go
```diff
@@ -52,15 +52,22 @@ var _ = ginkgo.Describe(samesiteTestName, func() {
...
+		clusterVersionObj, err := h.GetClusterVersion()
 		Expect(err).NotTo(HaveOccurred(), "failed getting cluster version")
 		Expect(clusterVersionObj).NotTo(BeNil())
 
 		// Get the cluster version and slice it, then convert the major/minor version to int Ex. majMinVersion := 46
-		splitVersion := strings.Split(currentVersion, ".")
+		splitVersion := strings.Split(clusterVersionObj.Status.Desired.Version, ".")
+
+		// Assume the len is <= 3 since semver major/minor could be something like 4.11.x
+		if len(splitVersion) > 3 && len(splitVersion) != 0 {
```
whereswaldon, line 63: I don't think this handles the (unlikely) case of len == {1,2,3}. 3 and 2 both don't cause panics, but 1 does.

whereswaldon:
Mostly this looks great! I think there's still an edge case in one place though.

pkg/e2e/verify/samesite_cookie.go
treymer: Good point, let me make it simpler.

treymer: @whereswaldon I can just do this:
`		if len(splitVersion) != 2 || len(splitVersion) != 3 {
			return true
		}`
This way we explicitly expect a len or 2 or 3, anything else is unsupported.

treymer:
@whereswaldon This change should account for {0,1} and >3 so we should always expect either {2,3}. Let me know what you think. Thanks!

pkg/e2e/verify/samesite_cookie.go
```diff
@@ -16,28 +17,72 @@ const (
...
+					return true
+				}
+			}
+
+			if len(splitVersion) == 3 {
+				majMinVersion, err := strconv.Atoi(splitVersion[0] + splitVersion[1] + splitVersion[2])
+				Expect(err).NotTo(HaveOccurred(), "failed normalizing major/minor version %v to integer %v", majMinVersion, err)
+				if majMinVersion < supportVersion {
+					return true
+				}
```
whereswaldon, line 79: So if `supportVersion` is `46`, and we have a cluster at version `4.7.11`, we end up comparing `4711 < 46`. Is that the intention? In general, integer comparison here might be problematic once the minor/patch versions have multiple digits? I don't know the constraints that you're working under here, so perhaps there's a reason that you're not worried about that?

I guess I should ask: What is the goal of this function? What does it verify about the version? What does a true or false return value *mean*?

treymer: SameSite cookie being enables on routes in only supported in OSD versions >4.6. So, I am trying to verify the version here. In cases for version 5.11 for example I want 511 < 46 to fail. That is the intention here. Comparing integers is the way I am able to use conditionals in this sense to verify versions. I cannot simply look for versions >4.6 because 5.11 won't pass. So, this is the only way I can think of doing it. If you have other suggestions, then I am all ears.

This is why I check the length to be either 2 or 3 to account for multiple digits in the minor version (5.11) for example. Also, keep in mind I am splitting off the patch version, because I do not care about that in this case.

treymer: I actually do not need the conditional to check if len is 3, so I will remove that.

treymer: I can clean it up to this:
`if len(splitVersion) == 2 {
			majMinVersion, err := strconv.Atoi(splitVersion[0] + splitVersion[1])
			Expect(err).NotTo(HaveOccurred(), "failed normalizing major/minor version %v to integer %v", majMinVersion, err)
			if majMinVersion < supportVersion {
				return true
			} else {
				return false
			}
		}`

pkg/e2e/verify/samesite_cookie.go
```diff
@@ -27,59 +29,57 @@ var _ = ginkgo.Describe(samesiteTestName, func() {
...
+func verifyVersionSupport(h *helper.H) func() bool {
 	return func() bool {
 		clusterVersionObj, err := h.GetClusterVersion()
 		Expect(err).NotTo(HaveOccurred(), "failed getting cluster version")
 		Expect(clusterVersionObj).NotTo(BeNil())
 
-		// Get the cluster version and slice it, then convert the major/minor version to int Ex. majMinVersion := 46
-		splitVersion := strings.Split(clusterVersionObj.Status.Desired.Version, ".")
+		//splitVersion := strings.Split(clusterVersionObj.Status.Desired.Version, ".")
+		splitVersion := []string{"3", "5", "11"}
```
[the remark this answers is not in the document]
treymer: You beat me to it! :laughing:

treymer:
@whereswaldon ready for your review. I tested against all use cases discussed here.

whereswaldon (commented):
/lgtm

PR #757, opened 2021-03-24
PV test/ OSD-6467

aliceh:
In this commit, per [OSD-6467](https://issues.redhat.com/browse/OSD-6467), I added a test that lists all available storage classes on the cluster,  and then for each storage class creates a pod with a volume of that class attached. The test also creates a file in the volume to verify that it was attached correctly.

### DIFF
--- a/pkg/e2e/verify/storage.go
+++ b/pkg/e2e/verify/storage.go
@@ -93,7 +93,7 @@ var _ = ginkgo.Describe(storageTestName, func() {
 	})
 })
 
-//Extracts Storage Class names and
+// Get Storage Class names and cloud provider
 func getScNames(list *v1.StorageClassList, h *helper.H) ([]string, string) {
 	var scs []string
 	var provider string
--- a/pkg/e2e/verify/storage.go
+++ b/pkg/e2e/verify/storage.go
@@ -189,7 +189,7 @@ func createPVC(h *helper.H, ns string, pvc *corev1.PersistentVolumeClaim) (*core
 	return pvc, nil
 }
 
-// This function Deletes the PVC.
+// deletePersistentVolumeClaim deletes the PVC.
 func deletePersistentVolumeClaim(h *helper.H, pvcName string, ns string) error {
 	if h != nil && len(pvcName) > 0 {
 		log.Printf("Deleting PersistentVolumeClaim %q", pvcName)
--- a/pkg/e2e/verify/storage.go
+++ b/pkg/e2e/verify/storage.go
@@ -282,6 +282,7 @@ func createTestPod(h *helper.H, namespace string, pvclaims []*corev1.PersistentV
 // the pod has already reached completed state.
 var errPodCompleted = fmt.Errorf("pod ran to completion")
 
+// podRunning checks if pod is running 
 func podRunning(h *helper.H, podName, namespace string) wait.ConditionFunc {
 	return func() (bool, error) {
 		pod, err := h.Kube().CoreV1().Pods(namespace).Get(context.TODO(), podName, metav1.GetOptions{})
--- a/pkg/e2e/verify/storage.go
+++ b/pkg/e2e/verify/storage.go
@@ -2,14 +2,22 @@ package verify
 
 import (
 	"context"
+	"fmt"
+	"log"
+	"strings"
+	"time"
 
 	"github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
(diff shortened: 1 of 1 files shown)


pkg/e2e/verify/storage.go
```diff
@@ -18,17 +26,284 @@ func init() {
...
+		}, 300)
+	})
 })
+
+//Extracts Storage Class names and
+func getScNames(list *v1.StorageClassList, h *helper.H) ([]string, string) {
+	var scs []string
+	var provisioner string
+	provisioner = ""
+	provider := "gcp"
```
dustman9000, line 100: i'd think we'd want to default to "aws" here?

pkg/e2e/verify/storage.go
```diff
@@ -18,17 +26,284 @@ func init() {
...
+	for _, sc := range list.Items {
+		name := sc.ObjectMeta.Name
+		if provisioner == "" {
+			provisioner = sc.Provisioner
+			if strings.Contains(provisioner, "aws") {
```
dustman9000, line 105: Can we check for gcp as the corner case instead?

pkg/e2e/verify/storage.go
```diff
@@ -18,17 +26,284 @@ func init() {
...
+			GenerateName: "pvc-tester-",
+			Namespace:    ns,
+		},
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
+				{
+					Name:    "dummy",
+					Image:   "registry.access.redhat.com/ubi8/ubi-minimal",
+					Command: []string{"/bin/sh"},
+					Args:    []string{"-c", "touch /mnt/volume1/hello.txt; sleep 3600"},
```
dustman9000, line 238: can we echo a string to the file and read the string back? this would be a simple write/read test of the file system...

pkg/e2e/verify/storage.go
aliceh: I added that change, thanks for the suggestion @dustman9000.

pkg/e2e/verify/storage.go
aliceh: I added that change, thanks for the suggestion @dustman9000!

pkg/e2e/verify/storage.go
aliceh: I  removed the default provider, now we don't default to anything.

pkg/e2e/verify/storage.go
```diff
@@ -18,17 +26,285 @@ func init() {
...
+			for _, sc := range scList.Items {
+				Expect(sc.AllowVolumeExpansion).To(Not(BeNil()))
+				Expect(*sc.AllowVolumeExpansion).To(BeTrue())
+			}
+
+		}, 300)
+	})
 })
+
+//Extracts Storage Class names and
```
dustman9000, line 96:
```suggestion
// Get Storage Class names and cloud provider
```

pkg/e2e/verify/storage.go
```diff
@@ -18,17 +26,285 @@ func init() {
...
+// createPVC creates the PVC resource. Fails test on error.
+func createPVC(h *helper.H, ns string, pvc *corev1.PersistentVolumeClaim) (*corev1.PersistentVolumeClaim, error) {
+	pvc, err := h.Kube().CoreV1().PersistentVolumeClaims(ns).Create(context.TODO(), pvc, metav1.CreateOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("PVC Create API error: %v", err)
+	}
+	return pvc, nil
+}
+
+// This function Deletes the PVC.
```
dustman9000, line 192:
```suggestion
// deletePersistentVolumeClaim deletes the PVC.
```

pkg/e2e/verify/storage.go
```diff
@@ -18,17 +26,285 @@ func init() {
...
+		return pod, fmt.Errorf("pod Get API error: %v", err)
+	}
+	return pod, nil
+}
+
+// errPodCompleted is returned by PodRunning to indicate that
+// the pod has already reached completed state.
+var errPodCompleted = fmt.Errorf("pod ran to completion")
+
+func podRunning(h *helper.H, podName, namespace string) wait.ConditionFunc {
```
dustman9000, line 285:
```suggestion
// podRunning checks if pod is running 
func podRunning(h *helper.H, podName, namespace string) wait.ConditionFunc {
```

PR #759, opened 2021-03-25
Modified pending pods health check

agopala95:
Added a few changes on reviewing the PR comments for the initial pull request on the same -  https://github.com/openshift/osde2e/pull/737.

Changes made:-
1. The key in the pod error counter has been changed from pod name to pod UID.
2. The pod error counter deletes pod entries once a pod changes from "pending" state.
3. The healthcheck only passes if the pod error counter is entirely empty.


whereswaldon (commented):
/lgtm

PR #760, opened 2021-03-25
SDCICD-595: Refactor out synchronizedbeforesuite to a new function

jeefy:
We were getting errors when attempting to run Ginkgo a second time post-Upgrades. Since the use case for SynchronizedBeforeSuite is not needed anymore, we should remove it. 

Ran a full set of upgrade tests locally after this change and things came back solid. 

During debugging, also upgraded `gomega` and `ginkgo` packages.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -52,8 +52,7 @@ require (
 	github.com/tsenart/go-tsz v0.0.0-20180814235614-0bd30b3df1c3 // indirect
 	github.com/tsenart/vegeta v12.7.0+incompatible
 	github.com/vmware-tanzu/velero v1.5.0-beta.1.0.20200831161009-1dcaa1bf7512
-	golang.org/x/net v0.0.0-20200707034311-ab3426394381
-	golang.org/x/sys v0.0.0-20201015000850-e3ed0017c211 // indirect
+	golang.org/x/net v0.0.0-20201202161906-c7110b5ffcbb
 	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776
 	k8s.io/api v0.20.0
 	k8s.io/apimachinery v0.20.0
--- a/pkg/common/events/events.go
+++ b/pkg/common/events/events.go
@@ -2,8 +2,6 @@ package events
 
 import (
 	"log"
-
-	"github.com/onsi/gomega"
 )
 
 // Events records individual events that occur during the execution of osde2e
@@ -23,8 +21,8 @@ func initializeEvents() {
 	Instance.Events = map[string]bool{}
 }
 
-// HandleErrorWithEvents returns a gomega assertion and records events depending on the error state.
-func HandleErrorWithEvents(err error, successEvent EventType, failEvent EventType) gomega.Assertion {
+// HandleErrorWithEvents records events depending on the error state.
+func HandleErrorWithEvents(err error, successEvent EventType, failEvent EventType) {
 	if err != nil {
 		log.Printf("Fail event: %v", failEvent)
 		RecordEvent(failEvent)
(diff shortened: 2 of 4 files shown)


pkg/common/events/events.go
```diff
@@ -24,15 +22,14 @@ func initializeEvents() {
 }
 
 // HandleErrorWithEvents returns a gomega assertion and records events depending on the error state.
```
whereswaldon, line 24: Can we update the docs for this function given that we changed the signature?

pkg/e2e/e2e.go
```diff
@@ -81,19 +80,16 @@ var _ = ginkgo.BeforeEach(func() {
 	}
 })
 
-// Setup cluster before testing begins.
-var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
-	defer ginkgo.GinkgoRecover()
-
+func beforeSuite() bool {
```
whereswaldon, line 83: Can we document the semantics of the boolean return value?

pkg/e2e/e2e.go
```diff
@@ -241,8 +233,8 @@ func runGinkgoTests() error {
 	gomega.RegisterFailHandler(ginkgo.Fail)
 
 	ginkgoConfig.DefaultReporterConfig.NoisySkippings = !viper.GetBool(config.Tests.SuppressSkipNotifications)
-	ginkgoConfig.GinkgoConfig.SkipString = viper.GetString(config.Tests.GinkgoSkip)
-	ginkgoConfig.GinkgoConfig.FocusString = viper.GetString(config.Tests.GinkgoFocus)
+	ginkgoConfig.GinkgoConfig.SkipStrings = []string{viper.GetString(config.Tests.GinkgoSkip)}
```
whereswaldon, line 236: Is this an unrelated bugfix?

go.mod
```diff
@@ -24,8 +24,8 @@ require (
 	github.com/joshdk/go-junit v0.0.0-20201221202203-061ee62ada40
 	github.com/kylelemons/godebug v1.1.0
 	github.com/markbates/pkger v0.17.1
-	github.com/onsi/ginkgo v1.14.1
-	github.com/onsi/gomega v1.10.2
+	github.com/onsi/ginkgo v1.15.0
+	github.com/onsi/gomega v1.11.0
```
whereswaldon, line 28: Why update these deps? Was there a feature we needed in there?

whereswaldon (changes requested):
I have a bunch of little questions, but nothing major jumped out.

go.mod
jeefy: There wasn't, this will be removed.

pkg/e2e/e2e.go
jeefy: This was related to upgrading Ginkgo versions and will be removed.

jeefy:
@whereswaldon PTAL :)

pkg/e2e/e2e.go
```diff
@@ -650,6 +644,13 @@ func runTestsInPhase(phase string, description string, dryrun bool) bool {
 	phaseReporter := ginkgorep.NewPhaseReporter(phase, phaseReportPath)
 	ginkgoPassed := false
 
+	if !dryrun || ginkgoConfig.GinkgoConfig.DryRun {
+		if result := beforeSuite(); result == false {
```
whereswaldon, line 648:
```suggestion
		if !beforeSuite() {
```

pkg/e2e/e2e.go
```diff
@@ -650,6 +644,13 @@ func runTestsInPhase(phase string, description string, dryrun bool) bool {
 	phaseReporter := ginkgorep.NewPhaseReporter(phase, phaseReportPath)
 	ginkgoPassed := false
 
+	if !dryrun || ginkgoConfig.GinkgoConfig.DryRun {
```
whereswaldon, line 647: So, we only want this to happen if it's *not* a dryrun, right? Shouldn't it be:

```suggestion
	if !dryrun || !ginkgoConfig.GinkgoConfig.DryRun {
```

whereswaldon (commented):
/lgtm

PR #761, opened 2021-03-25
SDCICD-459: Add install_config option, randomize imagecontentsource

jeefy:
This allows us to specify/use the `install_config` option within OCM now. It also allows us to specify an ImageContentSource as a config (with several options pre-populated as a key-value pair in the ocmprovider.)

If no Image Source is specified, it will default to a random entry within the ocmprovider map. 

Note: This is only applicable in Int/Stage currently.

### DIFF
--- a/pkg/common/providers/ocmprovider/cluster_image_sources.go
+++ b/pkg/common/providers/ocmprovider/cluster_image_sources.go
@@ -1,0 +1,73 @@
+package ocmprovider
+
+import (
+	"log"
+	"math/rand"
+)
+
+var clusterImageSources = map[string]string{"standard": `imageContentSources:
+- mirrors:
+  - quay.io/openshift-release-dev/ocp-release
+  - 950916221866.dkr.ecr.us-east-1.amazonaws.com/ocp-release
+  source: quay.io/openshift-release-dev/ocp-release
+- mirrors:
+  - quay.io/openshift-release-dev/ocp-v4.0-art-dev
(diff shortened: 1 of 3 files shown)


pkg/common/providers/ocmprovider/cluster_image_sources.go
```diff
@@ -0,0 +1,73 @@
...
+	} else {
+		if source, ok = clusterImageSources[choice]; !ok {
+			log.Printf("Image source not found: %s", choice)
+			return ""
+		}
+	}
+
+	log.Printf("Choice: %s", choice)
+	log.Println(source)
+	return source
```
whereswaldon, line 72:
```suggestion
	if choice == "random" || choice == "" {
		var sources []string
		for key := range clusterImageSources {
			sources=append(sources, key)
		}
		choice = sources[rand.Intn(len(sources))]
	}
	if source, ok = clusterImageSources[choice]; !ok {
		log.Printf("Image source not found: %s", choice)
		return ""
	}

	log.Printf("Choice: %s", choice)
	log.Println(source)
	return source
```
I think this is equivalent, but a little easier to follow. Please point out my mistakes :D

whereswaldon (changes requested):
I think there's an opportunity to make the random choice code easier to understand.

jeefy:
Completely agree. Once verify completes I'm going to manually-merge this in post-changes so the PR can soak over the weekend. :)

PR #763, opened 2021-03-29
docs: update to ccs-only addon testing

whereswaldon:
Do not merge this until we get confirmation from leadership that all addon testing should be CCS-based.

/hold

### DIFF
--- a/docs/Addons.md
+++ b/docs/Addons.md
@@ -36,6 +36,8 @@ Please bump the quota for SKU `MW00530` by 2 so that we can provision additional
 
 ### Providing Secrets to Your Build
 
+If you are not a part of the public GitHub Organization `OpenShift`, join it by following [these instructions](https://source.redhat.com/groups/public/atomicopenshift/atomicopenshift_wiki/setting_up_your_accounts_openshift).
+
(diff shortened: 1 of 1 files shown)


docs/Addons.md
```diff
@@ -18,86 +18,33 @@ The [Prow Operator Test] is a good example of a [Basic operator test]. It verifi
...
-
-```yaml
----
-$schema: /user-1.yaml
-
-user_id: "your-ocm-user"
+### Providing Secrets to Your Build
 
-kerberos_id: "N/A"
+Follow the documentation [here](https://docs.ci.openshift.org/docs/how-tos/adding-a-new-secret-to-ci/) to create secrets and configure them to be mirrored into the `ci` namespace [like ours](https://github.com/openshift/release/blob/master/core-services/secret-mirroring/_mapping.yaml#L62).
```
k-wall, line 39: Following the link, I read "All users who wish to interact with the system must log in to the api.ci cluster", following the console link and chosing "git hub" as the identity provider, I see "An authentication error occurred.".   Could the doc make it clear how an addon author would go about getting permission?

k-wall, line 39: Is there a naming convention for this secret?

whereswaldon: Ehh, not really. Most people use `<addon-name>-credentials`

whereswaldon: I've pushed a change adding instructions for joining the OpenShift organization, which I believe is required for this.

whereswaldon:
This is the direction we're going.
/hold cancel

PR #764, opened 2021-03-30
fix: Don't set install_config for addon jobs

jeefy:
Don't set install_config in prod or if a job name has "addon" in it. :)


whereswaldon (approved):
/lgtm

PR #765, opened 2021-03-30
SDCICD-459: Add fourth test case for icsp's

jeefy:
This adds "ecr-primary" as a test case and changes "standard" to be "quay-primary"


whereswaldon (commented):
/lgtm

PR #769, opened 2021-04-01
fix: remove # character from slack channel name

whereswaldon:
This symbol isn't actually part of the channel name, and causes this notification attempt to always fail.


jeefy:
/lgtm
Or whatever
Github is hard.

PR #770, opened 2021-04-01
feat: merge PD incidents automatically

whereswaldon:

### DIFF
--- a/pkg/common/pagerduty/pagerduty.go
+++ b/pkg/common/pagerduty/pagerduty.go
@@ -2,6 +2,8 @@ package pagerduty
 
 import (
 	"fmt"
+	"log"
+	"sort"
 
 	pd "github.com/PagerDuty/go-pagerduty"
 )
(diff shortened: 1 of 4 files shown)


pkg/common/pagerduty/pagerduty.go
```diff
@@ -25,3 +27,110 @@ func (pdc Config) FireAlert(details pd.V2Payload) (*pd.V2EventResponse, error) {
...
+				return fmt.Errorf("handler failed: %w", err)
+			}
+		}
+	}
+	return nil
+}
+
+// Alerts uses the provided client to retrieve all Alerts associated with the provided
+// incident, calling the provided handler function on each alert.
+func Alerts(c *pd.Client, incident pd.Incident, options pd.ListIncidentAlertsOptions, handler func(pd.IncidentAlert) error) error {
```
jeefy, line 114: Is `Alerts` used anywhere? I can't find anything in this PR referencing it, maybe I'm blind.

jeefy (changes requested):
One question otherwise super lgtm :rainbow:

pkg/common/pagerduty/pagerduty.go
whereswaldon: I originally used it, then concluded that I didn't need it to perform this particular task. However, it is a working, tested implementation of something that I expect to need if we make this policy more sophisticated, so I thought I'd leave it in there. Happy to remove it for now if you'd prefer.

PR #772, opened 2021-04-01
OSD-6594 [ Added test to verify validation-webhook service is running.]

aliceh:
In this commit, I added tests to verify that the validation-webhook daemon set and service are running.

### DIFF
--- a/pkg/e2e/verify/pod_webhook.go
+++ b/pkg/e2e/verify/pod_webhook.go
@@ -22,11 +22,30 @@ func init() {
 	alert.RegisterGinkgoAlert(podWebhookTestName, "SD-SREP", "Matt Bargenquast", "sd-cicd-alerts", "sd-cicd@redhat.com", 4)
 }
 
+const (
+	serviceStartTimeout = 1 * time.Minute
+
+	daemonStartTimeout = 1 * time.Minute
+)
+
 var _ = ginkgo.Describe(podWebhookTestName, func() {
 
 	h := helper.New()
 
 	ginkgo.Context("pod webhook", func() {
+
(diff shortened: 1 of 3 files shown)


pkg/e2e/verify/pod_webhook.go
```diff
@@ -250,3 +269,35 @@ func impersonateDedicatedAdmin(h *helper.H, user string) *helper.H {
...
+		Expect(err).NotTo(HaveOccurred(), "failed to get daemonset %v\n", daemonSetName)
+		if err != nil {
+			return false, err
+		}
+		return true, nil
+	}
+}
+
+// waitTimeoutForDaemonSetInNamespace waits the given timeout duration for the specified ds.
+func waitTimeoutForDaemonSetInNamespace(h *helper.H, daemonSetName string, namespace string, timeout time.Duration) error {
```
whereswaldon, line 285: Could I convince you to relocate this to a new method on the Helper type? `func (h *Helper) WaitForDaemonSet()` (and do the same for the service polling function below)? They seem useful to many people writing tests, not just this specific use-case.

whereswaldon (changes requested):
I think this implementation totally makes sense, but we could make life easier for future test writers by exposing some of your helper functions to broader use.

pkg/e2e/verify/pod_webhook.go
aliceh: @whereswaldon thanks for the suggestion!  I updated the commit accordingly.

whereswaldon (approved):
Thanks for exposing those helpers!

/lgtm

PR #773, opened 2021-04-05
SDCICD-628: Remove reliance on CS metrics, leave stubs for AMS metrics

jeefy:
This removes code referring to the CS Metrics endpoint. It still leaves the OCM Metrics test as well as the SPI Metrics endpoint in place, for when AMS adds an endpoint to consume metrics from.


whereswaldon (approved):
/lgtm

PR #776, opened 2021-04-07
SDCICD-631: Update cluster deletion to handle hibernated clusters

jeefy:


pkg/common/providers/ocmprovider/cluster.go
```diff
@@ -390,16 +391,49 @@ func (o *OCMProvider) GenerateProperties() (map[string]string, error) {
...
+			cluster, err = o.GetCluster(clusterID)
+			if err != nil {
+				log.Printf("error retrieving cluster for deletion: %v", err)
+				return false, nil
+			}
+			if cluster.State() == spi.ClusterStateReady {
+				return true, nil
+			}
+			return false, nil
+		})
```
whereswaldon, line 436: I think that writing it in this (untested) order reads a little more easily, but I also kinda suspect that I'm missing a nuance here that makes this a bad idea:
```suggestion
	if cluster.State() == spi.ClusterStateHibernating {
		if err = retryer().Do(func() error {
			var err error
			resumeResp, err = o.conn.ClustersMgmt().V1().Clusters().Cluster(clusterID).Resume().Send()

			if err != nil {
				return fmt.Errorf("couldn't resume cluster '%s': %v", clusterID, err)
			}

			if resumeResp != nil && resumeResp.Error() != nil {
				err = errResp(resumeResp.Error())
				log.Printf("%v", err)
				return err
			}

			return nil
		}); err != nil {
			return err
		}
	}
// hopefully the cluster state changes by now
	if cluster.State() == spi.ClusterStateResuming {

		wait.PollImmediate(1*time.Minute, 15*time.Minute, func() (bool, error) {
			cluster, err = o.GetCluster(clusterID)
			if err != nil {
				log.Printf("error retrieving cluster for deletion: %v", err)
				return false, nil
			}
			if cluster.State() == spi.ClusterStateReady {
				return true, nil
			}
			return false, nil
		})
```

whereswaldon (changes requested):
I see an a potential simplification, but it might not actually work. Let's talk before I lgtm. Everything else looks great!

pkg/common/providers/ocmprovider/cluster.go
jeefy: I took a different approach simplifying it, mind taking a look again? :)

whereswaldon (approved):
/lgtm

PR #778, opened 2021-04-09
Extract route monitor data

agopala95:
Extract the plot numbers generated for route monitors during osde2e tests and store them in a separate variable.


pkg/e2e/routemonitors/routemonitors.go
```diff
@@ -237,3 +243,73 @@ func createPlot(title string) *plot.Plot {
 		plot.Label(plot.ErrorLabeler),
 	)
 }
+
+// Extracts the numbers and title from the route monitor plot files
+func (rm *RouteMonitors) ExtractData(baseDir string) error {
+	outputDirectory := filepath.Join(baseDir, "route-monitors")
+	if _, err := os.Stat(outputDirectory); os.IsNotExist(err) {
+		if err := os.Mkdir(outputDirectory, os.FileMode(0755)); err != nil {
```
whereswaldon, line 251: Are we 100% certain that `baseDir` already exists? If not, this can fail due to trying to create a child of a nonexistent parent. You can avoid that with `os.MkdirAll`

pkg/e2e/routemonitors/routemonitors.go
```diff
@@ -237,3 +243,73 @@ func createPlot(title string) *plot.Plot {
...
+
+		// Regex to match numbers for the data variable inside the plot file
+		start_regex := regexp.MustCompile(`var data`)
+		datanum_regex := regexp.MustCompile(`[0-9]*\.?[0-9]+,[0-9]*\.?[0-9]+`)
+
+		// Boolean to only scan for numbers once the line `var data` has passed
+		readdata := false
+
+		// This list stores the numbers
+		datalist := make([][]float64, 0)
```
whereswaldon, line 274: Since is this a list of `time, value` pairs, why not:

```go
type Datum struct {
    Time, Value float64
}

datalist := make([]Datum, 0)
```

I think it'd be easier for others trying to understand the structure of the data if each sample had named fields (though I'm not attached to these names).

pkg/e2e/routemonitors/routemonitors.go
```diff
@@ -237,3 +243,73 @@ func createPlot(title string) *plot.Plot {
...
+func (rm *RouteMonitors) ExtractData(baseDir string) error {
+	outputDirectory := filepath.Join(baseDir, "route-monitors")
+	if _, err := os.Stat(outputDirectory); os.IsNotExist(err) {
+		if err := os.Mkdir(outputDirectory, os.FileMode(0755)); err != nil {
+			return fmt.Errorf("error while creating route monitor report directory %s: %v", outputDirectory, err)
+		}
+	}
+
+	for title, pl := range rm.Plots {
+		_ = pl
```
whereswaldon, line 257:
```suggestion
	for title := range rm.Plots {
```

pkg/e2e/routemonitors/routemonitors.go
```diff
@@ -134,10 +139,11 @@ func Create() (*RouteMonitors, error) {
...
 	return &RouteMonitors{
-		Monitors:  make(map[string]<-chan *vegeta.Result, 0),
-		Metrics:   make(map[string]*vegeta.Metrics, 0),
-		Plots:     make(map[string]*plot.Plot, 0),
-		targeters: targeters,
+		Monitors:   make(map[string]<-chan *vegeta.Result, 0),
+		Metrics:    make(map[string]*vegeta.Metrics, 0),
+		Plots:      make(map[string]*plot.Plot, 0),
+		targeters:  targeters,
+		ReportData: make(map[string][][]float64, 0),
```
whereswaldon, line 146:
```suggestion
		Monitors:   make(map[string]<-chan *vegeta.Result),
		Metrics:    make(map[string]*vegeta.Metrics),
		Plots:      make(map[string]*plot.Plot),
		targeters:  targeters,
		ReportData: make(map[string][][]float64),
```

Maps don't need to be initialized with a size, unlike slices. They're backed by some kind of fancy b-tree rather than an array.

whereswaldon (commented):
Lookin' good! I have a few suggestions/questions, but nothing major.

pkg/e2e/routemonitors/routemonitors.go
agopala95: Thanks for the input. I've included `os.MkdirAll`.

pkg/e2e/routemonitors/routemonitors.go
agopala95: Yup. I've made the changes. Hope the new name's alright though.

pkg/e2e/routemonitors/routemonitors.go
agopala95: Got it. I've removed the size parameters. Thanks!

whereswaldon (approved):
/lgtm

PR #779, opened 2021-04-09
Sdcicd 577

whereswaldon:
I believe that this *finally* resolves SD-CICD 577. Essentially, these changes alter our healthcheck in the following ways:

- waiting for OCM to finish provisioning a cluster has been broken out into its own reusable function
- the different reasons that we might wait on cluster health now have different functions. This allows us to specialize the implementation to use the healthcheck most appropriate for the circumstance
- `WaitForClusterReadyPostInstall` properly waits for OCM and then the healthcheck cron job.

### DIFF
--- a/cmd/osde2ectl/create/cmd.go
+++ b/cmd/osde2ectl/create/cmd.go
@@ -298,7 +298,7 @@ func setupCluster(wg *sync.WaitGroup, successfulClustersCounter *int32) {
 			}
 		}()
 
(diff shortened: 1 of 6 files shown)


whereswaldon:
This contained a rebase onto a branch that reverted some history, so the initial PR was broken. However, I think I've got all of the right commits in place again ;D

jeefy:
Herculean effort
/lgtm

PR #780, opened 2021-04-12
fix: avoid api server restart race that failed healthcheck

whereswaldon:


jeefy:
praise be
/lgtm

PR #782, opened 2021-04-13
SDCICD-637: Simplify build log output

jeefy:
Instead of re-implementing multiwriter, we should just use multiwriter. :)


whereswaldon (commented):
/lgtm

PR #783, opened 2021-04-13
feat: implement concurrentviper via code generation

whereswaldon:
This PR creates a thread-safe version of viper for us to use via code generation from the official viper package. We're occasionally crashing due to races within viper, so this seems like the best way forward (upstream viper doesn't seem interested in supporting concurrent use).


whereswaldon:
/hold

I want to look at this with a fresh brain tomorrow to ensure I didn't do a dumb.

jeefy:
So freakin' cool
/lgtm

PR #784, opened 2021-04-14
Fix: Swap ECR registries for Regional Quays in ICSP testing

jeefy:
Also remove redundant test cases. We only need to test Regional Quay as the primary mirror and then Main Quay as the primary mirror.


whereswaldon (approved):
/lgtm

PR #787, opened 2021-04-19
Check roles with prefix exist

arjunrn:
The `ClusterServiceVersion` for the `configure-alertmanager-operator` now contains the roles for the operator as part of the `permissions` field. This means that the name of the `Role` and the `RoleBinding` are not fixed but are prefixed with the CSV name. This PR changes the test so that the roles and role bindings are verified in the same namespace based only on the prefix.


whereswaldon:
@arjunrn Looks like this doesn't currently compile. Logically, the changes look good though.

whereswaldon (approved):
/lgtm

PR #792, opened 2021-04-23
Added prometheus pod check for collect prometheus pod

agopala95:
Added a check to ensure prometheus pods are up and running before executing extraction of prometheus data within the "collect-prometheus" pod

### DIFF
--- a/pkg/e2e/state/prometheus.go
+++ b/pkg/e2e/state/prometheus.go
@@ -1,11 +1,16 @@
 package state
 
 import (
+	"errors"
+	"time"
+
 	"github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
 	"github.com/openshift/osde2e/pkg/common/helper"
 	"github.com/openshift/osde2e/pkg/common/runner"
+	"github.com/openshift/osde2e/pkg/e2e/verify"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 const (
@@ -25,6 +30,27 @@ var _ = ginkgo.Describe(clusterStateInformingName, func() {
 		// 1 as tar 1.26 will exit 1 if files change while the tar is running, as is
 		// common for a running prometheus instance
 		cmd := promCollectCmd + " > " + runner.DefaultRunner.OutputDir + "/prometheus.tar.gz; err=$? ; if (( $err != 0 )) ; then exit $err ; fi"
+
+		poderr := wait.PollImmediate(2*time.Second, 5*time.Minute, func() (bool, error) {
+			podCount := 0
+			var podErr error
+			list, _ := verify.FilterPods("openshift-monitoring", "app=prometheus", h)
+			names, podNum := verify.GetPodNames("prometheus-operator", list, h)
+			if podNum > 0 {
+				for _, value := range names {
+					if (value == "prometheus-k8s-0") || (value == "prometheus-k8s-1") {
+						podCount += 1
+					}
+				}
+				if podCount == 2 {
+					podErr = nil
+					return true, podErr
+				}
+			}
+			podErr = errors.New("Prometheus pods are not in running state")
+			return false, podErr
+		})
+		Expect(poderr).NotTo(HaveOccurred())
 
 		h.SetServiceAccount("system:serviceaccount:%s:cluster-admin")
 		r := h.Runner(cmd)
--- a/pkg/e2e/state/prometheus.go
+++ b/pkg/e2e/state/prometheus.go
@@ -2,6 +2,7 @@ package state
 
 import (
 	"errors"
+	"strings"
 	"time"
 
 	"github.com/onsi/ginkgo"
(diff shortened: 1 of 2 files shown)


pkg/e2e/state/prometheus.go
```diff
@@ -26,6 +31,27 @@ var _ = ginkgo.Describe(clusterStateInformingName, func() {
...
 		cmd := promCollectCmd + " > " + runner.DefaultRunner.OutputDir + "/prometheus.tar.gz; err=$? ; if (( $err != 0 )) ; then exit $err ; fi"
 
+		poderr := wait.PollImmediate(2*time.Second, 5*time.Minute, func() (bool, error) {
+			podCount := 0
+			var podErr error
+			list, _ := verify.FilterPods("openshift-monitoring", "app=prometheus", h)
+			names, podNum := verify.GetPodNames("prometheus-operator", list, h)
+			if podNum > 0 {
+				for _, value := range names {
+					if (value == "prometheus-k8s-0") || (value == "prometheus-k8s-1") {
```
whereswaldon, line 41:
```suggestion
					if strings.HasPrefix("prometheus-k8s-", value) {
```
Technically doesn't doesn't check the number at all, but I'm not sure that matters. You can take this or leave it.

pkg/e2e/state/prometheus.go
```diff
@@ -26,6 +31,27 @@ var _ = ginkgo.Describe(clusterStateInformingName, func() {
...
+						podCount += 1
+					}
+				}
+				if podCount == 2 {
```
whereswaldon, line 45:
```suggestion
				if podCount >= 2 {
```
What if the default number of pods increases in the future?

pkg/e2e/state/prometheus.go
```diff
@@ -26,6 +31,27 @@ var _ = ginkgo.Describe(clusterStateInformingName, func() {
...
+					podErr = nil
+					return true, podErr
+				}
+			}
+			podErr = errors.New("Prometheus pods are not in running state")
+			return false, podErr
```
whereswaldon, line 51:
```suggestion
			return false, errors.New("Prometheus pods are not in running state")
```

pkg/e2e/state/prometheus.go
```diff
@@ -26,6 +31,27 @@ var _ = ginkgo.Describe(clusterStateInformingName, func() {
...
+			names, podNum := verify.GetPodNames("prometheus-operator", list, h)
+			if podNum > 0 {
+				for _, value := range names {
+					if (value == "prometheus-k8s-0") || (value == "prometheus-k8s-1") {
+						podCount += 1
+					}
+				}
+				if podCount == 2 {
+					podErr = nil
+					return true, podErr
```
whereswaldon, line 47:
```suggestion
					return true, nil
```

pkg/e2e/state/prometheus.go
```diff
@@ -26,6 +31,27 @@ var _ = ginkgo.Describe(clusterStateInformingName, func() {
 		// common for a running prometheus instance
 		cmd := promCollectCmd + " > " + runner.DefaultRunner.OutputDir + "/prometheus.tar.gz; err=$? ; if (( $err != 0 )) ; then exit $err ; fi"
 
+		poderr := wait.PollImmediate(2*time.Second, 5*time.Minute, func() (bool, error) {
+			podCount := 0
+			var podErr error
+			list, _ := verify.FilterPods("openshift-monitoring", "app=prometheus", h)
```
whereswaldon, line 37: Why discard the error here?

pkg/e2e/state/prometheus.go
```diff
@@ -26,6 +31,27 @@ var _ = ginkgo.Describe(clusterStateInformingName, func() {
 		// common for a running prometheus instance
 		cmd := promCollectCmd + " > " + runner.DefaultRunner.OutputDir + "/prometheus.tar.gz; err=$? ; if (( $err != 0 )) ; then exit $err ; fi"
 
+		poderr := wait.PollImmediate(2*time.Second, 5*time.Minute, func() (bool, error) {
+			podCount := 0
+			var podErr error
```
whereswaldon, line 36: I don't think you need this variable, though perhaps I'm missing something.

whereswaldon (changes requested):
I think there are some opportunities to simplify, but this looks solid.

pkg/e2e/state/prometheus.go
agopala95: Yeah this makes sense. I've updated the check.

pkg/e2e/state/prometheus.go
agopala95: That could be a possibility. I've implemented the change.

pkg/e2e/state/prometheus.go
agopala95: Yup I guess we wouldn't be needing this variable.

pkg/e2e/state/prometheus.go
agopala95: Changes done!

pkg/e2e/state/prometheus.go
agopala95: The FilterPods function prints an error so I thought that would be enough but I guess including the error here is far more robust.

whereswaldon (commented):
/lgtm

PR #794, opened 2021-04-28
OSD-7026 make dedicated-admin tests more robust

mrbarge:
In [OSD-7026](https://issues.redhat.com/browse/OSD-7026) the CI-Watcher identified a case where the `dedicated-admin permissions` test was irregularly failing. [0]

This test failure is the result of the `subscription` resource (retrieved early in the test) having being updated to a new `resourceVersion` before the test later attempts to perform an `Update` on it. Because the resource version on-cluster is newer than the version loaded into memory, the `Update()` request is rejected, failing the test.

This situation is easily handled through the use of the `RetryOnConflict` k8s helper. 

Previous to this PR change, I was able to replicate the test failure every few test runs. Since the proposed fix I have not been able to recreate the failure.

[0] https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/osde2e-stage-aws-informing-default/1384144901322575872

### DIFF
--- a/pkg/e2e/osd/dedicatedadmin.go
+++ b/pkg/e2e/osd/dedicatedadmin.go
@@ -9,14 +9,15 @@ import (
 	projectv1 "github.com/openshift/api/project/v1"
 	userv1 "github.com/openshift/api/user/v1"
 	"github.com/openshift/osde2e/pkg/common/alert"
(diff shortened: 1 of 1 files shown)


whereswaldon (commented):
/lgtm

PR #795, opened 2021-04-28
Fix for prometheus pod check

agopala95:
Added a minor fix for the wait poll check (to see if prometheus pods are up) that happens before the collect-prometheus pod tries to extract data from prometheus.


whereswaldon (commented):
/lgtm

PR #799, opened 2021-04-30
Poll the roles and rolebinding instead of a one-time check

arjunrn:
Even the presence of the CSV is verified before this test the CSV might be in the `Pending` state. It might take some time for the resources to be provisioned by the `catalog-operator` and `olm-operator`.


whereswaldon (approved):
/approve

PR #800, opened 2021-04-30
Poll the osd-metrics-exporter endpoint

arjunrn:
Both the `Service` object the HTTP endpoint are available only after the operator pod is running. So even though the CSV is present the endpoint and the service may be unavailable.


whereswaldon (approved):
/approve

PR #802, opened 2021-04-30
SDCICD-624: Pass through non-osde2e secrets into test cluster for addon testing

jeefy:
This will insert any non-OSDe2e-specific secrets specified by Addon Authors into the test cluster under the `osde2e-ci-secrets` namespace as a secret named `ci-secrets`

### DIFF
--- a/pkg/common/config/config.go
+++ b/pkg/common/config/config.go
@@ -65,6 +65,10 @@ const (
 
 	// Default network provider for OSD
 	DefaultNetworkProvider = "OpenShiftSDN"
+
+	// NonOSDe2eSecrets is an internal-only Viper Key.
+	// End users should not be using this key, there may be unforeseen consequences.
+	NonOSDe2eSecrets = "nonOSDe2eSecrets"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/common/config/config.go
```diff
@@ -65,6 +65,8 @@ const (
 
 	// Default network provider for OSD
 	DefaultNetworkProvider = "OpenShiftSDN"
+
+	NonOSDe2eSecrets = "nonOSDe2eSecrets"
```
whereswaldon, line 69: Do you mind dropping a comment to explain the purpose of this constant?

whereswaldon: Especially since this viper key isn't meant to be used by our users, I think it's important to document that.

whereswaldon (changes requested):
Just a quick doc request, implementation looks good.

pkg/common/config/config.go
jeefy: ACK -- good point, done. :)

whereswaldon (approved):
/lgtm

PR #804, opened 2021-04-30
Removed HTTP call which may not be running on-cluster

arjunrn:
It appears the test don't actually run on the test cluster but instead from a separate location. While writing this test I initially assumed that the test binary would run in the cluster and added an HTTP call to verify that an endpoint is up and running. This PR removes that test.


whereswaldon (commented):
/lgtm

PR #809, opened 2021-05-03
[OSD-6633] Create osde2e test to confirm that no platform workloads on worker nodes

aliceh:
In this commit, I added a test to verify that static operator pods run only on infra nodes.


pkg/e2e/verify/check_workloads.go
```diff
@@ -0,0 +1,133 @@
...
+		"configure-alertmanager-operator": "openshift-monitoring",
+		"custom-domains-operator":         "openshift-custom-domains-operator",
+		"managed-upgrade-operator":        "openshift-managed-upgrade-operator",
+		"managed-velero-operator":         "openshift-velero",
+		"must-gather-operator":            "openshift-must-gather-operator",
+		"osd-metrics-exporter":            "openshift-osd-metrics",
+		"rbac-permissions-operator":       "openshift-rbac-permissions"}
+
+	listOpts := metav1.ListOptions{}
+	list, err := podClient.Pods(metav1.NamespaceAll).List(context.TODO(), listOpts)
```
whereswaldon, line 55: Given that there may be many pods, do we need to handle pagination here?

pkg/e2e/verify/check_workloads.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+
+	}
+
+	if len(violators) > 0 {
+		return violators, fmt.Errorf("Found infrastructure pods that do not run on infra nodes %v.", err)
+	}
+
+	//log.Printf("Pods %v", filteredPods)
+
```
whereswaldon, line 88:
```suggestion
```

pkg/e2e/verify/check_workloads.go
```diff
@@ -0,0 +1,133 @@
...
+		})
+	})
+})
+
+func checkPods(podClient v1.CoreV1Interface, infra []string) (map[string]string, error) {
+
+	violators := make(map[string]string)
+
+	operators := map[string]string{"cloud-ingress-operator": "openshift-cloud-ingress-operator",
```
whereswaldon, line 45:
```suggestion
	type (
		Namespace = string
		OperatorName = string
		Node = string
	)
	violators := make(map[OperatorName]NodeName)

	operators := map[OperatorName]Namespace{"cloud-ingress-operator": "openshift-cloud-ingress-operator",
```
This is totally subjective, and definitely not required for this PR to be accepted. I was curious what you thought of using some type aliases to make it clearer what mapped to what in these two `map[string]string`?

pkg/e2e/verify/check_workloads.go
```diff
@@ -0,0 +1,133 @@
...
+	if len(violators) > 0 {
+		return violators, fmt.Errorf("Found infrastructure pods that do not run on infra nodes %v.", err)
+	}
+
+	//log.Printf("Pods %v", filteredPods)
+
+	return violators, nil
+}
+
+func checkWorkloadsOnWorkerNodes(nodeClient v1.CoreV1Interface) ([]string, []string, error) {
```
whereswaldon, line 92:
```suggestion
func getWorkerNodesByType(nodeClient v1.CoreV1Interface) ([]string, []string, error) {
```
Can we call this something else? It doesn't "check" the workloads so much as filter the list of nodes to those with a specific role.

whereswaldon (changes requested):
Hey Alice, this is looking good! I have some suggestions and questions that I'd like to talk through, but nothing major.

pkg/e2e/verify/check_workloads.go
```diff
@@ -0,0 +1,133 @@
...
+	if len(list.Items) == 0 {
+		return nil, fmt.Errorf("pod list is empty. this should NOT happen")
+	}
+
+	for _, pod := range list.Items {
+		//For each pod.ObjectMeta.Name, pod.Spec.NodeName we will check if they are a match for the operaor namespace and pod name prefix.
+		for op, namespace := range operators {
+			if pod.ObjectMeta.Namespace == namespace {
+				if strings.HasPrefix(pod.ObjectMeta.Name, op) {
+
```
dustman9000, line 69: nit: remove extra line

pkg/e2e/verify/check_workloads.go
```diff
@@ -0,0 +1,133 @@
...
+					if !stringInSlice(pod.Spec.NodeName, infra) {
+
```
dustman9000, line 71: nit: remove line

pkg/e2e/verify/check_workloads.go
```diff
@@ -0,0 +1,133 @@
...
+	if len(list.Items) == 0 {
+		return nil, fmt.Errorf("pod list is empty. this should NOT happen")
+	}
+
+	for _, pod := range list.Items {
+		//For each pod.ObjectMeta.Name, pod.Spec.NodeName we will check if they are a match for the operaor namespace and pod name prefix.
+		for op, namespace := range operators {
+			if pod.ObjectMeta.Namespace == namespace {
+				if strings.HasPrefix(pod.ObjectMeta.Name, op) {
```
dustman9000, line 68: We also need to make sure the pod name does not contain the "registry" string until https://issues.redhat.com/browse/RFE-1843 is fixed.

pkg/e2e/verify/check_workloads.go
```diff
@@ -0,0 +1,133 @@
...
+})
+
+func checkPods(podClient v1.CoreV1Interface, infra []string) (map[string]string, error) {
+
+	violators := make(map[string]string)
+
+	operators := map[string]string{"cloud-ingress-operator": "openshift-cloud-ingress-operator",
+		"configure-alertmanager-operator": "openshift-monitoring",
+		"custom-domains-operator":         "openshift-custom-domains-operator",
+		"managed-upgrade-operator":        "openshift-managed-upgrade-operator",
```
mrbarge, line 48: I'll just forewarn that `managed-upgrade-operator` is going to move to master nodes (rather than infra nodes) in OSD-7104 this sprint, so it will soon end up failing this test.

aliceh: @mrbarge thanks for the heads up!

pkg/e2e/verify/check_workloads.go
aliceh: I agree that it makes the code easier to read. I will update the code accordingly, the violators map is PodName: Node.

pkg/e2e/verify/check_workloads.go
aliceh: I'll rename it, thanks for the suggestion.

pkg/e2e/verify/check_workloads.go
aliceh: Done.

pkg/e2e/verify/check_workloads.go
aliceh: Right now the code runs through this bit very fast. I don't think that the number of pods will reach such huge numbers that it will make a difference.

pkg/e2e/verify/check_workloads.go
```diff
@@ -0,0 +1,132 @@
...
+			if key == "node-role.kubernetes.io/infra" {
+				infra = append(infra, node.Name)
+				log.Printf("Node, infra: %v", node.Name)
+			}
+		}
+
+	}
+
+	return worker, infra, nil
+}
```
whereswaldon, line 123: A final suggestion: this is a way to communicate which returned slice contains which kind of nodes without significant code change. A comment would also work. Right now, the only way to figure this out is to read the source code, and I'd rather the info be surfaced another way.

```suggestion
func listNodesByType(nodeClient v1.CoreV1Interface) (worker, infra []string, err error) {

	log.Printf("Getting node list")
	var (
		listOpts metav1.ListOptions
		list *v1.NodeList
	)
	list, err = nodeClient.Nodes().List(context.TODO(), listOpts)
	if err != nil {
		return nil, nil, fmt.Errorf("error getting node list: %v", err)
	}

	if len(list.Items) == 0 {
		return nil, nil, fmt.Errorf("no nodes found")
	}

	for _, node := range list.Items {
		labels := node.Labels

		for key, _ := range labels {
			if key == "node-role.kubernetes.io/worker" {
				worker = append(worker, node.Name)
			}
			if key == "node-role.kubernetes.io/infra" {
				infra = append(infra, node.Name)
				log.Printf("Node, infra: %v", node.Name)
			}
		}

	}

	return worker, infra, nil
}
```

whereswaldon (commented):
@aliceh Lookin' good. There's a function that could use some form of documentation about its return types, but that's not a huge deal. I'd be happy to approve this once someone from your team gives it the go-ahead.

pkg/e2e/verify/check_workloads.go
aliceh: I have trouble with declarations of this type  list *v1.NodeList, I get "NodeList not declared by package v1compiler" error

aliceh: The list, err := nodeClient.Nodes().List(context.TODO(), listOpts) is similar to other calls we use to list kubernetes objects, I will put some extra comments in the code to clarify.

PR #818, opened 2021-05-07
SDCICD-641: Preserve metadata.json during failed cluster installation

jeefy:
metadata.json is created during a openshift-install operation, but when the install-step fails, metadata.json isn't preserved. When this happens, the destroy-step doesn't have any knowledge of the cluster or resources created and thus can't clean up resources from a failed installation.

This will trap any error during cluster creation and, if it occurs, still copy the metadata.json and install-config into the shared directory for the destroy-step to work its magic.

### DIFF
--- a/ci/create-ocp-cluster.sh
+++ b/ci/create-ocp-cluster.sh
@@ -37,8 +37,16 @@ chmod +x oc
 
 chmod +x openshift-install
 
-./openshift-install create cluster --dir=./installer/ --log-level info
+{
+    ./openshift-install create cluster --dir=./installer/ --log-level info &&
 
-cp "${INSTALLER_CONFIG}" "${SHARED_DIR}"
-cp ./installer/metadata.json "${SHARED_DIR}"
-cp ./installer/auth/kubeconfig "${SHARED_DIR}"
+    cp "${INSTALLER_CONFIG}" "${SHARED_DIR}"
+    cp ./installer/metadata.json "${SHARED_DIR}"
+    cp ./installer/auth/kubeconfig "${SHARED_DIR}"
+    echo "Copied installer files."
+} || {
+    cp "${INSTALLER_CONFIG}" "${SHARED_DIR}"
+    cp ./installer/metadata.json "${SHARED_DIR}"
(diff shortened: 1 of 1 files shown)


ci/create-ocp-cluster.sh
```diff
@@ -37,8 +37,16 @@ chmod +x oc
...
-cp ./installer/metadata.json "${SHARED_DIR}"
-cp ./installer/auth/kubeconfig "${SHARED_DIR}"
\ No newline at end of file
+{
+    ./openshift-install create cluster --dir=./installer/ --log-level info &&
+
+    cp "${INSTALLER_CONFIG}" "${SHARED_DIR}"
+    cp ./installer/metadata.json "${SHARED_DIR}"
+    cp ./installer/auth/kubeconfig "${SHARED_DIR}"
+    echo "Copied installer files."
```
whereswaldon, line 46:
```suggestion
    ./openshift-install create cluster --dir=./installer/ --log-level info &&
    cp "${INSTALLER_CONFIG}" ./installer/metadata.json ./installer/auth/kubeconfig "${SHARED_DIR}" &&
    echo "Copied installer files."
```

ci/create-ocp-cluster.sh
```diff
@@ -37,8 +37,16 @@ chmod +x oc
...
+} || {
+    cp "${INSTALLER_CONFIG}" "${SHARED_DIR}"
+    cp ./installer/metadata.json "${SHARED_DIR}"
+    echo "Error creating cluster.";
+    exit 1;
```
whereswaldon, line 51:
```suggestion
    cp "${INSTALLER_CONFIG}" ./installer/metadata.json "${SHARED_DIR}" &&
    echo "Error creating cluster.";
    exit 1;
```

whereswaldon (changes requested):
I didn't realize that `&&` continued across multiple newlines, so thanks for teaching me that! I do think this can be simplified a little bit without losing much in terms of readability. Let me know what you think!

whereswaldon (approved):
/lgtm

PR #819, opened 2021-05-07
Updates to clarify what the variables are in OSD-6633

aliceh:
In this commit I added some extra comments to make the code easier to understand and changed the definition of a map, to be descriptive and easy to understand.


whereswaldon (approved):
/approve

PR #826, opened 2021-05-14
SDCICD-183: Add weight to choosing a random machine type

jeefy:
Since the heart of [SDCICD-183](https://issues.redhat.com/browse/SDCICD-183) is to convert our pipelines to choose a random machine type, but most machine types are a bit more expensive, this adds a weight so every 1 out of 3 times, when the machine type is set to `random`, it will actually pick a random machine type from OCM. Otherwise, it blanks the variable out internally and creates the cluster with the default.

### DIFF
--- a/pkg/common/providers/ocmprovider/cluster.go
+++ b/pkg/common/providers/ocmprovider/cluster.go
@@ -346,10 +346,11 @@ func ChooseRandomRegion(regions ...CloudRegion) (CloudRegion, bool) {
 // by the OCM provider. (Returns a random machine type if the config suggests it to be random.)
 func (o *OCMProvider) DetermineMachineType(cloudProvider string) (string, error) {
 	computeMachineType := viper.GetString(ComputeMachineType)
+	returnedType := ""
 
-	// If a region is set to "random", it will poll OCM for all the regions available
-	// It then will pull a random entry from the list of regions and set the ID to that
-	if computeMachineType == "random" {
+	// If a machineType is set to "random", it will poll OCM for all the machines available
+	// It then will pull a random entry from the list of machines and set the machineTypes to that
+	if computeMachineType == "random" && rand.Intn(3) >= 2 {
 		searchstring := fmt.Sprintf("cloud_provider.id like '%s'", cloudProvider)
 		machinetypeClient := o.conn.ClustersMgmt().V1().MachineTypes().List().Search(searchstring)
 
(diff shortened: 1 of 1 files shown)


pkg/common/providers/ocmprovider/cluster.go
```diff
@@ -346,10 +346,11 @@ func ChooseRandomRegion(regions ...CloudRegion) (CloudRegion, bool) {
 // by the OCM provider. (Returns a random machine type if the config suggests it to be random.)
 func (o *OCMProvider) DetermineMachineType(cloudProvider string) (string, error) {
 	computeMachineType := viper.GetString(ComputeMachineType)
+	returnedType := ""
 
 	// If a region is set to "random", it will poll OCM for all the regions available
 	// It then will pull a random entry from the list of regions and set the ID to that
-	if computeMachineType == "random" {
+	if computeMachineType == "random" && rand.Intn(3) > 2 {
```
whereswaldon, line 353: This can never be true. `Intn` returns a number less than the provided integer, so the highest possible value is 2.
```suggestion
	if computeMachineType == "random" && rand.Intn(3) >= 2 {
```

pkg/common/providers/ocmprovider/cluster.go
```diff
@@ -360,16 +361,17 @@ func (o *OCMProvider) DetermineMachineType(cloudProvider string) (string, error)
 
 		for range machinetypes.Items().Slice() {
 			machinetypeObj := machinetypes.Items().Slice()[rand.Intn(machinetypes.Total())]
-			computeMachineType = machinetypeObj.ID()
+			returnedType = machinetypeObj.ID()
 			break
 		}
+	}
```
whereswaldon, line 367: So if we don't take the `if` statement above, when we get here we set the machine type to the empty string. Is that the intended behavior? Does empty string have semantics like "use the default" that I don't know about?

pkg/common/providers/ocmprovider/cluster.go
```diff
@@ -360,16 +361,17 @@ func (o *OCMProvider) DetermineMachineType(cloudProvider string) (string, error)
...
-		log.Printf("Random machine type requested, selected %s machine type.", computeMachineType)
+	log.Printf("Random machine type requested, selected `%s` machine type.", returnedType)
 
-		// Update the Config with the selected random region
-		viper.Set(ComputeMachineType, computeMachineType)
-	}
-	return computeMachineType, nil
+	// Update the Config with the selected random region
+	viper.Set(ComputeMachineType, returnedType)
+
```
whereswaldon, line 373: One or both of these statements seem like they may belong inside of the `if` block.

whereswaldon (changes requested):
I'm not sure that I understand how this is meant to work. I left a few questions inline.

jeefy:
Merging this in as all comments were addressed.

PR #827, opened 2021-05-17
fix(es): Prevent cluster claim conflicts, add flag preventing recycled clusters

jeefy:
Once we retrieve a cluster, we should immediately tag it with the current Job ID before doing any other operation. 

We should then validate via a second GET whether the Job ID is the same as the current job. If not, recuuuuuurse. 

Additionally, there is now a flag to allow Recycled Clusters for a job (default: True) which means some jobs we can set to always require a fresh cluster.


whereswaldon (commented):
/lgtm

PR #829, opened 2021-05-18
Sdcicd 662: Define a postgres database schema for test data

whereswaldon:
/assign @jeefy 

This creates initial tests for schema migrations as well as creating an initial schema. The schema is broken into two migrations to demonstrate the functionality of the tests to handle multiple migrations.

I guess this will fail in prow because docker isn't available. I'm happy to retool so that we can connect to a different instance of postgres, but this is already large as it is. I'd like some eyes on it before I try to add that feature.


pkg/db/db_test.go
```diff
@@ -0,0 +1,194 @@
...
+			if err := m.Up(); err != nil {
+				t.Fatalf("Failed running all up migrations: %v", err)
+			}
+			maxVersion, _, err := m.Version()
+			if err != nil {
+				t.Fatalf("Did not expect error fetching final migration version: %v", err)
+			}
+			if err := m.Down(); err != nil {
+				t.Fatalf("Failed running all down migrations: %v", err)
+			}
```
jeefy, line 163: Is this testing a full-up then full-down.

That seems somewhat duplicative to the tests below, is there value in doing both?

pkg/db/migrations/001_job.up.sql
```diff
@@ -0,0 +1,27 @@
+CREATE TYPE job_result AS ENUM ('passed', 'failed', 'aborted');
```
jeefy, line 1: We should also have 'running' as an option, and have new rows default to that. I could see value in that for reporting or alerting.

pkg/db/migrations/002_testcase.up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TYPE test_result as ENUM ('passed', 'failed', 'skipped', 'pending');
+
+CREATE TABLE IF NOT EXISTS testcases (
```
jeefy, line 3: Open question: Is there value in storing which JUnit xml file the row was parsed from? We're quickly going to be parsing multiple files to generate db data.

pkg/db/migrations/001_job.up.sql
```diff
@@ -0,0 +1,27 @@
...
+
+CREATE TABLE IF NOT EXISTS jobs (
```
jeefy, line 3: More a comment than a requested change, I'd typically use varchars for fields we know have finite limits and ones we'll want to potentially build indicies off of. 

Some of these fields are OCM fields and we can find out and duplicate their limits.

jeefy (commented):
Overall lgtm. Couple questions/comments to satisfy my curiosity 😁

pkg/db/db_test.go
whereswaldon: This was the only way I could find to determine how many migrations there are. I do this to get the value of `maxVersion`. Seems to me that our migrations should always be able to handing full up and then down like this, and the DB should be empty at this point so it should be pretty quick.

I considered writing the loop to just go until an error indicating that there weren't any more migrations. If you'd prefer that, I can try to reframe it that way.

pkg/db/migrations/001_job.up.sql
whereswaldon: In my head, I was thinking we'd only be creating rows here at the end of a job, so we'd already know the pass/fail state. I guess we could create the record at the beginning, but I'm curious whether we'd gain much other than complexity. Thoughts?

pkg/db/migrations/002_testcase.up.sql
whereswaldon: I don't see an issue with storing it. Do you mean the URL, or just the file name?

jeefy:
Naw, that's fine. :)

pkg/db/migrations/001_job.up.sql
whereswaldon: Hmm. We certainly can do this. Do you have any links I can look at for the index-building implications of each option? I'm not familiar with all of that, though I know that the `text` type is stored differently than `varchar`.

pkg/db/db_test.go
jeefy: Nah, I'm good with that!

pkg/db/migrations/001_job.up.sql
jeefy: I was mentally thinking a 1:1 mapping between Prow job state and DB job state. For simplicity's sake now, I'm cool with keeping it to those 3 / end-of-job. 

Later on if we want that type of data, the change won't be too difficult.

pkg/db/migrations/002_testcase.up.sql
jeefy: Just the filename -- we can compose a URL with the rest of the data stored within the job.

pkg/db/migrations/001_job.up.sql
jeefy: Did a quick look up to find some docs for you and found that in the background there's no functional difference (anymore?) between Text and varchar besides varchar allowing length-checking. 

I say ship it as is!

jeefy:
besides go.sum conflicts, lgtm

pkg/db/migrations/002_testcase.up.sql
jeefy: We can bump this to be an incremental change later.

PR #832, opened 2021-05-20
SDCICD-669: Fix ROSA version selection with other channels

jeefy:
Retrieve and merge stable channel versions with other specified channel version list


whereswaldon (commented):
/lgtm

PR #840, opened 2021-06-02
add post verification steps for managed upgrade

bmeng:


mrbarge:
This seems good to me, I just have one suggestion about where the code eventually sits.

After the upgrade completes, E2E runs a check called the `WaitForClusterReadyPostUpgrade` check ([link](https://github.com/openshift/osde2e/blob/d5f9e3860d3eb603bbad2ffa8b591607c0d23558/pkg/common/upgrade/upgrade.go#L137)) which verifies the cluster is in a healthy state

This in turn calls a PollClusterHealth function ([link](https://github.com/openshift/osde2e/blob/d5f9e3860d3eb603bbad2ffa8b591607c0d23558/pkg/common/cluster/clusterutil.go#L380)) which performs various checks on the cluster state to ensure it is healthy.

This check feels like it belongs as part of that broader overall 'is the cluster healthy' check rather than being separate. Then E2E doubly benefits by having this check performed post-install as well as post-upgrade.

Happy to sign off on this either way, but want to throw it out there as an option.

whereswaldon:
@mrbarge I agree. I'd rather this functionality go into `PollClusterHealth` if this is a generally-useful way to assessing the health of a cluster.

bmeng:
Close this one as the related checking move to the generic cluster health check in #842

PR #841, opened 2021-06-03
fix the upgrade config name as it is renamed in MUO

bmeng:


mrbarge:
Without this fix, all E2E upgrades will be broken at the moment, so I hope @jeefy won't mind if I approve this.

/lgtm
/approve

PR #842, opened 2021-06-07
add ds/rs check for cluster health check

bmeng:
add the ds rs replica count check as part of the cluster health check

the managed upgrade will also be able to consume this


bmeng:
cc @mrbarge

whereswaldon:
@bmeng Does this supercede #840 ?

bmeng:
@whereswaldon ah, yes, this is going to address the last comment in #840 
I will close that one once this merged.

bmeng:
tested this on a local staging cluster. it works well

PR #843, opened 2021-06-07
SDCICD-672: Fix race condition grabbing expiring clusters

jeefy:
There are instances where we have grabbed a cluster that will expire mid-test. This should prevent that by only ever grabbing a cluster that expires 4+ hours after the current time.


jeefy:
Fixed a condition I thought about after reviewing. 
/hold cancel

whereswaldon (commented):
/lgtm

PR #846, opened 2021-06-08
SDCICD-673: Add time between the apply and checking for pod state

jeefy:
There may be instances where we check for the state of a pod's containers before the pod has been created. 

Infamous words ring true: "If there's a race condition, just add another sleep to it."


whereswaldon (commented):
/lgtm

PR #852, opened 2021-06-09
OSD-6471 - Add encrypted_storage.go to test gcp pvc encryption

tnierman:
Created per https://issues.redhat.com/browse/OSD-6471.

As a quick overview, this test:
- Creates a serviceaccount in GCP using the cloud-credential-operator
- Enables KMS & creates an encryption key
- Creates encrypted storageclass
- Creates pvc+pod & checks basic functionality of the disk

I opted not to add this logic to the existing storage.go test, since these files seemed to verify distinct aspects of OSD (verifying existing storageclasses/volumes vs modifying a gcp project & creating a new storageclass). This can be changed if it's deemed more desirable to keep all storage-related tests in a single file.


pkg/e2e/verify/encrypted_storage.go
```diff
@@ -0,0 +1,354 @@
...
+				TypeMeta: metav1.TypeMeta{
+					Kind: "PersistentVolumeClaim",
+					APIVersion: "core/v1",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name: testInstanceName,
+					Namespace: h.CurrentProject(),
+				},
+				Spec: corev1.PersistentVolumeClaimSpec{
+					StorageClassName: &testInstanceName,
```
dustman9000, line 133: We need to ensure this StorageClass is coming from the pd-ssd provisioner, not "standard" or "default" storageclass.

pkg/e2e/verify/encrypted_storage.go
```diff
@@ -0,0 +1,354 @@
...
+			pvc, err := h.Kube().CoreV1().PersistentVolumeClaims(h.CurrentProject()).Create(context.TODO(), &corev1.PersistentVolumeClaim{
+				TypeMeta: metav1.TypeMeta{
+					Kind: "PersistentVolumeClaim",
+					APIVersion: "core/v1",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name: testInstanceName,
+					Namespace: h.CurrentProject(),
+				},
+				Spec: corev1.PersistentVolumeClaimSpec{
```
dustman9000, line 132: Am I missing something? It doesn't seem like you are passing a CMK to this StorageClass...

tnierman: The key is passed to the storageclass here https://github.com/openshift/osde2e/blob/76e95160d422f6d200ef73d784c06009405ae4f9/pkg/e2e/verify/encrypted_storage.go#L112, and verified to be used to encrypt the disk here https://github.com/openshift/osde2e/blob/76e95160d422f6d200ef73d784c06009405ae4f9/pkg/e2e/verify/encrypted_storage.go#L203 (I figured it'd be a bit out of the scope of this test to verify that Google is telling the truth when it claims a PV is encrypted with a specific key :) ).

pkg/e2e/verify/encrypted_storage.go
tnierman: I *believe* that if the storageclass already exists in OSD on cluster creation, it would be tested to work here: https://github.com/openshift/osde2e/blob/76e95160d422f6d200ef73d784c06009405ae4f9/pkg/e2e/verify/storage.go#L45

Do you mean that the test should instead copy & modify this class to create encrypted disks? If so, I can certainly update to make the storageclass used here to conform more closely to the one you linked

dustman9000: I think the "standard" SC encrypted test is fine. We can add the "ssd" encrypted test in a subsequent PR.

dustman9000:
Really nice work @tnierman!
/lgtm

PR #855, opened 2021-06-11
Sync entries from MCC MUO ConfigMap

ravitri:
Based on PR https://github.com/openshift/managed-cluster-config/pull/853, adding entries in MUO Configmap with more entries for ignoring namespaces. Also, based on latest MCC configmap [file](https://github.com/openshift/managed-cluster-config/blob/master/deploy/managed-upgrade-operator-config/10-managed-upgrade-operator-configmap.yaml), changed the entries of some of the alerts to be ignored during control plane upgrade.

### DIFF
--- a/assets/upgrades/config.template
+++ b/assets/upgrades/config.template
@@ -6,10 +6,9 @@ maintenance:
   controlPlaneTime: {{ .ControlPlaneTime }}
   ignoredAlerts:
   controlPlaneCriticals:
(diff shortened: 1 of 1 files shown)


mrbarge:
The lack of this change is actively breaking our ability to successfully run/finish E2E upgrades right now (an oversight on our part, sorry!), so I'd like this in ASAP.
/approve

PR #856, opened 2021-06-11
Fixing #855 to ignore namespaces

ravitri:
Fixing #855 to move ignore projects under healthcheck.
cc @mrbarge


mrbarge:
Yep, this is good. :+1: 
/lgtm
/approve

PR #861, opened 2021-06-14
Use the test results database to control pagerduty alert management

whereswaldon:
This commit makes our database of test results into the single source of truth for what we do in pagerduty.

/hold

I want to see how the verify job behaves before un-holding this. I tested as much as I could locally.

/assign @jeefy


whereswaldon:
/hold cancel

Let's see how this goes!

PR #866, opened 2021-06-16
SDCICD-651: Add create/destroy scripts for aro cluster creation

jeefy:
These scripts handle installation of requirements for, and creation/deletion of, ARO clusters. They make similar assumptions that the OCP scripts use. Once a cluster is created it generates and outputs a Kubeconfig for later steps to use without need of the Azure CLI/creds.

/assign @whereswaldon @jewzaam 

/hold

I explicitly want both @whereswaldon's and @jewzaam's reviews before this gets merged in.

### DIFF
--- a/ci/create-aro-cluster.sh
+++ b/ci/create-aro-cluster.sh
@@ -1,0 +1,75 @@
+#!/usr/bin/env bash
+
+echo "Installing Azure CLI"
+
+curl -L https://aka.ms/InstallAzureCli | bash
+
+echo "Installing oc binary"
+
+curl -s https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/openshift-client-linux.tar.gz | tar zxvf - oc
+
+chmod +x oc
+
+echo "Setting config variables"
+
+CLUSTER_NAME="osde2e-$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 6 | head -n 1)"
+
+echo $CLUSTER_NAME > ${SHARED_DIR}/cluster-name
+
+LOCATION=$2
+PULL_SECRET_FILE="/usr/local/osde2e-credentials/stage-ocm-pull-secret"
+ 
+if [ "$LOCATION" == "" ];
+then
+    LOCATION="eastus"
+fi
+
+# just automating the steps in https://docs.microsoft.com/en-us/azure/openshift/tutorial-create-cluster
+
+echo "Logging into Azure"
+
+az login --service-principal --username "$(cat /usr/local/osde2e-credentials/aro-app-id)" --password "$(cat /usr/local/osde2e-credentials/aro-password)" --tenant "$(cat /usr/local/osde2e-credentials/aro-tenant)"
+
+echo "Creating required Azure objects"
+
+RESOURCEGROUP_NAME=$CLUSTER_NAME
+
+az provider register -n Microsoft.RedHatOpenShift --wait
+az provider register -n Microsoft.Compute --wait
+az provider register -n Microsoft.Storage --wait
+az provider register -n Microsoft.Authorization --wait
+
+az group create \
+    --name $RESOURCEGROUP_NAME \
+    --location $LOCATION
+    
+az network vnet create \
+    --resource-group $RESOURCEGROUP_NAME \
+    --name aro-vnet \
+    --address-prefixes 10.0.0.0/22
+    
+az network vnet subnet create \
+    --resource-group $RESOURCEGROUP_NAME \
+    --vnet-name aro-vnet $CLUSTER_NAME \
+    --resource-group $RESOURCEGROUP_NAME \
+    --vnet-name aro-vnet \
+    --disable-private-link-service-network-policies true
+
+CREATE_CMD="az aro create --resource-group $RESOURCEGROUP_NAME --name $CLUSTER_NAME --vnet aro-vnet --master-subnet master-subnet --worker-subnet worker-subnet "
+
+if [ "$PULL_SECRET_FILE" != "" ];
+then
+    CREATE_CMD="$CREATE_CMD --pull-secret @$PULL_SECRET_FILE"
+fi
+
+echo "Running ARO create command"
+
+AROINFO="$(eval "$CREATE_CMD")"
+KUBEAPI=$(echo "$AROINFO" | jq -r '.apiserverProfile.url')
+KUBECRED=$(az aro list-credentials --name $CLUSTER_NAME --resource-group $CLUSTER_NAME)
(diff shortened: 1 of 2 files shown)


ci/create-aro-cluster.sh
```diff
@@ -0,0 +1,75 @@
...
+if [ "$LOCATION" == "" ];
+then
+    LOCATION="eastus"
+fi
+
+# just automating the steps in https://docs.microsoft.com/en-us/azure/openshift/tutorial-create-cluster
+
+echo "Logging into Azure"
+
+az login --service-principal --username "$(cat /usr/local/osde2e-credentials/aro-app-id)" --password "$(cat /usr/local/osde2e-credentials/aro-password)" --tenant "$(cat /usr/local/osde2e-credentials/aro-tenant)"
```
T0MASD, line 31: should this better be set as an environment variable?

ci/destroy-aro-cluster.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+echo "Installing Azure CLI"
+
+curl -L https://aka.ms/InstallAzureCli | bash
+
+echo "Logging into Azure"
+
+az login --service-principal --username "$(cat /usr/local/osde2e-credentials/aro-app-id)" --password "$(cat /usr/local/osde2e-credentials/aro-password)" --tenant "$(cat /usr/local/osde2e-credentials/aro-tenant)"
```
T0MASD, line 9: should osde2e-credentials better be set as an environment variable?

ci/destroy-aro-cluster.sh
```diff
@@ -0,0 +1,13 @@
...
+
+echo "Deleting ARO cluster $(cat {$SHARED_DIR}/cluster-name)"
+
+az aro delete --name="$(cat {$SHARED_DIR}/cluster-name)" --resource-group="$(cat {$SHARED_DIR}/cluster-name)"
```
T0MASD, line 13:
```suggestion
az aro delete --name="$(cat {$SHARED_DIR}/cluster-name)" --resource-group="$(cat {$SHARED_DIR}/cluster-name)"

```

ci/create-aro-cluster.sh
```diff
@@ -0,0 +1,75 @@
...
+
+AROINFO="$(eval "$CREATE_CMD")"
+KUBEAPI=$(echo "$AROINFO" | jq -r '.apiserverProfile.url')
+KUBECRED=$(az aro list-credentials --name $CLUSTER_NAME --resource-group $CLUSTER_NAME)
+KUBEUSER=$(echo "$KUBECRED" | jq -r '.kubeadminUsername')
+KUBEPASS=$(echo "$KUBECRED" | jq -r '.kubeadminPassword')
+
+oc login "$KUBEAPI" --username="$KUBEUSER" --password="$KUBEPASS"
+
+oc config view --raw > ${SHARED_DIR}/kubeconfig
```
T0MASD, line 75:
```suggestion
oc config view --raw > ${SHARED_DIR}/kubeconfig

```

ci/create-aro-cluster.sh
whereswaldon: I'm not trying to say what's right or wrong, but there are security benefits to reading it from a file, most notably that the env isn't inherited by subprocesses that don't need the creds and that the creds aren't visible in `/proc`.

ci/create-aro-cluster.sh
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env bash
```
whereswaldon, line 1: So whenever I mint a new bash script, I try to use at least `set -euo pipefail` from this little gem of an article: http://redsymbol.net/articles/unofficial-bash-strict-mode/

It makes the script much more predictable, and because it's new, you're not going to break a behavior that the existing logic relied upon.

If you think there's enough value in that article's approach, would you consider applying that `set` line and retesting to ensure that the script behaves as expected?

ci/create-aro-cluster.sh
```diff
@@ -0,0 +1,75 @@
...
+
+echo "Installing Azure CLI"
+
+curl -L https://aka.ms/InstallAzureCli | bash
```
whereswaldon, line 5: I take it that there isn't a better-trusted way to acquire this? There's a whole beautiful thread on memo-list about this very thing and how scary it is.

ci/create-aro-cluster.sh
```diff
@@ -0,0 +1,75 @@
...
+CREATE_CMD="az aro create --resource-group $RESOURCEGROUP_NAME --name $CLUSTER_NAME --vnet aro-vnet --master-subnet master-subnet --worker-subnet worker-subnet "
+
+if [ "$PULL_SECRET_FILE" != "" ];
+then
+    CREATE_CMD="$CREATE_CMD --pull-secret @$PULL_SECRET_FILE"
+fi
+
+echo "Running ARO create command"
+
+AROINFO="$(eval "$CREATE_CMD")"
```
whereswaldon, line 67: I think that you can get in trouble with word splitting here. The way that the `CREATE_CMD` is constructed, it doesn't contain quotes around `$RESOURCEGROUP_NAME` or `$CLUSTER_NAME`. I'm sure that we expect them not to contain whitespace, semicolons, or other things that will cause bash to mis-parse, but I think it'd be a little safer to embed quotes around them in the variable definition.

```bash
CREATE_CMD="az aro create --resource-group \"$RESOURCEGROUP_NAME\" --name \"$CLUSTER_NAME\" --vnet aro-vnet --master-subnet master-subnet --worker-subnet worker-subnet "
```

You'd also need to do that for the pull secret part.

ci/destroy-aro-cluster.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
```
whereswaldon, line 1: insert the same thoughts around bash strict mode here

whereswaldon (changes requested):
Looking good! Just a few random scripting thoughts.

ci/create-aro-cluster.sh
jeefy: The way Prow mounts secrets into a container is as a file, so there's not really a choice here :)

ci/create-aro-cluster.sh
jeefy: Yep I like that.

ci/create-aro-cluster.sh
jeefy: There are, depending on the distro. But this future proofs any base-image changes and it's supported via the Azure docs so I'm comfortable with it.

ci/destroy-aro-cluster.sh
jeefy: ACK

whereswaldon (commented):
/lgtm

PR #870, opened 2021-06-21
Bump ginkgo version to 1.16.4

jeefy:


whereswaldon (approved):
/lgtm

whereswaldon:
For posterity, this is our attempt to fix a common failure we're seeing in osde2e upgrade testing, namely `You may only call BeforeEach from within a Describe, Context or When`. We see this error attributed like so:
```
/go/src/github.com/openshift/osde2e/pkg/e2e/e2e.go:79
You may only call BeforeEach from within a Describe, Context or When
/go/src/github.com/openshift/osde2e/pkg/common/helper/helper.go:47
```
It appears to be some kind of pathological interaction between ginkgo's lifecycle and osde2e's upgrade testing lifecycle. We are hoping that it's simply a bug that may be fixed by the most recent release. If not, we're going to need to spend even more hours digging through the logic hunting for it.

PR #872, opened 2021-06-22
Document the database package development flow

whereswaldon:
/assign @jeefy

### DIFF
--- a/pkg/db/db_helpers_test.go
+++ b/pkg/db/db_helpers_test.go
@@ -54,7 +54,7 @@ func getDBConfig(t *testing.T) dbConfig {
 	if dbPool != nil {
 		const password = "secret"
 		// pulls an image, creates a container based on it and runs it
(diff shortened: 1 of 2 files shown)


pkg/db/db_helpers_test.go
```diff
@@ -54,7 +54,7 @@ func getDBConfig(t *testing.T) dbConfig {
 	if dbPool != nil {
 		const password = "secret"
 		// pulls an image, creates a container based on it and runs it
-		resource, err := dbPool.Run("postgres", "13", []string{"POSTGRES_PASSWORD=" + password})
```
jeefy, line 57: Was this change intentional?

whereswaldon: Yup, this is the version we have in RDS. I noticed the mismatch and corrected it, but only mentioned that in a commit message.

jeefy: :+1:

PR #876, opened 2021-06-25
fix: track upgrade versions within job database

jeefy:
We need to track the upgrade version for jobs within the database.


whereswaldon (approved):
Very nice! Looks great.

/lgtm

PR #885, opened 2021-07-14
Add a selection of default OSD namespaces to inspect on cleanup

mrbarge:
This is a continuation of PR #881 to add a configurable list of OSD projects to perform an `oc adm inspect` on during E2E cleanup.

This PR defines a default collection of namespaces which are normally present on OSD clusters.

The process for performing the `oc adm inspect` has also been improved such that it will verify each project in the "to be inspected" list actually exists on the cluster before it attempts to inspect it (preventing any inadvertent error in the process executing).

### DIFF
--- a/pkg/common/config/config.go
+++ b/pkg/common/config/config.go
@@ -3,6 +3,7 @@ package config
 
 import (
 	"log"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


jeefy:
Awesome!
/lgtm

PR #886, opened 2021-07-17
Custom Domains: Handle DNS Errors During HTTP Response And Adjust Timeouts

dustman9000:
Handle URL Error and DNS Error "no such host" during HTTP response test phase and adjust timeouts.

https://issues.redhat.com/browse/OSD-7472


dustman9000:
After further testing, it looks like if we handle the DNS "no such host" error during the HTTP response phase, the test should be much more reliable.

/hold cancel

PR #891, opened 2021-08-04
Accept managed upgrades using image digest

mrbarge:
### What type of PR is this?

Feature

### What this PR does / why we need it?

The `managed-upgrade-operator` now supports upgrades using an image digest (done for future ARO support).

This PR's primary intent is to change `osde2e` to use the `managed-upgrade-operator` for performing upgrades that would normally be triggered by setting the `Upgrade.Image` (`UPGRADE_IMAGE` env var) configuration. No functionality has been removed through this change, just replacing the underlying means by which image-based upgrades are undertaken.

Doing this allows us to deprecate and remove a large portion of `osde2e`-related code devoted to supporting the image-based upgrades by patching `clusterversion`. Now, all upgrade-related functions can be performed via the `managed-upgrade-operator` operator. Because of this, the following configuration items have been removed:

- `Upgrade.ManagedUpgrade`
- `Upgrade.WaitForWorkersToManagedUpgrade`
(no configs in `configs/` were using these configuration items)

The inclusion of this feature also allows preliminary support for allowing `osde2e` to support upgrading ARO clusters if need be. This can be controlled through the inclusion of a new `Upgrade.Type` configuration (or `UPGRADE_TYPE` env var). By default this is set to `OSD`, but if set to `ARO` it will inform the `managed-upgrade-operator` to apply its ARO-specific upgrade routines.

(The next planned follow-up from this PR will be introduction of something like a `UPGRADE_TO_LATEST_IMAGE` config that can drive an upgrade using an auto-detected image, so that we can have a permanent image-based upgrade test job into Prow.)

### Which Jira/Github issue(s) this PR fixes?

[OSD-7724](https://issues.redhat.com/browse/OSD-7724)

### Pre-checks (if applicable):
- [x] Ran a full test run (including cluster creation and upgrade) successfully, upgrading using OCM/release version.
- [x] Ran a full test run (including upgrade, but not cluster creation) successfully, upgrading using image digest.
- [x] Included documentation changes with PR


pkg/common/upgrade/managed.go
```diff
@@ -483,3 +471,112 @@ func isUpgradeTriggered(h *helper.H, desired *configv1.Update) (bool, error) {
...
+		if err != nil {
+			return fmt.Errorf("unable to setup node drain test workload for upgrade: %v", err)
+		}
+	}
+
+	return nil
+}
+
+// determine which policy provider to use for an upgrade policy source
+func getPolicyProvider() string {
```
T0MASD, line 501: could this be named `getProviderSource`?

pkg/common/upgrade/managed.go
```diff
@@ -125,20 +99,29 @@ func TriggerManagedUpgrade(h *helper.H) (*configv1.ClusterVersion, error) {
...
+	cUpdate := &configv1.Update{}
+
+	releaseName := viper.GetString(config.Upgrade.ReleaseName)
+	if releaseName != "" {
+		upgradeVersion, err := util.OpenshiftVersionToSemver(releaseName)
+		if err != nil {
+			cUpdate.Version = upgradeVersion.String()
+		}
+	}
+	image := viper.GetString(config.Upgrade.Type)
```
bmeng, line 119: Should be `config.Upgrade.Image` here?

pkg/common/config/config.go
```diff
@@ -148,10 +146,9 @@ var Upgrade = struct {
 	UpgradeToLatestY:                       "upgrade.ToLatestY",
 	ReleaseName:                            "upgrade.releaseName",
 	Image:                                  "upgrade.image",
+	Type:									"upgrade.type",
```
bmeng, line 149: this is not well aligned?

mrbarge: Fixed :+1:

pkg/common/upgrade/managed.go
mrbarge: Thanks, nice catch!!
This is now fixed.

pkg/common/upgrade/managed.go
mrbarge: Makes sense - function now renamed.

dofinn:
@jeefy looks good to me in regards to upgrade logic/flows. Have placed in hold in case you want to review :)

jeefy:
/hold cancel

Thanks all!

PR #900, opened 2021-09-17
SDCICD-607: Support creating and testing against STS clusters

jeefy:
Adds support for creating ROSA STS clusters.

Updates CIO and MVO tests to be skipped when it's an STS cluster.

Currently blocked by [SDA-4806](https://issues.redhat.com/browse/SDA-4806)
/hold

### DIFF
--- a/pkg/common/providers/ocmprovider/cluster.go
+++ b/pkg/common/providers/ocmprovider/cluster.go
@@ -294,6 +294,9 @@ func (o *OCMProvider) FindRecycledCluster(originalVersion, cloudProvider, produc
 				return ""
 			}
 			viper.Set(config.Cluster.Reused, true)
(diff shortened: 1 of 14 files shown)


pkg/common/providers/ocmprovider/cluster.go
```diff
@@ -294,6 +294,9 @@ func (o *OCMProvider) FindRecycledCluster(originalVersion, cloudProvider, produc
 				return ""
 			}
 			viper.Set(config.Cluster.Reused, true)
+			if recycledCluster.AWS().STS().RoleARN() != "" {
+				viper.Set("rosa.STS", true)
+			}
```
jharrington22, line 299: nit but can you do this just once above this block if its a recycled cluster?

vkareh:
This LGTM

PR #901, opened 2021-09-20
Add workaround for etcd operator

NautiluX:
* etcd operator sometimes reports RecentBackup condition as Unknown
* This is not critical for the operator but blocks the test run
* See [Bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=2005952)


NautiluX:
/hold
still testing locally

PR #902, opened 2021-09-24
Allow condition unknown

NautiluX:


pkg/common/cluster/healthchecks/operators.go
```diff
@@ -46,18 +46,15 @@ func CheckOperatorReadiness(configClient configclient.ConfigV1Interface, logger
 	for _, co := range list.Items {
 		if _, ok := operatorSkipList[co.GetName()]; !ok {
 			for _, cos := range co.Status.Conditions {
-				if cos.Type == "Disabled" && cos.Status == "True" {
+				if cos.Status == "Unknown" || cos.Status == "False" {
```
jharrington22, line 49: Do we need to be specific about which operator status we're looking at with the `Unknown` condition?

jeefy:
I would like to, since it would be good to know if other Operators start
exhibiting this behavior.

pkg/common/cluster/healthchecks/operators.go
jeefy: I would like to, since it would be good to know if other Operators start
exhibiting this behavior.

NautiluX: that's what we had before, but as far as I understood CVO team said Unknown shall be considered a good state.

PR #905, opened 2021-09-30
OSD-8357 - Improve repeatability for encrypted storage testing.

tnierman:
Resolves: https://issues.redhat.com/browse/OSD-8357

---

Encrypted Storage tests were found to be failing because the e2e pipeline was running the test multiple times, which caused a conflict between old resources that weren't cleaned up properly and newly generated resources.  

To solve this problem, `CredentialsRequest` cleanup was added to the `AfterEach` function, the naming logic was improved to minimize the chance of collisions between tests, and logic was added to reuse GCP `cryptoKeys` if they exist already (`cryptoKeys` cannot be deleted from KMS, therefore they must be reused rather than recreated).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -34,7 +34,7 @@ require (
 	github.com/markbates/pkger v0.17.1
 	github.com/mitchellh/mapstructure v1.1.2
 	github.com/onsi/ginkgo v1.16.4
-	github.com/onsi/gomega v1.10.2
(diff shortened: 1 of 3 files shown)


yithian:
fwiw, this looks sane to me

tnierman:
To be clear, I have had some trouble reproducing the failures during my own testing, so these changes may not be entirely sufficient. My goal is to get thsi

PR #909, opened 2021-10-11
Add etcd conditions which can be True

NautiluX:


NautiluX:
cc @lilic

pkg/common/cluster/healthchecks/operators.go
```diff
@@ -49,7 +49,7 @@ func CheckOperatorReadiness(configClient configclient.ConfigV1Interface, logger
 				if cos.Status == "Unknown" || cos.Status == "False" {
 					continue
 				}
-				if cos.Type == "Disabled" || cos.Type == "Available" || cos.Type == "Upgradeable" {
+				if cos.Type == "Disabled" || cos.Type == "Available" || cos.Type == "Upgradeable" || cos.Type == "RecentBackup" || cos.Type == "UpgradeBackupSuccessful" {
```
lilic, line 52: UpgradeBackupSuccessful is the reason, not the type so not needed here. But you can do something like:

`if cos.Type == "RecentBackup" && cos.Reason == "UpgradeBackupSuccessful" `

NautiluX: oooh understood. I don't think we explicitly need to check for the Reason though.

PR #919, opened 2021-10-29
[OSD-7617] Implements e2e tests for cloud-ingress-operator

clcollins:
PR implements @lnguyen1401's e2e tests for the cloud-ingress-operator

e2e tests implemented:
    default ingresscontroller should get deleted and recreated with desired result if publishinstrategy.spec.applicationIngress.listening changes.
    1. Change publishinstrategy.spec.applicationIngress.listening to Internal
    2. check if the Generation == 1, which means it has been deleted and recreated
    3. Change publishinstrategy.spec.applicationIngress.listening to External
    4. check if the Generation == 1, which means it has been deleted and recreated
    default ingresscontroller should get deleted and recreated with desired result if publishinstrategy.spec.applicationIngress.dnsName changes
    1. Change dnsName
    2. check if the Generation == 1, which means it has been deleted and recreated
    3. return dnsName to the original name
    4. check if the Generation == 1, which means it has been deleted and recreated
    Default ingresscontroller should be patched and NOT deleted if publishinstrategy.spec.applicationIngress.certificate changes.
    1. change certificate name
    2. check if the Generation > 1, which means it has been patched
    3. change the certificate name to the original
    4. check if the Generation > 1, which means it has been patched
    Default ingresscontroller should be patched and NOT deleted if publishinstrategy.spec.applicationIngress.routeSelector changes
    1. change routeSelector name
    2. check if the Generation > 1, which means it has been patched
    3. change the routeSelector name to the original
    4. check if the Generation > 1, which means it has been patched
    every ingresscontroller created by the cloud-ingress-operator should have an annotation “Owner: cloud-ingress-operator”
    remove the annotation from apps2 ingresscontroller, then delete apps2 from publishingstrategy, apps2 ingresscontroller should not get deleted. This is because we only want to watch the ingress that we (CIO) owns. (app2 is in secondary router)
    1. remove "cloud-ingress-controller" annotation from apps2 ingresscontroller -> this makes apps2 ingresscontroller not belong to the publishingstrategy anymore
    2. remove apps2 from publishingstrategy -> normally id the ingresscontroller belongs to the publishing strategy, it would delete the ingresscontroller. but since apps2 doesn't belong to the publishingstrategy anymore, it shouldn't be deleted.
    3. clean up the test secondary ingresscontroller: re-add "owner: cloud-ingress-operator" back to the apps2 ingresscontroller. re-add apps2 ingresscontroller to publishingstrategy. this makes the CIO able to delete the apps ingresscontroller for clean-up purpose.
    
REF:  https://issues.redhat.com/browse/OSD-7617

### DIFF
--- a/pkg/e2e/operators/cloudingress/publishingstrategies_cr.go
+++ b/pkg/e2e/operators/cloudingress/publishingstrategies_cr.go
@@ -39,6 +39,8 @@ var _ = ginkgo.Describe(constants.SuiteOperators+TestPrefix, func() {
 			err := addPublishingstrategy(h, ps)
 			Expect(apierrors.IsForbidden(err)).To(BeTrue())
 
+			//ingress, _ := getingressController(h, "default")
+			//Expect(ingress.Annotations["Owner"]).To(Equal("cloud-ingress-operator"))
 		})
 
 		ginkgo.It("cluster admin should be allowed to manage publishingstrategies CR", func() {
@@ -49,8 +51,11 @@ var _ = ginkgo.Describe(constants.SuiteOperators+TestPrefix, func() {
 				publishingstrategyCleanup(h, publishingstrategyName)
 			}()
 			Expect(err).NotTo(HaveOccurred())
-
+			//check the annotation for owned ingress
+			//ingress, _ := getingressController(h, "default")
+			//Expect(ingress.Annotations["Owner"]).To(Equal("cloud-ingress-operator"))
 		})
+
 	})
 
 })
--- a/pkg/e2e/operators/cloudingress/routeSelector.go
+++ b/pkg/e2e/operators/cloudingress/routeSelector.go
@@ -1,0 +1,142 @@
+package cloudingress
+
+import (
+	"context"
+	"log"
+	"reflect"
+	"time"
+
+	"github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/openshift/osde2e/pkg/common/constants"
+	"github.com/openshift/osde2e/pkg/common/helper"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// tests
+var _ = ginkgo.Describe(constants.SuiteInforming+TestPrefix, func() {
+	h := helper.New()
+
+	ginkgo.Context("publishingstrategy-route-selector", func() {
+		ginkgo.It("IngressController should be patched when update routeSelector matchLabels", func() {
+			log.Print("Gonna Update the RouteSelector to add MatchLabels")
+			updateMatchLabels(h, "tier", "frontend")
+
+			time.Sleep(time.Duration(60) * time.Second)
+			log.Print("Updated the RouteSelector in Publishingstrategy")
+			ingress, _ := getingressController(h, "default")
+			temp := (int64(1))
+			Expect(string(ingress.Spec.RouteSelector.MatchLabels["tier"])).To(Equal("frontend"))
+			Expect(ingress.Generation == temp).To(Equal(false))
+			Expect(ingress.Annotations["Owner"]).To(Equal("cloud-ingress-operator"))
+			log.Print("Ok Done with the test")
+		})
+		ginkgo.It("IngressController should be patched when update routeSelector matchExpressions", func() {
+			log.Print("Gonna Update the RouteSelector to add MatchExpressions")
+			updateMatchExpressions(h, "foo", "In", "bar")
+
+			time.Sleep(time.Duration(60) * time.Second)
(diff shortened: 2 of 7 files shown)


mrbarge:
Looks pretty good! I have a few general stylistic comments to make which apply to several parts of the PR.

Try to remove the number of `log.Print`s. Much of what the test is doing or checking should be conveyed through the `Expect()` calls. The log messages indicating when a test has started or ended are particularly superfluous and don't need to be included.

If there's logging that is felt to be essential, maybe use it as an opportunity to output particular variable values at the same time. ie.
```
// Instead of this:
log.Print("Got ingresscontroller Name")
// This:
log.Printf("Got ingresscontroller Name: %s", ingressControllerName)
```

The various `sleep()` calls can also potentially be problematic - E2E runs already take quite a long time, and unnecessary waiting can eat into the overall job run time. Instead of sleeping, you can instead use the [k8s.io `wait` helpers](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait) to repeatedly poll for a condition and optionally fail out if an error occurs. [See here](https://github.com/openshift/osde2e/blob/e12a5a3c16f37253a96fb3743beb4d3199bce703/pkg/e2e/operators/operators.go#L288) for an example of it being used in E2E code. The gomega `Eventually()` assertion can also be helpful here, here is a [usage example](https://github.com/openshift/osde2e/blob/e12a5a3c16f37253a96fb3743beb4d3199bce703/pkg/e2e/operators/operators.go#L262). I'd recommend replacing the sleeps with something more like this, wherever possible.

Another potential source of trouble can come when performing `Update()` on objects. If the object happens to have been modified by something else on-cluster in-between the test first reading it and the test's subsequent update, the `Update()` will error, and that'll appear as a test failure. It is safer to surround the `Get()` of the cluster object and the `Update()` of it using the k8s client-go's `retry.retryOnConflict`, an example of which [can be seen here](https://github.com/openshift/osde2e/blob/7ccac0251ee09c455dc18646f267c09a40de1e4d/pkg/e2e/osd/dedicatedadmin.go#L386). This isn't essential, but adopting it makes for good, robust, reliable tests.

Those cover the general observations, for anything more specific I'll add against the actual code line(s).

pkg/e2e/operators/cloudingress/secondary_router.go
```diff
@@ -69,16 +71,16 @@ var _ = ginkgo.Describe(constants.SuiteInforming+TestPrefix, func() {
 				time.Sleep(time.Duration(120) * time.Second)
 			}
 
-			ingressControllerName := strings.Split(secondaryIngress.DNSName, ".")[0]
+			ingressControllerName := strings.Split(secondaryIngress.DNSName, ".")[1]
```
mrbarge, line 74: To prevent a potential "index out of range" panic occurring here, it will be safer to `Expect()` that the length of the split array is > 1 first.

pkg/e2e/operators/cloudingress/routeSelector.go
```diff
@@ -0,0 +1,142 @@
...
+			log.Print("Gonna Update the RouteSelector to add MatchExpressions")
+			updateMatchExpressions(h, "foo", "In", "bar")
+
+			time.Sleep(time.Duration(60) * time.Second)
+			log.Print("Updated the RouteSelector.MatchExpressions in Publishingstrategy")
+			ingress, _ := getingressController(h, "default")
+			temp := (int64(1))
+			//tempVal := []string{"bar"}
+			//tempOp := metav1.LabelSelectorOperator("In")
+			//temp1 := metav1.LabelSelectorRequirement{"foo", tempOp, tempVal}
```
mrbarge, line 46: Can be removed.

pkg/e2e/operators/cloudingress/publishingstrategies_cr.go
```diff
@@ -49,8 +51,11 @@ var _ = ginkgo.Describe(constants.SuiteOperators+TestPrefix, func() {
 				publishingstrategyCleanup(h, publishingstrategyName)
 			}()
 			Expect(err).NotTo(HaveOccurred())
-
+			//check the annotation for owned ingress
+			//ingress, _ := getingressController(h, "default")
+			//Expect(ingress.Annotations["Owner"]).To(Equal("cloud-ingress-operator"))
```
mrbarge, line 56: Can be removed.

pkg/e2e/operators/cloudingress/secondary_router.go
lnguyen1401: Done

pkg/e2e/operators/cloudingress/routeSelector.go
lnguyen1401: done

MrSantamaria:
@mrbarge @clcollins This PR is old, is this still something that we want to add? 
If so could we get a rebase.

mrbarge:
I think these tests were eventually implemented and merged in #932
This seems safe to close

PR #922, opened 2021-11-03
 Fixed pr osd 7617

lnguyen1401:
Fixed PR according to comments from https://github.com/openshift/osde2e/pull/919.

### DIFF
--- a/pkg/e2e/operators/cloudingress/certificate.go
+++ b/pkg/e2e/operators/cloudingress/certificate.go
@@ -1,0 +1,65 @@
+package cloudingress
+
+import (
+	"context"
+	"time"
+
+	"github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/openshift/osde2e/pkg/common/constants"
+	"github.com/openshift/osde2e/pkg/common/helper"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// tests
+var _ = ginkgo.Describe(constants.SuiteInforming+TestPrefix, func() {
+	h := helper.New()
+	var originalCert string
+	ginkgo.Context("publishingstrategy-certificate", func() {
+		ginkgo.It("IngressController should be patched when update Certificate", func() {
+			ingress1, _ := getingressController(h, "default")
+			originalCert = string(ingress1.Spec.DefaultCertificate.Name)
+			updateCertificate(h, "foo-bar")
+			time.Sleep(time.Duration(60) * time.Second)
+			ingress, _ := getingressController(h, "default")
+
+			Expect(string(ingress.Spec.DefaultCertificate.Name)).To(Equal("foo-bar"))
+			Expect(ingress.Generation == int64(1)).To(Equal(false))
+			Expect(ingress.Annotations["Owner"]).To(Equal("cloud-ingress-operator"))
+		})
+		ginkgo.It("IngressController should be patched when return the original Certificate", func() {
+			updateCertificate(h, originalCert)
+			time.Sleep(time.Duration(60) * time.Second)
+			ingress, _ := getingressController(h, "default")
+			Expect(string(ingress.Spec.DefaultCertificate.Name)).To(Equal(originalCert))
+			Expect(ingress.Generation == int64(1)).To(Equal(false))
+			Expect(ingress.Annotations["Owner"]).To(Equal("cloud-ingress-operator"))
+		})
+	})
+})
+
+func updateCertificate(h *helper.H, newName string) {
+	var err error
+	PublishingStrategyInstance, ps := getPublishingStrategy(h)
+
+	// Grab the current list of Application Ingresses from the Publishing Strategy
+	AppIngress := PublishingStrategyInstance.Spec.ApplicationIngress
+	name := newName
+	// Find the default router and update its scheme
+	for i, v := range AppIngress {
+		if v.Default == true {
+			AppIngress[i].Certificate.Name = name
+		}
+	}
+
+	PublishingStrategyInstance.Spec.ApplicationIngress = AppIngress
+
+	ps.Object, err = runtime.DefaultUnstructuredConverter.ToUnstructured(&PublishingStrategyInstance)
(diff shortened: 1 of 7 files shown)


lnguyen1401:
e2e tests implemented:
default ingresscontroller should get deleted and recreated with desired result if publishinstrategy.spec.applicationIngress.listening changes.
1. Change publishinstrategy.spec.applicationIngress.listening to Internal
2. check if the Generation == 1, which means it has been deleted and recreated
3. Change publishinstrategy.spec.applicationIngress.listening to External
4. check if the Generation == 1, which means it has been deleted and recreated
default ingresscontroller should get deleted and recreated with desired result if publishinstrategy.spec.applicationIngress.dnsName changes
1. Change dnsName
2. check if the Generation == 1, which means it has been deleted and recreated
3. return dnsName to the original name
4. check if the Generation == 1, which means it has been deleted and recreated
Default ingresscontroller should be patched and NOT deleted if publishinstrategy.spec.applicationIngress.certificate changes.
1. change certificate name
2. check if the Generation > 1, which means it has been patched
3. change the certificate name to the original
4. check if the Generation > 1, which means it has been patched
Default ingresscontroller should be patched and NOT deleted if publishinstrategy.spec.applicationIngress.routeSelector changes
1. change routeSelector name
2. check if the Generation > 1, which means it has been patched
3. change the routeSelector name to the original
4. check if the Generation > 1, which means it has been patched
every ingresscontroller created by the cloud-ingress-operator should have an annotation “Owner: cloud-ingress-operator”
remove the annotation from apps2 ingresscontroller, then delete apps2 from publishingstrategy, apps2 ingresscontroller should not get deleted. This is because we only want to watch the ingress that we (CIO) owns. (app2 is in secondary router)
1. remove "cloud-ingress-controller" annotation from apps2 ingresscontroller -> this makes apps2 ingresscontroller not belong to the publishingstrategy anymore
2. remove apps2 from publishingstrategy -> normally id the ingresscontroller belongs to the publishing strategy, it would delete the ingresscontroller. but since apps2 doesn't belong to the publishingstrategy anymore, it shouldn't be deleted.
3. clean up the test secondary ingresscontroller: re-add "owner: cloud-ingress-operator" back to the apps2 ingresscontroller. re-add apps2 ingresscontroller to publishingstrategy. this makes the CIO able to delete the apps ingresscontroller for clean-up purpose.

REF: https://issues.redhat.com/browse/OSD-7617

lnguyen1401:
Things fixed from the previous https://github.com/openshift/osde2e/pull/919:

robotmaxtron:
Thank you for the review on #919 @mrbarge, changes made in this PR lgtm but I would like to see the commits squashed.

pkg/e2e/operators/cloudingress/certificate.go
```diff
@@ -0,0 +1,67 @@
...
+		ginkgo.It("IngressController should be patched when update Certificate", func() {
+			ingress1, _ := getingressController(h, "default")
+			originalCert = string(ingress1.Spec.DefaultCertificate.Name)
+			updateCertificate(h, "foo-bar")
+			time.Sleep(time.Duration(60) * time.Second)
+			ingress, _ := getingressController(h, "default")
+
+			temp := (int64(1))
+			Expect(string(ingress.Spec.DefaultCertificate.Name)).To(Equal("foo-bar"))
+			Expect(ingress.Generation == temp).To(Equal(false))
```
fahlmant, line 30:
```suggestion
			Expect(ingress.Generation == int64(1)).To(Equal(false))
```
Does this work? If so, I think this is a bit cleaner than declaring a single use variable.

lnguyen1401: yep i'll fix that !

fahlmant (approved):
/lgtm

pkg/e2e/operators/cloudingress/deletetapps2.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/openshift/osde2e/pkg/common/helper"
+	"github.com/openshift/osde2e/pkg/e2e/operators"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+var _ = ginkgo.Describe(constants.SuiteInforming+TestPrefix, func() {
+	h := helper.New()
+	ginkgo.Context("Delete apps2 ingresscontroller", func() {
```
mrbarge, line 24: Thanks for making all those changes @lnguyen1401 , appreciated! They all look good to me.

I just want to revisit one note from my earlier review around the number of `time.Sleep`s occuring. This test in particular is pretty heavy on them, there's a 30s sleep, a 60s sleep, a 2 minute sleep to delete the ingress controller, a ten minute sleep after that function finishes, then an optional 2 minute sleep again after that. Lots of sleepyness, especially in the case of upgrade jobs, where this test would run twice. :) 

The main issue with this is not just the test doing a lot of nothing for a lof of the time, but more that it eats into the overall job's duration on Prow (which is limited), so wherever we can it's good to get this tighter. 

Is it possible to detect on-cluster the sorts of conditions you're expecting to have occurred after the sleeps and poll for that condition periodically?

ie. Here is how I would consider refactoring the `removeIngressController` to not have a sleep (disclaimer: not tested!):

```
func removeIngressController(h *helper.H, name string) error {
	_, exists, index := appIngressExits(h, false, name)
	// only remove the secondary ingress if it already exist in the publishing strategy
	if exists {
		removeAppIngress(h, index)
	}
	err = wait.PollImmediate(10*time.Second, 2*time.Minute, func() (bool, error) {
		_, err := h.Dynamic().Resource(schema.GroupVersionResource{Group: "operator.openshift.io", Version: "v1", Resource: "ingresscontrollers"}).Namespace("openshift-ingress-operator").Get(context.TODO(), ingressControllerName, metav1.GetOptions{})
		if k8serrors.IsNotFound(err) {
			return true, nil
		}
		return false, err
	})
	return err
}
```

and then back in the calling code:

```
			err = removeIngressController(h, ingressControllerName)
			Expect(err).NotTo(HaveOccurred()
```

Totally up to you if you want to explore this (I'm happy to help more if you're interested). Otherwise I'm okay to lgtm it personally.

cblecker: Yeah I think this would be a good thing to do

PR #930, opened 2021-11-22
Added a retry logic for 20 minutes when attempting to resume a cluster and skipping health checks

MrSantamaria:


pkg/e2e/e2e.go
```diff
@@ -144,9 +145,20 @@ func beforeSuite() bool {
 		}
 
 		var kubeconfigBytes []byte
-		if kubeconfigBytes, err = provider.ClusterKubeconfig(viper.GetString(config.Cluster.ID)); err != nil {
+		clusterConfigerr := wait.PollImmediate(2*time.Second, 5*time.Minute, func() (bool, error) {
+			kubeconfigBytes, err := provider.ClusterKubeconfig(viper.GetString(config.Cluster.ID))
```
jeefy, line 149: This should be `=` not `:=`

By making it `:=` it creates a `kubeconfigBytes` variable only within the scope of the inner PollImmediate function and not honoring the defined variable on line 147

jeefy (changes requested):
Besides the one comment, we should also remove the `viper.Set` on line 165 since it's redundant and should only ever be set in the PollImmediate function :)

Quick change, then the verify job should pass.

pkg/e2e/e2e.go
MrSantamaria: Pr updated good catches!

Issue #921, opened 2021-11-01
compile time error in nodelabels.go, publishingstrategies_cr.go


georgettica:
```
You may only call It from within a Describe, Context or When
```

this seems to be fine when looking at the code, but shows this was not tested locally

https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/osde2e-prod-aws-e2e-upgrade-to-latest/1454870829660639232
https://github.com/openshift/osde2e/blob/e12a5a3c16f37253a96fb3743beb4d3199bce703/pkg/e2e/osd/nodelabels.go#L25

---
similar issue in
https://github.com/openshift/osde2e/blob/e12a5a3c16f37253a96fb3743beb4d3199bce703/pkg/e2e/operators/cloudingress/publishingstrategies_cr.go#L21
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/osde2e-prod-aws-e2e-upgrade-rescheduled/1454991614320578560
though here it shows as flakey

mrbarge:
My current suspicion of why this is happening is it might be potentially related to:

- https://github.com/onsi/ginkgo/issues/485 
- https://github.com/onsi/ginkgo/issues/704

If ginkgo's `RunSpecs()` is called twice then it can lead odd and unpredictable behaviour in the results. This would potentially explain why it's a bit random.

As best I can tell osde2e currently does call it twice, once pre-upgrade and once post-upgrade. Seems to be failing in the second post-upgrade invocation. We do consistently see this across various jobs, with varying tests, and the unifying factor is that they're upgrade-related jobs which lends some credence to the aforementioned theory.

georgettica:
this might be solved by adding a channel of `pre-upgrade` tests complete before the `post-upgrade` ones run?

these issues seems to make sense if this premise is correct, now we need to decide on how to make this go away?

mrbarge:
A potential - although high-effort-investment - option is that ginkgo v2 supports the ability to [repeat test suites](https://github.com/onsi/ginkgo/blob/ver2/docs/NEW_INDEX.md#repeating-test-runs-and-managing-flaky-specs). It will offer this via the ginkgo CLI with a `-repeat` option, but would need some doing to engineer inside OSDE2E.

[...]

PR #936, opened 2021-11-30
Ginkgo V2 Update

jeefy:
This PR updates osde2e to start using the V2 branch of Ginkgo. 

We run into several flakes with test runs using the GinkgoV1, so moving us to (a still yet unreleased) V2 should clean up a significant number of unresolvable flakes. 

The biggest changes in behavior are `ginkgo.It` and the ability to update configs on-the-fly, so that's what most of this PR is actually tackling. See https://github.com/onsi/ginkgo/blob/ver2/docs/MIGRATING_TO_V2.md for more info when reviewing this PR.

/assign @MrSantamaria @cblecker @mrbarge 

Explicitly assigning the three of you. Given the nature of this PR, each of you have a (Taken voice) particular set of skills(TM) I'd like reviewing these changes. 

Quick edit/note: I've run this numerous times locally against various test clusters and the behavior is as desired.


mrbarge:
Oh this is awesome! This may ultimately help us with issue #921 too. I'll give this a good look through, thanks!

mrbarge:
Finally got around to start having a look at this. On the whole, looks fine to me!

But, I did notice a few odd things when actually running it though. Granted I only ran it once, but I noticed a number of tests where the `GinkgoIt` timeout caused failures for taking too long. 

Even though we've had timeouts on the `It` tests prior to this PR, I don't think I've ever seen them actually result in a timeout (I don't know why, was some other part of the code forcing synchronous test running? was it ever working? :confused: ). Until testing this PR where it seems like timeouts actually are being applied properly now.

It was immediately obvious in a couple of tests that didn't have a timeout defined at all, so they started failing after [the new 1 second default](https://github.com/openshift/osde2e/blob/8178a3b3c28a0530b6c8a1639b080b726db19f5f/pkg/common/util/util.go#L49) if no timeout is specified, which is way too short of a time. Timeout-less examples that failed:

- guestbook: https://github.com/openshift/osde2e/blob/8178a3b3c28a0530b6c8a1639b080b726db19f5f/pkg/e2e/workloads/guestbook/guestbook.go#L95
- inhibitions: https://github.com/openshift/osde2e/blob/8178a3b3c28a0530b6c8a1639b080b726db19f5f/pkg/e2e/osd/inhibitions.go#L200
- machinehealthcheck: https://github.com/openshift/osde2e/blob/8178a3b3c28a0530b6c8a1639b080b726db19f5f/pkg/e2e/osd/machinehealthcheck.go#L114  (_definitely_ not gonna run in 1 second..)
- others found in: rhapi_endpoint, namespace_webhook

In addition, many other tests failed for breaching the default 60s timeout too. Many of these tests have logic that involves waiting for up to 5 minutes for particular failing conditions to be met, so it seems at odds with a 60s test timeout. The code is setting itself up for failure.

Granted, many of these issues have been present for quite a while in the code, but it seems like it's only now that the actual timeouts are being exercised and it might weed a lot of these out of the woodwork. I think we're going to have to do a review of the test timeouts and either tune, apply or drop some of them. Which I'm happy to help with, but what do you think about all this first @jeefy ?

PR #938, opened 2021-12-03
SDCICD-706: wait for a latestCSV status before moving forward or fail

jeefy:
This should cause the checkUpgrade function to fail if a latestCSV isn't returned within 5 minutes of looking at an operator's subscription.

### DIFF
--- a/pkg/e2e/operators/operators.go
+++ b/pkg/e2e/operators/operators.go
@@ -255,7 +255,7 @@ func checkUpgrade(h *helper.H, subNamespace string, subName string, packageName 
 				}
 				return false, nil
 			})
-			Expect(pollErr).NotTo(HaveOccurred(), fmt.Sprintf("failed trying to get Subscription %s in %s namespace: %s", subName, subNamespace, err.Error()))
+			Expect(pollErr).NotTo(HaveOccurred(), fmt.Sprintf("failed trying to get Subscription %s in %s namespace: %v", subName, subNamespace, err))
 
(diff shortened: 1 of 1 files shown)


mrbarge:
Thanks so much for this fix! 
/lgtm

pkg/e2e/operators/operators.go
```diff
@@ -240,9 +240,22 @@ func checkUpgrade(h *helper.H, subNamespace string, subName string, packageName
...
+				if err != nil {
+					return false, err
+				}
+				latestCSV = sub.Status.CurrentCSV
+				if latestCSV != "" {
+					return true, nil
+				}
+				return false, nil
+			})
+			Expect(pollErr).NotTo(HaveOccurred(), fmt.Sprintf("failed trying to get Subscription %s in %s namespace: %s", subName, subNamespace, err.Error()))
```
dofinn, line 258: This will try to print a nil pointer `err`

mrbarge: Just re-iterating this (good pick-up @dofinn), because it got lost in the sea of test errors. However, even if it's changed to `pollErr` it will still panic because on upon a successful exit, `pollErr` will be `nil`, the `Sprintf` gets called regardless, and accessing the `Error()` function will panic.

```suggestion
			Expect(pollErr).NotTo(HaveOccurred(), fmt.Sprintf("failed trying to get Subscription %s in %s namespace: %v", subName, subNamespace, err))
```

should fix it up.

PR #940, opened 2021-12-09
OSD-8402 bug(pods.go): account for installer pod retries

Tessg22:
> As an SRE Platform on-call engineer, I do not want to receive alerts the `osd-cluster-ready` job for components that have had a successful retry or follow up.  The cluster ready job currently fails if an installer pod is listed in an Error state, despite retries succeeding.

osde2e tests should now account for installer pod retries. If the pod has no associated cronJob, it should check the Phase and return an error if needed.

I have tested the code in scenarios of the absence of a label, different labels/namespaces/states of pods, and combinations. If you see something that I have missed, please let me know.

### DIFF
--- a/pkg/common/cluster/healthchecks/pods.go
+++ b/pkg/common/cluster/healthchecks/pods.go
@@ -32,7 +32,6 @@ func CheckClusterPodHealth(podClient v1.CoreV1Interface, logger *log.Logger) ([]
 		IsClusterPod,
 		IsNotReadinessPod,
 		IsNotRunning,
-		IsNotCompleted,
 	}
 	podlist, err := checkPods(podClient, logger, filters...)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/common/cluster/healthchecks/pods.go
```diff
@@ -79,12 +78,48 @@ func checkPods(podClient v1.CoreV1Interface, logger *log.Logger, filters ...PodP
...
+		if pod.ObjectMeta.Labels["job-name"] == "" {
+			if pod.Status.Phase != kubev1.PodPending {
+				return nil, fmt.Errorf("Pod %s in unexpected phase %s: reason: %s message: %s", pod.GetName(), pod.Status.Phase, pod.Status.Reason, pod.Status.Message)
+			}
+			logger.Printf("%s is not ready. Phase: %s, Message: %s, Reason: %s", pod.Name, pod.Status.Phase, pod.Status.Message, pod.Status.Reason)
+		}
+		if namespace == "" || namespace != pod.Namespace {
+			filteredPods[pod.Namespace] = map[string]interface{}{}
+		}
+		jobName = strings.LastIndex(pod.ObjectMeta.Labels["job-name"], "-")
```
mrbarge, line 96: It would be good just for safety to check whether this has returned `-1` or not (meaning the `LastIndex` didn't find anything) and return an error, because otherwise it will cause a panic when used in the subsequent lines.

Tessg22: Isn't content of labels already checked on the line 87? Though I agree that specifically 'strings.LastIndex' is not.

Tessg22: I have added a condition that checks if `LastIndex` returned -1 which creates the safety net.

mrbarge: Thanks a bunch for that!

PR #950, opened 2021-12-17
Pr-Check psudo queue

MrSantamaria:
Implemented config 'pr-check' which will allow a multiple osde2e prcheck jobs to not run on the same cluster

### DIFF
--- a/pkg/e2e/e2e.go
+++ b/pkg/e2e/e2e.go
@@ -158,17 +158,6 @@ func beforeSuite() bool {
 		log.Println("Skipping health checks as requested")
 	}
 
-	//Creates a configmap for PrCheck to use as a psudo queue
-	if viper.GetBool(config.Tests.EnablePrCheck) {
-		if err := cluster.PrCheckQueue(); err != nil {
-			log.Printf("Failed to create queue for PrCheck: %v", err)
-			viper.Set(config.Tests.EnablePrCheck, false)
-			viper.Set(config.Cluster.HibernateAfterUse, false)
-			getLogs()
-			return false
-		}
-	}
-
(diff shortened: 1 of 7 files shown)


pkg/e2e/e2e.go
```diff
@@ -158,6 +159,17 @@ func beforeSuite() bool {
 		log.Println("Skipping health checks as requested")
 	}
 
+	//Creates a configmap for PrCheck to use as a psudo queue
```
jeefy, line 162: I think having a config is useful for testing (and other non-prow-invocations) but before doing the check on line 163 we could also look at the `JOB_NAME` env var and see if it matches `pull-ci-openshift-osde2e-main-verify` -- Prow will always pass the name of the job into the pod it's running. :) 

Then once this is merged we don't need to go and touch another repo for this to take effect.

pkg/e2e/verify/pr_check_queue.go
```diff
@@ -0,0 +1,72 @@
+package verify
```
jeefy, line 1: This should not be in `pkg/e2e` since it's not actually an e2e test.

This would be more appropriate in `pkg/common/`, maybe as part of the `cluster` package?

jeefy (changes requested):
Couple small things but overall looks great! 😁

pkg/e2e/verify/pr_check_queue.go
MrSantamaria: Moved Over in the next commit.

pkg/e2e/e2e.go
MrSantamaria: Closing this change request since we changed the approach to be ocm.Property dependent.

PR #985, opened 2022-01-13
OSD-8474 test cloud-ingress-operator rh-api loadbalancer protection

bng0y:
This test deletes the Loadbalancer associated with rh-api service and ensures that cloud-ingress-operator recreates it

### DIFF
--- a/pkg/e2e/operators/cloudingress/rhapi_lb.go
+++ b/pkg/e2e/operators/cloudingress/rhapi_lb.go
@@ -1,0 +1,99 @@
+package cloudingress
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	viper "github.com/openshift/osde2e/pkg/common/concurrentviper"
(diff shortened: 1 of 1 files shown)


pkg/e2e/operators/cloudingress/rhapi_lb.go
```diff
@@ -0,0 +1,99 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+var _ = ginkgo.Describe(constants.SuiteInforming+TestPrefix, func() {
+	ginkgo.BeforeEach(func() {
+		if viper.GetBool("rosa.STS") {
+			ginkgo.Skip("for now we skip this suite for STS")
+		}
+		if viper.GetString(config.CloudProvider.CloudProviderID) != "aws" {
+			ginkgo.Skip("for now we only support aws provider")
```
fahlmant, line 30: Why can't GCP also be added?

bng0y: this part will be coming in a different PR

PR #992, opened 2022-01-13
OSD-9003 - Verify managed-cluster-config's managed-resources are up-to-date

tnierman:
Completes: https://issues.redhat.com/browse/OSD-9003

---

Adds a test to verify the `managed-resources` ConfigMaps in [managed-cluster-config](https://github.com/openshift/managed-cluster-config/tree/master/deploy/osd-managed-resources) are up-to-date.

### DIFF
--- a/pkg/e2e/verify/namespace_lists.go
+++ b/pkg/e2e/verify/namespace_lists.go
@@ -1,0 +1,261 @@
+package verify
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"time"
+
+	ginkgo "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/openshift/osde2e/pkg/common/alert"
+	"github.com/openshift/osde2e/pkg/common/helper"
+
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+const (
+	namespaceListsTestName        = "[Suite: e2e] Namespace Lists"
+	namespaceListTestPollInterval = 30 * time.Second
+	namespaceListTestPollTimeout  = 20 * time.Minute
+)
+
+func init() {
+	alert.RegisterGinkgoAlert(namespaceListsTestName, "SD-SREP", "Trevor Nierman", "sd-cicd-alerts", "sd-cicd@redhat.com", 4)
+}
+
+var _ = ginkgo.Describe(namespaceListsTestName, func() {
+	ginkgo.Context("are accurate", func() {
+		h := helper.New()
+
+		ginkgo.It("for all managed-object configMaps in the cluster", func() {
+			log.Printf("Generating & retrieving configMaps")
+			testName := fmt.Sprintf("test-namespace-lists-%s-%d-%d", time.Now().Format("20060102-150405"), time.Now().Nanosecond()/1000000, ginkgo.GinkgoParallelProcess())
+			currentConfigMaps, newConfigMaps, err := generateNamespaceConfigMaps(h, testName)
+			Expect(err).ToNot(HaveOccurred(), "Error while generating and retrieving configMaps")
+
+			log.Printf("Comparing configMaps")
+			Expect(len(currentConfigMaps)).To(Equal(len(newConfigMaps)), fmt.Sprintf("Expected to generate same configmaps as configured in managed-cluster-config.\nNumber of generated ConfigMaps: %d\nNumber of configMaps in managed-cluster-config: %d", len(newConfigMaps), len(currentConfigMaps)))
+
+			for name, newList := range newConfigMaps {
+				currentList, present := currentConfigMaps[name]
+				Expect(present).To(BeTrue(), "Expected ConfigMap %s to be present in managed-cluster-config.")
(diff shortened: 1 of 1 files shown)


MrSantamaria:
@tnierman Adding a block on the ci test for now. Let me know when you want this tested to be merged.

pkg/e2e/verify/namespace_lists.go
```diff
@@ -0,0 +1,264 @@
...
+			log.Printf("Comparing configMaps")
+			Expect(len(currentConfigMaps)).To(Equal(len(newConfigMaps)), fmt.Sprintf("Expected to generate same configmaps as configured in managed-cluster-config.\nNumber of generated ConfigMaps: %d\nNumber of configMaps in managed-cluster-config: %d", len(newConfigMaps), len(currentConfigMaps)))
+
+			for name, newList := range newConfigMaps {
+				currentList, present := currentConfigMaps[name]
+				Expect(present).To(BeTrue(), "Expected ConfigMap %s to be present in managed-cluster-config.")
+				Expect(newList).To(Equal(currentList), fmt.Sprintf("ConfigMap '%s' is out of date and needs to be regenerated in managed-cluster-config (https://github.com/openshift/managed-cluster-config/tree/master/deploy/osd-managed-resources#readme) and managed-cluster-validating-webhooks (https://github.com/openshift/managed-cluster-validating-webhooks#updating-namespace-and-service-account-list and https://github.com/openshift/managed-cluster-validating-webhooks#updating-documenation-files).\n\nmanaged-cluster-config contains:\n%s\nShould be:\n%s", name, currentList, newList))
+			}
+
+		}, float64(viper.GetFloat64(config.Tests.PollingTimeout)))
```
mrbarge, line 51: This should maybe be no less than `namespaceListTestPollTimeout` given that's the minimum grace period that the test gives for the job to run.

tnierman: Yeah, agreed; the [current default](https://github.com/openshift/osde2e/blob/main/pkg/common/config/config.go#L627) is 60 (minutes, I assume). This didn't pose an issue while testing locally, but it's possible that the value could be overridden and cause the test to fail unduly. 

An alternative would be to scale it based on `namespaceListTestPollTimeout`: ie- change from `config.Tests.PollingTimeout` to `namespaceListTestPollTimeout * 4`. 

Is this more along the lines of what you were thinking @mrbarge ?

mrbarge: The time unit for the polling timeout is actually seconds, but at the moment it's not actively being enforced.. which is why it hasn't caused any issues (so far I've noticed that with moving to ginkgo v2 in #936, that will eventually change). In the meantime I've been on the lookout for tests where this could pose an issue, which is what made me notice this one. :) 

E2E runs only last for a maximum of four (? I think) hours, so where possible it's good to keep things tight. I agree with your proposal to change it from `config.Tests.PollingTimeout` to something based on the `namespaceListTestPollTimeout`, definitely had similar thoughts. 60 minutes is quite generous though - could the maximum waits for job/configmap deletion be reduced down from their current 15 minutes? I imagine that if a configmap doesn't delete within a minute or two the situation on-cluster is such that it's not going to delete in 15 either.

It's not a huge dealbreaker for me (esp. given the polling timeout is not currently enforced) but I'm just on the lookout for good test hygiene. :)

tnierman: Makes sense, thanks for the info!

I updated the test to timeout after 30 minutes (`namespaceListTestPollTimeout.Seconds()*2`). The `make-all-managed-lists.sh` script can take awhile to run, so I was trying to keep the timeout generous for those particular operations, but if you think that might be a little too long I can certainly tweak things further.

mrbarge:
@tnierman thanks a lot for making that change! Sounds fine to me. 👍

go.mod
```diff
@@ -34,6 +34,7 @@ require (
 	github.com/lib/pq v1.10.4
 	github.com/markbates/pkger v0.17.1
 	github.com/mitchellh/mapstructure v1.4.3
+	github.com/onsi/ginkgo v1.16.4
```
MrSantamaria, line 37: This one needs to be removed. 
This is causing the:

panic: /tmp/go-build1444516780/b1195/state.test flag redefined: ginkgo.seed

pkg/e2e/verify/namespace_lists.go
```diff
@@ -0,0 +1,263 @@
+package verify
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/onsi/ginkgo"
```
MrSantamaria, line 9: "github.com/onsi/ginkgo/v2"

MrSantamaria (changes requested):
Let's start with these two.

pkg/e2e/verify/namespace_lists.go
tnierman: Ah, nice. I think this is what I was missing

MrSantamaria:
The last failure was based on the current timeout. Retesting in case it can pass in that timeframe but if there's any other fail the timeout should be increased.

tnierman:
Prow issues ended up being due to image pull failures, I updated the test to just pull the `latest` images.

Looks like managed-cluster-config is once again out of sync, so we'll need to reconstruct the namespace lists before we can merge this.

tnierman:
/unhold

Looks like other tests also use non-RH images (ie - https://github.com/openshift/osde2e/blob/main/pkg/common/runner/git.go#L9)

@MrSantamaria @mrbarge - any qualms about using a [bitnami image](https://github.com/openshift/osde2e/pull/992/files#diff-a588b180808c671563a1554e71f3ae3e9e013cc555c5c9435ae57a3b3c94a8d6R99) to clone a repo for this test?

tnierman:
/retest

https://github.com/openshift/managed-cluster-config/pull/1067 was merged

mrbarge:
> @MrSantamaria @mrbarge - any qualms about using a [bitnami image](https://github.com/openshift/osde2e/pull/992/files#diff-a588b180808c671563a1554e71f3ae3e9e013cc555c5c9435ae57a3b3c94a8d6R99) to clone a repo for this test?

No issues here personally, and I couldn't find a better choice in quay. Moving away from a `docker.io` dependency seems like a good move for sure.

tnierman:
/retest

Test is passing on other clusters

PR #994, opened 2022-01-18
Removed time.Sleep statements and Annotation check

lnguyen1401:
removes time.Sleeps statements, they are unnecessary. removes annotation[owner]-"cloud-ingress-operator" because the test is redundant. Updated the test to adopt GinkgoV2

### DIFF
--- a/pkg/e2e/operators/cloudingress/certificate.go
+++ b/pkg/e2e/operators/cloudingress/certificate.go
@@ -29,8 +29,6 @@ var _ = ginkgo.Describe(constants.SuiteInforming+TestPrefix, func() {
 			ingress, _ := getingressController(h, "default")
 
 			Expect(string(ingress.Spec.DefaultCertificate.Name)).To(Equal("foo-bar"))
-			Expect(ingress.Generation == int64(1)).To(Equal(false))
-			Expect(ingress.Annotations["Owner"]).To(Equal("cloud-ingress-operator"))
 		}, pollingDuration.Seconds())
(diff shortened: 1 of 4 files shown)


pkg/e2e/operators/cloudingress/certificate.go
```diff
@@ -22,18 +21,14 @@ var _ = ginkgo.Describe(constants.SuiteInforming+TestPrefix, func() {
 			ingress1, _ := getingressController(h, "default")
 			originalCert = string(ingress1.Spec.DefaultCertificate.Name)
 			updateCertificate(h, "foo-bar")
-			time.Sleep(time.Duration(120) * time.Second)
 			ingress, _ := getingressController(h, "default")
-
+			time.Sleep(time.Duration(120) * time.Second)
```
jeefy, line 25: This sleep should also be removed and replaced with a `wait.PollImmediate` block that checks for the condition you're testing or times out.

As it stands, I don't think this sleep does anything useful.

jeefy (changes requested):
Couple changes please :)

pkg/e2e/operators/cloudingress/certificate.go
lnguyen1401: yes you're right! i was trying to remove all the sleep statements and missed this one !

lnguyen1401: removed the sleep statement above.

PR #997, opened 2022-01-19
Passing ocm-token from osde2e secrets for addons to use

MrSantamaria:


pkg/common/load/load.go
```diff
@@ -70,17 +70,24 @@ func Configs(configs []string, customConfig string, secretLocations []string) er
 		for _, folder := range secretLocations {
 			passthruSecrets := viper.GetStringMapString(config.NonOSDe2eSecrets)
 			if !strings.Contains(folder, "osde2e-credentials") && !strings.Contains(folder, "osde2e-common") {
-				continue
+				if viper.Get(config.Addons.IDs) != nil {
+					passthruSecrets["ocm-token-refresh"] = viper.Get("ocm.token").(string)
```
jeefy, line 74: You should be do `viper.GetString("ocm.token")` rather than casting `Get`'s output to a string.

pkg/common/load/load.go
```diff
@@ -70,17 +70,24 @@ func Configs(configs []string, customConfig string, secretLocations []string) er
...
+				} else {
+					continue
+				}
 			}
 
 			err := filepath.Walk(folder, func(path string, info os.FileInfo, err error) error {
 				if err != nil {
 					return fmt.Errorf("Error walking folder %s: %s", folder, err.Error())
 				}
+				fmt.Printf(path)
```
jeefy, line 84: Probably an artifact from debugging, remove

pkg/common/load/load.go
```diff
@@ -70,17 +70,24 @@ func Configs(configs []string, customConfig string, secretLocations []string) er
...
+
 				data, err := ioutil.ReadFile(path)
 				if err != nil {
 					return fmt.Errorf("error loading passthru-secret file %s", path)
 				}
+				fmt.Printf(info.Name())
```
jeefy, line 90: Probably an artifact from debugging, remove

jeefy (changes requested):
Logic looks good, just need to clean up the code a bit :)

MrSantamaria:
Made review changes

PR #998, opened 2022-01-20
feat(local-testing): adds suite

georgettica:
# NOT SURE WHERE TO PUT THAT INTO THE README

this only adds the suite
the way to use is to add 'dummy-suite' to your --config (instead of e2e or such)
and then add the SuiteDummy to the SuiteName


georgettica:
@jeefy how do you see us adding this to the readme?

PR #1008, opened 2022-01-27
Handle workload tests more reliably

mrbarge:
This PR improves the reliability of the `guestbook` workload test and does a general tidy-up of both the `redmine` and `guestbook` workload tests.

I've been observing that the `guestbook` workload test has been failing regularly in E2E lately, after taking a closer look it seems that it makes one attempt to contact the service and if it doesn't respond then the test fails. The test should take into account that the service might take a little while to become ready, though.

So this introduces a minute-long loop of polling the service.

The `redmine` test already had this logic in place, but I noticed it was implemented in a way that could be greatly simplified by just throwing it in a `wait.PollImmediate` block - so I refactored both.

Change has been successfully tested on a staging cluster.


jeefy:
/lgtm
<3 <3

PR #1009, opened 2022-01-27
Make AlertmanagerInhibitions test more robust

mrbarge:
I noticed the AlertmanagerInhibitions test can start to perma-fail if a previous run of the test leaves behind the bad IDP it creates.

This most notably becomes a problem on the PR Check cluster where, if the bad IDP is ever left behind, every subsequent `osde2e` run will fail (as it currently is).

The test will now remove the bad IDP prior to running the test if it already exists.

A good test that this will work is if the PR check for this PR succeeds :)


jeefy:
/lgtm

<3

PR #1045, opened 2022-02-17
rosaprovider: Use new --network-type option

russellb:
The rosaprovider code included the ability to make use of OVNKubernetes
by hacking the install-config.  rosa now accepts a `--network-type`
command line option to do this, so osde2e should make use of it.

Details on the new option can be found in the original PR to rosa:
https://github.com/openshift/rosa/pull/559


russellb:
not actually tested yet ...

MrSantamaria:
Putting a hold since our current verify doesn't test this config.

PR #1053, opened 2022-02-25
Changing des-id to match out vault secret

MrSantamaria:


MrSantamaria:
Force merged this in. We should start seeing the Azure cluster creation working next periodic!

PR #1055, opened 2022-03-01
Add a way to specify the ADDON_PARAMETERS in secret as addon-parameters 

tplevko:
Provides a way to specify addon-parameters in vault - in case the parameters contain secret, so it can't be specified in GH repo.


tplevko:
@MrSantamaria could you please have a look?

PR #1059, opened 2022-03-08
added more logs

MrSantamaria:


pkg/common/load/load.go
```diff
@@ -175,6 +175,8 @@ func loadSecretFileIntoKey(key string, filename string, secretLocations []string
 				viper.Set(key, cleanData)
 			}
 			return nil
+		} else {
+			log.Printf("Could not find secret file %s from location %s", filename, secretLocation)
```
cblecker, line 179: Is it also useful to print err here?

MrSantamaria: It's been a minute since I opened this PR. 
Originally I wanted to to make the logs friendlier for addon users to self debug, but as I looked at the logic in load.go, I decided it's best we update that logic.

PR #1069, opened 2022-03-17
[OHSS-10367] CheckClusterPodHealth was incorrectly passing non-pending pods along

mjlshen:
In #940 a PodPredicate for filtering `Completed` pods was removed in order to implement smarter CronJob pod health checks (returning healthy even previously unhealthy CronJob pods, if the most recent one is healthy). However, there was a [bug](https://github.com/openshift/osde2e/compare/main...mjlshen:image-pruner-bugfix?expand=1#diff-140e0200e3dd96daa38c05e5a3877f19ec4b3b11337d50db24ac917f619b2020L133-R160) which resulted in `Completed` job pods to be forwarded along via https://github.com/openshift/osde2e/blob/main/pkg/common/cluster/clusterutil.go#L454-L465 to `CheckPendingPods`. That function does no further checks on the status of the pod, so it's important to only send pending pods to it.

By breaking out the smarter CronJob pod health check to its own function, CheckJobPods, it can be individually tested. By doing so I found other assumptions, such as pods being already sorted, resulting in bugs that I fixed by sorting the CronJob pods by their creation timestamp.

### DIFF
--- a/pkg/common/cluster/healthchecks/pod_predicates.go
+++ b/pkg/common/cluster/healthchecks/pod_predicates.go
@@ -1,8 +1,9 @@
 package healthchecks
 
 import (
-	kubev1 "k8s.io/api/core/v1"
 	"time"
+
+	kubev1 "k8s.io/api/core/v1"
 )
 
 type PodPredicate func(kubev1.Pod) bool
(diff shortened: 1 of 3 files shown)


mjlshen:
Without this fix, `osd-cluster-ready` reports pod health failures:

```
pod image-pruner-xyz in namespace openshift-image-registry is pending beyond normal threshold: - "
```

Even when the pod's status is `Completed`.

PR #1070, opened 2022-03-17
Allow dependabot to run tests

cblecker:
Always allow dependabot to run CI tests


cblecker:
/override verify
Don't need to test this as it's not related to the osde2e tests themselves

cblecker:
/override ci/prow/verify
Don't need to test this as it's not related to the osde2e tests themselves

PR #1074, opened 2022-03-21
[OSD-10334] Add ocm-agent test suite

a7vicky:


mrbarge:
Tested it and it's working well. Thanks for the addition!
/lgtm

PR #1094, opened 2022-04-11
SDCICD-772 allow ReleaseAccepted as a valid CVO condition

mrbarge:
In recent E2E runs I observed clusters began failing the cluster health check due to the presence of a new `ReleaseAccepted` condition in CVO which was `true`.

This condition is valid to be true and is not an indication of an unhealthy cluster. This PR allows that condition to be in a `true` state without failing the health check.

The original code that did the analysis of CVO conditions seemed quite confusing to me, so I've tried to refactor and comment it a little to make it read a little cleaner as to what it was trying to do. One omission I noticed in the original code is that it seemed to not care about any situation where 'Available' could be 'False', which seems somewhat wrong to me. So the revised logic is:

- if `Available` is false, that's an unhealthy CVO (hopefully there's not an intentional reason this was being ignored)
- if any condition other than `<list of conditions which can be true>` is true, then that's an unhealthy CVO

(I've consciously avoided updating the `go.mod` dependencies to allow parts of this code to be written nicer, ie. referring to the actual `ClusterVersionConditionStatusType` variables rather than raw strings, to reduce the risk involved in pushing this change out ASAP)

Refs [SDCICD-772](https://issues.redhat.com//browse/SDCICD-772)

### DIFF
--- a/pkg/common/cluster/healthchecks/clusterversionoperator.go
+++ b/pkg/common/cluster/healthchecks/clusterversionoperator.go
@@ -33,11 +33,25 @@ func CheckCVOReadiness(configClient configclient.ConfigV1Interface, logger *log.
 	var metadataState []string
 
 	for _, v := range cvInfo.Status.Conditions {
-		if (v.Type != "Available" && v.Status != "False") && v.Type != "Upgradeable" && v.Type != "RetrievedUpdates" {
-			metadataState = append(metadataState, fmt.Sprintf("%v", v))
-			logger.Printf("CVO State not complete: %v: %v %v", v.Type, v.Status, v.Message)
-			success = false
+		switch v.Type {
+		case "Available":
+			// Available must be true
+			if v.Status == "True" {
+				continue
+			}
+		case "Upgradeable", "RetrievedUpdates", "ReleaseAccepted":
+			// These conditions don't matter to readiness state
+			continue
+		default:
+			// Ignore any condition that isn't true
+			// Any other true condition will return an error
+			if v.Status != "True" {
+				continue
+			}
 		}
+		metadataState = append(metadataState, fmt.Sprintf("%v", v))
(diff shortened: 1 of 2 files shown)


mrbarge:
To be honest I do wonder if this whole bit of code could just be simplified by _only_ checking for Available=False, but I worry I am not considering the full breadth of scenarios which warranted the behaviour to be written as it originally was.

pkg/common/cluster/healthchecks/clusterversionoperator.go
```diff
@@ -33,7 +33,28 @@ func CheckCVOReadiness(configClient configclient.ConfigV1Interface, logger *log.
...
+			if !foundAllowedCondition {
+				// A condition is true which is not an accepted-true condition
+				conditionError = true
+			}
+		}
+
+		if conditionError {
 			metadataState = append(metadataState, fmt.Sprintf("%v", v))
 			logger.Printf("CVO State not complete: %v: %v %v", v.Type, v.Status, v.Message)
 			success = false
```
cblecker, line 60:
```suggestion
		switch v.Type {
		case "Available":
			// Available must be true
			if v.Status == "True" {
				continue
			}
		case "Upgradeable", "RetrievedUpdates", "ReleaseAccepted":
			// These conditions don't matter to readiness state
			continue
		default:
			// Ignore any condition that isn't true
			// Any other true condition will return an error
			if v.Status != "True" {
				continue
			}
		}

		metadataState = append(metadataState, fmt.Sprintf("%v", v))
		logger.Printf("CVO State not complete: %v: %v %v", v.Type, v.Status, v.Message)
		success = false
```

mrbarge: Thanks, that switch-around reads a lot better. Applied this change.

PR #1100, opened 2022-04-14
OSD-10981 Handle case when job names are 63 characters long

mjlshen:
`openshift-marketplace` commonly creates jobs that are 63 characters long, like `c50a055663942c74bda6a897a5331c076ec93c668b2fcc4b5a2c5e0acd6326f`, which was not being handled correctly by the pod healthcheck logic, causing osd-cluster-ready to fail.

### DIFF
--- a/ci/prow_pr_check.sh
+++ b/ci/prow_pr_check.sh
@@ -16,7 +16,7 @@ set -eo pipefail
 
     make build
 
-    CLUSTER_ID=1rhqf9e62793ffs2hko7ch8f59mfs2gl \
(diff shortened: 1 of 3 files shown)


MrSantamaria:
thisisalongjobnamethatisawhoppingsixtythreecharacterslongtotest <- I didn't believe it but it is 63 chars long.

MrSantamaria:
I'll keep an eye on that run. The cluster might need to be recreated.

MrSantamaria:
Cleaning the cluster, that's the current failure. :) I'll relaunch shortly.

PR #1116, opened 2022-04-28
[OHSS-10233] Initial FedRAMP configuration file

wshearn:
Ticket: https://issues.redhat.com/browse/OSD-10233

This sets up a config file for manual runs(for now) inside the FedRAMP environment. This is the initial config and we will most likely be expanding it with more tests as time goes on and the fedramp env matures some.

### DIFF
--- a/configs/fedramp.yaml
+++ b/configs/fedramp.yaml
@@ -1,0 +1,11 @@
+tests:
+  enableFips: true
+  testsToRun:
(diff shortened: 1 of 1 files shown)


wshearn:
Looks unrelated, rerunning test
/test verify

wshearn:
Yeah looked at other PR's it seems all PRs are having the same failure

PR #1150, opened 2022-05-10
Pods should be Running or Succeeded

MrSantamaria:


MrSantamaria:
@cblecker We got a pass. 
I'm ready to remove this job if you are. 

I was digging through the logs of why this kept failing and I'll ping you separately on that(Seems like another old job that might be worth looking at.)

PR #1153, opened 2022-05-12
[OSD-11101] add testing for remove proxy

bmeng:
To close [OSD-11101](https://issues.redhat.com//browse/OSD-11101)

Tested locally with following

```
osde2e test --configs proxy-suite --kube-config $KUBECONFIG -e stage --skip-health-check
```

It will run the adding proxy firstly, and then run the removing proxy. Both succeeded.

### DIFF
--- a/pkg/e2e/proxy/postinstall.go
+++ b/pkg/e2e/proxy/postinstall.go
@@ -32,16 +32,16 @@ func init() {
 }
 
 var _ = ginkgo.Describe(postInstallProxyTestName, func() {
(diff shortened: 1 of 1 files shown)


MrSantamaria:
@bmeng few questions. 
I just talked to Matt recently on this effort, and just wondering was this tested with a proxy that was already set up?

bmeng:
Hi @MrSantamaria 
yes, this needs the proxy to be setup before testing.
It likes the add proxy testing, but in reverse.

mrbarge:
Tested this successfully.
/lgtm

PR #1156, opened 2022-05-13
Update PR Check cluster id

mrbarge:
This PR moves us to using the new pr check cluster.


mrbarge:
/close
Will be done in #1150

PR #1163, opened 2022-05-17
[OSD-11524] enable cluster provision with proxy enabled for osd and rosa

bmeng:
Append the proxy related parameters to the ocm and rosa provider.

To test with OSD cluster:
```
export OCM_CCS=true
export OCM_AWS_ACCOUNT=$AWS_ACCOUNT_ID
export OCM_AWS_ACCESS_KEY=$AWS_ACCESS_KEY_ID
export OCM_AWS_SECRET_KEY=$AWS_SECRET_ACCESS_KEY
export CLOUD_PROVIDER_ID=aws
export OCM_AWS_VPC_SUBNET_IDS=subnet-052ad08ae5fd1b9bd,subnet-0b73c8670379ea008
export TEST_HTTP_PROXY=<test http proxy server>
export TEST_HTTPS_PROXY=<test https proxy server>
export USER_CA_BUNDLE=@ca.pem
export CLOUD_PROVIDER_REGION=us-east-1
export OCM_TOKEN=$(ocm token)

./osde2e test --configs stage -e stage --focus-tests "RBAC Operator"
```
The cluster will be created as 
```
$ ocm describe cluster 1s8f9ur5jq5s83bqe7cuigsfbvgbbdon

ID:			1s8f9ur5jq5s83bqe7cuigsfbvgbbdon
External ID:		a6ca850e-6dd2-4512-8ab6-aac96763c338
Name:			osde2e-qtfoz
State:			installing
API URL:		
API Listening:		external
Console URL:		
Masters:		3
Infra:			2
Computes:		2
Product:		osd
Provider:		aws
Version:		
Region:			us-east-1
Multi-az:		false
CCS:			true
Subnet IDs:		[subnet-052ad08ae5fd1b9bd subnet-0b73c8670379ea008]
PrivateLink:		false
STS:			false
Existing VPC:		true
Channel Group:		candidate
Cluster Admin:		true
Organization:		Red Hat OpenShift Online
Creator:		bmeng_sre.openshift
Email:			bmeng@redhat.com
Created:		2022-05-17T06:38:53Z
Expiration:		2022-05-17T12:38:51Z
Shard:			-------
HTTPProxy:	        http://<proxy_server_ip>:3128
HTTPSProxy:	        http://<proxy_server_ip>:3128
AdditionalTrustBundle:  REDACTED

```

To test with ROSA cluster:
```
export PROVIDER=rosa
export ROSA_AWS_ACCESS_KEY_ID=$AWS_ACCOUNT_ID
export ROSA_AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY
export ROSA_AWS_REGION=us-east-1
export ROSA_SUBNET_IDS=$OCM_AWS_VPC_SUBNET_IDS 
export ROSA_ENV=stage
export TEST_HTTP_PROXY=<test http proxy server>
export TEST_HTTPS_PROXY=<test https proxy server>
export USER_CA_BUNDLE=ca.pem               <-----------------------------------  direct file path here

./osde2e test --configs stage -e stage --focus-tests "RBAC Operator"
```
The rosa cluster will be created as:
```
I: Creating cluster 'osde2e-6pi5h'
I: To view a list of clusters and their status, run 'rosa list clusters'
I: Cluster 'osde2e-6pi5h' has been created.
I: Once the cluster is installed you will need to add an Identity Provider before you can login into the cluster. See 'rosa create idp --help' for more information.
Name:                       osde2e-6pi5h
ID:                         1s8gi65jsijjrt0r0ds51mro6ruro5v7
External ID:                
OpenShift Version:          
Channel Group:              candidate
DNS:                        REDACTED
AWS Account:                537136174241
API URL:                    
Console URL:                
Region:                     us-east-1
Multi-AZ:                   false
Nodes:
 - Control plane:           3
 - Infra:                   2
 - Compute:                 2
Network:
 - Service CIDR:            172.30.0.0/16
 - Machine CIDR:            10.0.0.0/16
 - Pod CIDR:                10.128.0.0/14
 - Host Prefix:             /23
Proxy:
 - HTTPProxy:               http://<proxy_server_ip>:3128
 - HTTPSProxy:              http://<proxy_server_ip>:3128
Additional trust bundle:    REDACTED
State:                      validating 
Private:                    No
Created:                    May 17 2022 08:04:41 UTC
Details Page:              REDACTED

```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/PagerDuty/go-pagerduty v1.5.1
 	github.com/PuerkitoBio/goquery v1.8.0
 	github.com/adamliesko/retry v0.0.0-20200123222335-86c8baac277d
+	github.com/antlr/antlr4/runtime/Go/antlr v0.0.0-20220209173558-ad29539cd2e9
 	github.com/aws/aws-sdk-go v1.44.14
 	github.com/bmizerany/perks v0.0.0-20141205001514-d9a9656a3a4b // indirect
 	github.com/code-ready/crc v1.10.0
--- a/pkg/common/providers/rosaprovider/cluster.go
+++ b/pkg/common/providers/rosaprovider/cluster.go
@@ -68,6 +68,7 @@ func (m *ROSAProvider) IsValidClusterName(clusterName string) (bool, error) {
 }
 
 // LaunchCluster will provision an AWS cluster.
+// nolint:gocyclo
 func (m *ROSAProvider) LaunchCluster(clusterName string) (string, error) {
 	// Calculate an expiration date for the cluster so that it will be automatically deleted if
 	// we happen to forget to do it:
@@ -120,6 +121,19 @@ func (m *ROSAProvider) LaunchCluster(clusterName string) (string, error) {
 		"--service-cidr", viper.GetString(ServiceCIDR),
 		"--pod-cidr", viper.GetString(PodCIDR),
 		"--host-prefix", viper.GetString(HostPrefix),
+	}
+	if viper.GetString(SubnetIDs) != "" {
+		subnetIDs := viper.GetString(SubnetIDs)
+		createClusterArgs = append(createClusterArgs, "--subnet-ids", subnetIDs)
+		if httpProxy := viper.GetString(config.Proxy.HttpProxy); httpProxy != "" {
+			createClusterArgs = append(createClusterArgs, "--http-proxy", httpProxy)
+		}
+		if httpsProxy := viper.GetString(config.Proxy.HttpsProxy); httpsProxy != "" {
+			createClusterArgs = append(createClusterArgs, "--https-proxy", httpsProxy)
+		}
+		if userCaBundle := viper.GetString(config.Proxy.UserCABundle); userCaBundle != "" {
+			createClusterArgs = append(createClusterArgs, "--additional-trust-bundle-file", userCaBundle)
+		}
 	}
 	if viper.GetBool(config.Cluster.MultiAZ) {
 		createClusterArgs = append(createClusterArgs, "--multi-az")
--- a/pkg/e2e/osd/dedicatedadmin.go
+++ b/pkg/e2e/osd/dedicatedadmin.go
@@ -327,7 +327,7 @@ func manageSecrets(nsList []string, h *helper.H) error {
 
 		// check 'create' permission
 		secrets := h.Kube().CoreV1().Secrets(ns)
-		dummySecret, err := secrets.Create(context.TODO(), &corev1.Secret{
+		_, err := secrets.Create(context.TODO(), &corev1.Secret{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      newSecretName,
 				Namespace: ns,
@@ -336,7 +336,7 @@ func manageSecrets(nsList []string, h *helper.H) error {
 		Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("failed to create secret %s in namespace %s", newSecretName, ns))
 
 		// check 'get' permission
-		dummySecret, err = secrets.Get(context.TODO(), newSecretName, metav1.GetOptions{})
+		dummySecret, err := secrets.Get(context.TODO(), newSecretName, metav1.GetOptions{})
 		Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("failed to get secret %s in namespace %s", newSecretName, ns))
 
 		// check 'list' permission
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -35,9 +35,13 @@ Some environment variables commonly used for pipelines under osde2e are indicate
  
 | Environment variable | Usage |
 | --------------| ------------------------| 
-|ROSA_AWS_ACCESS_KEY_ID| AWSAccessKeyID for provisioning clusters|
+|ROSA_ENV| Environment for the e2e testing, default to prod.|
+|ROSA_AWS_ACCESS_KEY_ID| AWSAccessKeyID for provisioning clusters.|
 |ROSA_AWS_SECRET_ACCESS_KEY| AWSSecretAccessKey for provisioning clusters.|
 |ROSA_AWS_REGION| AWSRegion for provisioning clusters.|
+|ROSA_STS| Boolean value to indicate the cluster is STS enabled or not.|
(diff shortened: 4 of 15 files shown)


pkg/e2e/osd/dedicatedadmin.go
```diff
@@ -336,7 +336,7 @@ func manageSecrets(nsList []string, h *helper.H) error {
 		Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("failed to create secret %s in namespace %s", newSecretName, ns))
 
 		// check 'get' permission
-		dummySecret, err = secrets.Get(context.TODO(), newSecretName, metav1.GetOptions{})
+		dummySecret, err := secrets.Get(context.TODO(), newSecretName, metav1.GetOptions{})
```
T0MASD, line 339: as in line 303

```suggestion
		_, err := secrets.Get(context.TODO(), newSecretName, metav1.GetOptions{})
```

go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/PagerDuty/go-pagerduty v1.5.1
 	github.com/PuerkitoBio/goquery v1.8.0
 	github.com/adamliesko/retry v0.0.0-20200123222335-86c8baac277d
+	github.com/antlr/antlr4/runtime/Go/antlr v0.0.0-20220209173558-ad29539cd2e9
```
T0MASD, line 11: was this dep added by accident? I don't see it being imported in this PR

T0MASD (commented):
overall lgtm couple of minor points

pkg/e2e/osd/dedicatedadmin.go
bmeng: So basically, this is not related to my change at all.
Just the `make` failed locally due to the golang lint failed.
The reason for line 330 is that the `dummySecret` declared and assigned but not being used at all. Or the value has been overridden by this line.

But the value of `dummySecret` is required since it is being used in several lines below. Line 347

go.mod
bmeng: I have no idea, I just run the `go mod tidy` and this should be updated automatically.

pkg/common/providers/rosaprovider/cluster.go
```diff
@@ -121,6 +122,19 @@ func (m *ROSAProvider) LaunchCluster(clusterName string) (string, error) {
...
+		subnetIDs := viper.GetString(SubnetIDs)
+		createClusterArgs = append(createClusterArgs, "--subnet-ids", subnetIDs)
+		if httpProxy := viper.GetString(config.Proxy.HttpProxy); httpProxy != "" {
+			createClusterArgs = append(createClusterArgs, "--http-proxy", httpProxy)
+		}
+		if httpsProxy := viper.GetString(config.Proxy.HttpsProxy); httpsProxy != "" {
+			createClusterArgs = append(createClusterArgs, "--https-proxy", httpsProxy)
+		}
+		if userCaBundle := viper.GetString(config.Proxy.UserCABundle); userCaBundle != "" {
+			createClusterArgs = append(createClusterArgs, "--additional-trust-bundle-file", userCaBundle)
```
mrbarge, line 135: There's a slight issue here we need to figure out.

The value of `userCABundle` here is derived from the `USER_CA_BUNDLE` config environment variable.

Currently, we have two ways of handling that value:
- it either represents a filepath preceded by an @, in which case the contents are loaded into memory.
- or it is the actual text representation of the CA bundle.

The problem that's happening is that this code demands that `USER_CA_BUNDLE` be a filepath, but not one prefixed by an `@` to load the contents into memory.

Which is fine, except that it then breaks the other dat-2 proxy tests we just introduced, as they would attempt to set a user CA bundle on clusters which is a value that's just a filepath, rather than the contents of the file.

So we need to align our strategy here so that one doesn't break the other.

- Only ever treat `USER_CA_BUNDLE` as a filepath (no @ prefix). In which case the day-2 tests need to be fixed to load the contents of this path rather than just take the value in `USER_CA_BUNDLE`.
- Have `USER_CA_BUNDLE` be either a filepath (with @ prefix) or the actual CA value (like it currently behaves). In which case this code needs to be fixed to write the contents of `config.Proxy.UserCABundle` to a temporary file which is passed to the calling arguments here.
- Something else..?

bmeng: Ok, so the current problem:
If we set the USER_CA_BUNLDE=@<path_ca_file>
It works in ocmprovider during installation, since the `AdditionalTrustBundle()` func in the `ClusterBuilder` accepts the value of the file only.
But it does not in rosaprovider during installation, since it appends the value directly to the cmd. `rosa create cluster --additional-trust-bundle-file`

If we set the USER_CA_BUNDLE=<path_ca_file> without `@`
It works in rosaprovider during installation, but after that, the `AddClusterProxy()`, `RemoveClusterProxy()` and `RemoveUserCABundle()` in package ocmprovider will be failed, since the USER_CA_BUNDLE will be treated as contents. And rosaprovider is referring to the same functions in ocmproviders above.

If we set the USER_CA_BUNDLE=<ca_contents>
If will not work for rosaprovider as we do not have a way to inject the contents of the ca certs to `rosa create` command now.

My thoughts:
1. We move the `LoadUserCABundle()` func out side the config.go, so we can process the `USER_CA_BUNDLE=@<file>` for ocmprovider during installation time only. (Seems the same approach with your 1st point?)
2. Add a new environment like `USER_CA_BUNDLE_FILE` which used for rosa installation. We can also generate the contents value of user ca bundle based on this, but seems increasing the complexity.
3. Write contents to temporary file is not a good idea IMO...

Personally, I think we always treat the value of `USER_CA_BUNDLE` as a filepath, and abstract the contents when needed is better approach. Unless we have a use case that we need to pass the environment as the contents value of the CA.

bmeng:
Updated based on the discussion above.

Now the environment will accept the file path name only.
The ocmprovider will extract the contents of the ca file and feed them to the `clusterBuilder.AdditionalTrustBundle()`.

Verified with the following scenarios:
- Install cluster with proxy configured for OSD cluster
- Install cluster with proxy configured for ROSA cluster
- Adding proxy testing for OSD cluster

pkg/common/providers/rosaprovider/cluster.go
mrbarge: Thanks for doing this change, I tested it and it is working well.

PR #1179, opened 2022-05-25
[WIP] SDCICD-773 BYO-VPC Feature

MrSantamaria:
To do: 

1. Viper Variables to be added. 
2. Figure out how the --privateLink flag works on ROSA (https://github.com/openshift/rosa/cmd/create/cluster/cmd.go#L1175)
3. Catch up with the ProxyPR https://github.com/openshift/osde2e/pull/1163


mrbarge:
So far so good. The VPC name doesn't have to align with the cluster name - it can be named anything.

MrSantamaria:
> So far so good. The VPC name doesn't have to align with the cluster name - it can be named anything.

My thoughts here is for cleanup purposes. If for any reason a remnant vpc is left over it could be matched with the cluster that was testing the flow?

MrSantamaria:
Next:
Behavior on already existing VPC?
VPC Cleanup on Test Completion
Test passing already existing VPC.

docs/Config.md
```diff
@@ -58,12 +47,23 @@ Some environment variables commonly used for pipelines under osde2e are indicate
...
+### AWS Related:-
+
+| Environment variable | Usage |
+| --------------| ------------------------| 
+|AWS_ACCOUNT | AWS Account ID|
+|AWS_ACCESS_KEY| AWSAccessKey for provisioning clusters|
+|AWS_SECRET_ACCESS_KEY| AWSSecretAccessKey for provisioning clusters.|
+|AWS_REGION| AWSRegion for provisioning clusters.|  
+|BYO_VPC| Define existing VPC via a vpc id|
+|SUBNET_IDS| A list of subnets used to create the cluster, required for proxy enabled cluster.|
```
ravitri, line 65: Is there a criteria if these need to be private or public subnets?

ravitri: As discussed, for proxy enabled cluster, it would require a private subnet for now. We could mention it explicitly too such that users don't provide public subnet when running osde2e manually.

mrbarge: This parameter is used by OSDE2E for installing a BYOVPC cluster into a pre-existing VPC. Both the public and private subnets are required here for a successful install to take place. The value of this parameter is supplied to the rosa/osd installer.

ravitri: Ah yes.. Understood.. The reference for the proxy's private subnet will be separate then.. This is cluster creation so both required.. Yup..

MrSantamaria:
@mrbarge has there been any progress in automating the squid proxy automation part of this PR? 
I'd like to attach where we can track this PR so the full flow can be completed.

jbpratt:
this can be closed, right? @MrSantamaria

PR #1234, opened 2022-07-12
Adds tests for Managed Node Metadata Operator

iamkirkbater:
This PR adds tests for Managed Node Metadata Operator.

This satisfies [OSD-10051](https://issues.redhat.com//browse/OSD-10051)

### DIFF
--- a/pkg/e2e/operators/mnmo.go
+++ b/pkg/e2e/operators/mnmo.go
@@ -1,0 +1,597 @@
+package operators
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	ocmTypes "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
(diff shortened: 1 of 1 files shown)


MrSantamaria:
Let me recreate the PR cluster that's shot right now. I'll retest this once it fails again.

macgregor:
I know you had to get a little clever with how you tested this since these clusters arent hive provisioned, some of the code comments hint at that, but could you add a block comment in a prevalent place giving some context to why testing it was tricky and how thats addressed? 

Otherwise, /lgtm

iamkirkbater:
The biggest problem that I had with testing it was that I had to use `client-go` and 1) I've never used that before and 2) it required re-building all of the tests that we already had defined in the operator itself.

With that said, the biggest problem was the MachinePool creation through OCM and only running that once (which will always make the first test take at least a few minutes and report as a slow test), but using ordered containers gives us that "BeforeAll" capability to run that in isolation.

MrSantamaria:
Let me recreate this cluster again after this test fails. The failures I'm seeing now can be caused by the cluster hibernating/waking up.

PR #1235, opened 2022-07-14
adding skips for focus test

ritmun:
added one liner to run quick focus test files


MrSantamaria:
That's a smart way to do the focus test in this environment. 

Looking at that error I've never seen something like that before.

ritmun:
> That's a smart way to do the focus test in this environment.
> 
> Looking at that error I've never seen something like that before.

Said "something is wrong with MNMO or there is a network issue with OCM".. given the update was 2 lines in README, safe to say the failing mnmo test it was unrelated.

PR #1289, opened 2022-08-22
Change default networkType to OVNKubernetes

a7vicky:


pkg/common/providers/ocmprovider/cluster.go
```diff
@@ -1415,7 +1415,7 @@ func (o *OCMProvider) GetNetworkConfig(networkProvider string) string {
 	if networkProvider == config.DefaultNetworkProvider {
 		return ""
 	}
-	if networkProvider != "OVNKubernetes" {
+	if networkProvider != "OpenShiftSDN" {
```
mrbarge, line 1418: Why the change to SDN here? (and on line 1429)

a7vicky: I am under impression that if we reverse the logic it will work to test cluster with **OpenShiftSDN** networkType if we needed. CLUSTER_NETWORK_PROVIDER="OpenShiftSDN" env will repect for stage during Launch cluster. Not sure if this installconfig is used for creating cluster[here](https://github.com/openshift/osde2e/blob/853ba77482cde426a0ffe1897f07e9562a264aa9/pkg/common/providers/ocmprovider/cluster.go#L112)

mrbarge: I would be more in favour of just deleting that whole function and whatever calls it, as it seems at first glance like a hack to satisfy an edge case that is no longer necessary, and at best creates weird confusion.

Let's see what @MrSantamaria thinks.

MrSantamaria: Oops I typed a reply and never submitted it. 

I know the jump to ovn and deprecating the openshiftSDN was an upcoming plan but are we fully OVN across all the cloud providers and FedRamp? 
The current pipelines I'm keeping track of have not been failing and this change would void all OpenshiftSDN. 

Once OpenshiftSDN is fully deprecated, I agree with Matt, we can deprecate the selection of a network provider. 
@russellb Do you have any input here as well?

mrbarge: As of right now, provided we're testing 4.11 clusters (which all of our pipelines should be), it'll be OVN everywhere for managed OpenShift clusters. I'm not 100% on Fedramp but I'll check on that.

a7vicky: Thanks @mrbarge , I have removed the fuction, build binary osde2e and tested by creating osd cluster and successfully created OVN enabled cluster.

 _HIBERNATE_AFTER_USE="false" CLUSTER_NETWORK_PROVIDER="OpenShiftSDN" OCM_TOKEN=$(ocm token) CLOUD_PROVIDER_ID="aws" PROVIDER="ocm" CLUSTER_VERSION="openshift-v4.11.1-candidate" out/osde2e test -e stage --configs e2e-suite_
 
 → ocm get cluster 1uahg6j6ctht0kar9fp39ccedlrqgh03 | jq .network
```
{
  "type": "OVNKubernetes",
  "machine_cidr": "10.0.0.0/16",
  "service_cidr": "172.30.0.0/16",
  "pod_cidr": "10.128.0.0/14",
  "host_prefix": 23
}
```

pkg/common/config/config.go
```diff
@@ -77,7 +77,7 @@ const (
 	CanaryChance = "canaryChance"
 
 	// Default network provider for OSD
-	DefaultNetworkProvider = "OpenShiftSDN"
+	DefaultNetworkProvider = "OVNKubernetes"
```
bmeng, line 80: This const is used in `pkg/common/providers/rosaprovider/cluster.go` only?
I think we can remove that part as well.

mrbarge:
I think where we've landed on this is that, with the removal of the OVN periodic jobs and the fact that OCM now fully supports OVN, the need to hack it in to the install config is no longer needed.

/lgtm

PR #1314, opened 2022-09-30
Putting-hibernation-behind-flags

MrSantamaria:
Moving Hibernation to be available behind flags.
Clusters will now delete post e2e suite being ran.

### DIFF
--- a/pkg/common/config/config.go
+++ b/pkg/common/config/config.go
@@ -682,7 +682,7 @@ func InitViper() {
 	viper.SetDefault(Cluster.Channel, "candidate")
 	viper.BindEnv(Cluster.Channel, "CHANNEL")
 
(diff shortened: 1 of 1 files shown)


MrSantamaria:
/hold
Let's not merge this till Monday...This will change OSDe2e behavior across the board.

MrSantamaria:
Ohhh cool, one of these flags is causing the normal OSD flow to enter the addons flow.

MrSantamaria:
Set the cluster to hibernate after the run. 
The cluster will still get deleted after 6 hours by OCM due to our expire flag. 

Also raised a bug for the DestroyAfterTest flag not working correctly.

The change here UseExistingCluster, will disable OSDe2e from native looking for clusters to be reused and instead create a new cluster every instance.

PR #1346, opened 2022-10-28
SDCICD-821 - DestroyAfterTest flag Debug

MrSantamaria:
The current use of the DestroyAfterTest flag is causing the pipeline to show as failed since it's not exiting gracefully. 
We need to be able to destroy clusters after runs in order to clean up STS resources during the same run as the delete cluster functions won't clean up the Cloud accounts.

### DIFF
--- a/pkg/common/cluster/clusterutil.go
+++ b/pkg/common/cluster/clusterutil.go
@@ -359,7 +359,8 @@ func waitForClusterReadyWithOverrideAndExpectedNumberOfNodes(clusterID string, l
 // ClusterConfig returns the rest API config for a given cluster as well as the provider it
 // inferred to discover the config.
 // param clusterID: If specified, Provider will be discovered through OCM. If the empty string,
-// 		assume we are running in a cluster and use in-cluster REST config instead.
+//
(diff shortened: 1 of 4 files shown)


MrSantamaria:
Troubleshooting on my end, the clusters are deleting but there's some steps OSDe2e uses after we are calling for the cluster to be destroyed that are causing the suite to throw a fail.

MrSantamaria:
Cool stuff. 
Debugged some more the clusterID gets lost at some point when OSDe2e creates a cluster from scratch.

MrSantamaria:
The logs ended at the cluster switching over to uninstall but it didn't wait for the Rosa cluster be fully uninstalled to delete the roles/oidc

MrSantamaria:
2022/10/31 22:47:35 cluster.go:304: [operator-roles --cluster xxxx --mode auto --yes]
2022/10/31 22:48:10 cluster.go:309: [oidc-provider --cluster xxxx --mode auto --yes]
? Operator roles deletion mode:  [Use arrows to move, type to filter, ? for more help]
> auto
  manual
[?25l

The delete for the operatorRoles is not working on auto mode.

PR #1358, opened 2022-11-01
SDCICD-760: remove unneeded sts account setup

jbpratt:
It is no longer needed to create specific roles which will help with the cluttering of the AWS account

/hold


jbpratt:
requires https://github.com/openshift/rosa/issues/914 or another fix

PR #1363, opened 2022-11-04
fix core/v1 api imported twice

jbpratt:
it is aliased to corev1


MrSantamaria:
Good catch thanks for cleaning it up! 
I'm on the mind that we shouldn't use aliases throughout implementations, maybe we should start keeping track of all the instances they are used. 
/lgtm

PR #345, opened 2020-04-18
SDCICD-264. Addon improvements and result download improvements.

meowfaceman:
All jobs will now try to capture the output of the runner containers as
well as flesh out the output of addons in order to help debug issues
that we get with addon test harness running.

Additionally, results will now download recursively rather than just the
top level directory. This will allow us to capture more structured
information on a per test basis.

Finally, the addon tests have been fixed after they were broken by the
recent change to add in a 5 character long random suffix to pod names.
We now detect pod names as opposed to depending on fixed pod names in
the addon test harness job.

PR #1372, opened 2022-11-07
fix common/runner importing random log package

jbpratt:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/bmizerany/perks v0.0.0-20141205001514-d9a9656a3a4b // indirect
 	github.com/cenkalti/backoff v2.2.1+incompatible // indirect
 	github.com/code-ready/crc v1.10.0
-	github.com/emicklei/go-restful v2.16.0+incompatible
+	github.com/emicklei/go-restful v2.16.0+incompatible // indirect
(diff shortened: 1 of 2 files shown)


MrSantamaria:
@jbpratt could you give me more of a description here? 
I'm not sure why go-restful was pulled for logs.

MrSantamaria:
Initial PR:
https://github.com/openshift/osde2e/pull/345

I'm not fully sure what this had to do with the addon suite but I'm not sure we need to make this swap right now.

jbpratt:
> @jbpratt could you give me more of a description here? I'm not sure why go-restful was pulled for logs.

it looks like it was just a mistake from goimports or something when the initial PR was made https://github.com/openshift/osde2e/blob/2de8798c7cb98c303eee60fb02815b4343ab1916/pkg/common/runner/service.go#L104

jbpratt:
> Initial PR: #345
> 
> I'm not fully sure what this had to do with the addon suite but I'm not sure we need to make this swap right now.

updated the commit message with more details, it is just a simple mistake of importing the wrong package, there is no functionality that package is providing that we need directly

MrSantamaria:
Huh that is interesting. Looking at my local branch it got pulled into the cloudingress operator test. 
Can't say I see a conflict though, this does seem random.

PR #1373, opened 2022-11-07
Adds test for prometheusrule webhook in managed namespaces

reetika-vyas:
This PR is for prometheusrule webhook validation in managed namespaces. 

It will close [OSD-13302](https://issues.redhat.com/browse/OSD-13302)

### DIFF
--- a/pkg/e2e/verify/namespace_webhook.go
+++ b/pkg/e2e/verify/namespace_webhook.go
@@ -12,12 +12,15 @@ import (
 	"github.com/openshift/osde2e/pkg/common/alert"
 	"github.com/openshift/osde2e/pkg/common/config"
 	"github.com/openshift/osde2e/pkg/common/util"
+	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/rest"
 
 	viper "github.com/openshift/osde2e/pkg/common/concurrentviper"
 	"github.com/openshift/osde2e/pkg/common/helper"
+	"k8s.io/apimachinery/pkg/api/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -34,6 +37,8 @@ var _ = ginkgo.Describe(namespaceWebhookTestName, func() {
 		DUMMY_GROUP = "random-group-name"
 		// User to use for impersonation
 		DUMMY_USER = "testuser@testdomain"
+
+		prometheusName = "prometheus-example-app"
 	)
 
 	PRIVILEGED_USERS := []string{
@@ -101,6 +106,18 @@ var _ = ginkgo.Describe(namespaceWebhookTestName, func() {
 				Expect(err).NotTo(HaveOccurred())
 			}
 
+			for privilegedNamespace, manageNamespace := range PRIVILEGED_NAMESPACES {
+				if manageNamespace {
+					h.Impersonate(rest.ImpersonationConfig{})
+					err := deletePrometheusRule(privilegedNamespace, prometheusName, h)
+					Expect(err).NotTo(HaveOccurred())
+				}
+			}
+			for _, namespace := range NONPRIV_NAMESPACES {
+				err := deletePrometheusRule(namespace, prometheusName, h)
+				Expect(err).NotTo(HaveOccurred())
+			}
+
 			// Wait until all namespaces have verified to be deleted
 			namespacesToCheck := make([]string, 0)
 			for ns, managed := range PRIVILEGED_NAMESPACES {
@@ -154,6 +171,40 @@ var _ = ginkgo.Describe(namespaceWebhookTestName, func() {
 			// Non-privileged users can manage all non-privileged namespaces
 			for _, nonPrivilegedNamespace := range NONPRIV_NAMESPACES {
 				err := updateNamespace(ctx, nonPrivilegedNamespace, DUMMY_USER, "dedicated-admins", h)
+				Expect(err).NotTo(HaveOccurred())
+			}
+		}, viper.GetFloat64(config.Tests.PollingTimeout))
+
+		util.GinkgoIt("dedicated admin can not create PrometheusRule", func(ctx context.Context) {
+			for privilegedNamespace := range PRIVILEGED_NAMESPACES {
+				err := e2eCreatePrometheusRule(ctx, prometheusName, privilegedNamespace, DUMMY_USER, "dedicated-admins", h)
+				Expect(apierrors.IsForbidden(err)).To(BeTrue())
+			}
+		}, viper.GetFloat64(config.Tests.PollingTimeout))
+
+		util.GinkgoIt("Non-privileged users cannot create prometheusrule in privileged namespaces", func(ctx context.Context) {
+			for privilegedNamespace := range PRIVILEGED_NAMESPACES {
+				err := e2eCreatePrometheusRule(ctx, prometheusName, privilegedNamespace, DUMMY_USER, DUMMY_GROUP, h)
+				Expect(errors.IsForbidden(err)).To(Equal(true))
+			}
+		}, viper.GetFloat64(config.Tests.PollingTimeout))
+
+		util.GinkgoIt("Privileged users can create prometheusrule in all namespaces", func(ctx context.Context) {
+			for _, privilegedUser := range PRIVILEGED_USERS {
+				for privilegedNamespace := range PRIVILEGED_NAMESPACES {
+					err := e2eCreatePrometheusRule(ctx, prometheusName, privilegedNamespace, privilegedUser, "", h)
+					Expect(err).NotTo(HaveOccurred())
+				}
+				for _, namespace := range NONPRIV_NAMESPACES {
+					err := e2eCreatePrometheusRule(ctx, prometheusName, namespace, privilegedUser, "", h)
+					Expect(err).NotTo(HaveOccurred())
+				}
+			}
+		}, viper.GetFloat64(config.Tests.PollingTimeout))
+
+		util.GinkgoIt("Non-privileged users can create prometheusrule all non-privileged namespaces", func(ctx context.Context) {
+			for _, nonPrivilegedNamespace := range NONPRIV_NAMESPACES {
+				err := e2eCreatePrometheusRule(ctx, prometheusName, nonPrivilegedNamespace, DUMMY_USER, "dedicated-admins", h)
 				Expect(err).NotTo(HaveOccurred())
 			}
 		}, viper.GetFloat64(config.Tests.PollingTimeout))
@@ -254,6 +305,15 @@ func deleteNamespace(ctx context.Context, namespace string, waitForDelete bool, 
 	return err
 }
 
+func deletePrometheusRule(namespace, prometheusName string, h *helper.H) error {
+	log.Printf("Deleting prometheusrule for prometheusrule validation webhook (%s)", prometheusName)
+	err := h.Prometheusop().MonitoringV1().PrometheusRules(namespace).Delete(context.TODO(), prometheusName, metav1.DeleteOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		return fmt.Errorf("failed to delete prometheusrule from managed namespace '%s': %v", prometheusName, err)
+	}
+	return nil
+}
+
 func createNamespace(ctx context.Context, namespace string, h *helper.H) (*v1.Namespace, error) {
 	// If the namespace already exists, we don't need to create it. Just return.
 	ns, err := h.Kube().CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
@@ -285,4 +345,49 @@ func createNamespace(ctx context.Context, namespace string, h *helper.H) (*v1.Na
 	})
 
 	return ns, err
+}
+
+func e2eCreatePrometheusRule(ctx context.Context, prometheusrulename, namespace string, asUser string, userGroup string, h *helper.H) error {
+	// reset impersonation upon return
+	defer h.Impersonate(rest.ImpersonationConfig{})
+
+	// reset impersonation at the beginning just-in-case
+	h.Impersonate(rest.ImpersonationConfig{})
+
+	// we need to add these groups for impersonation to work
+	userGroups := []string{"system:authenticated", "system:authenticated:oauth"}
+	if userGroup != "" {
+		userGroups = append(userGroups, userGroup)
+	}
+
+	// update the namespace as our desired user
+	h.Impersonate(rest.ImpersonationConfig{
+		UserName: asUser,
+		Groups:   userGroups,
+	})
+
+	e2eprometheusrule := &monv1.PrometheusRule{
+		TypeMeta: metav1.TypeMeta{},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      prometheusrulename,
+			Namespace: namespace,
+		},
+		Spec: monv1.PrometheusRuleSpec{
+			Groups: []monv1.RuleGroup{
+				{
+					Name: "example",
+					Rules: []monv1.Rule{
+						{
+							Alert: "VersionAlert",
+							Expr: intstr.IntOrString{
+								StrVal: "version{job==\"prometheus-example-app\"} == 0",
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+	_, err := h.Prometheusop().MonitoringV1().PrometheusRules(namespace).Create(ctx, e2eprometheusrule, metav1.CreateOptions{})
+	return err
 }


pkg/e2e/verify/namespace_webhook.go
```diff
@@ -159,6 +163,80 @@ var _ = ginkgo.Describe(namespaceWebhookTestName, func() {
...
+				prome2etestrule := e2eCreatePrometheusRule(prometheusName, nonPrivilegedNamespace, DUMMY_USER, "dedicated-admins", h)
+				err := wait.PollImmediate(5*time.Second, 3*time.Minute, func() (bool, error) {
+					_, err := h.Prometheusop().MonitoringV1().PrometheusRules(nonPrivilegedNamespace).Create(context.TODO(), prome2etestrule, metav1.CreateOptions{})
+					if err != nil {
+						log.Printf("failed creating prometheus rules: %v", err)
+						return false, nil
+					}
+					return true, nil
+				})
+				Expect(err).To(HaveOccurred())
```
jbpratt, line 237: This block can go in `e2eCreatePrometheusRule` to de-duplicate all of the tests doing this just on a different namespace (no need to return the prometheusrule)

pkg/e2e/verify/namespace_webhook.go
```diff
@@ -159,6 +163,80 @@ var _ = ginkgo.Describe(namespaceWebhookTestName, func() {
...
+					prome2etestrule := e2eCreatePrometheusRule(prometheusName, privilegedNamespace, privilegedUser, "", h)
+					err := wait.PollImmediate(5*time.Second, 3*time.Minute, func() (bool, error) {
+						_, err := h.Prometheusop().MonitoringV1().PrometheusRules(privilegedNamespace).Create(context.TODO(), prome2etestrule, metav1.CreateOptions{})
+						if err != nil {
+							log.Printf("failed creating prometheus rules: %v", err)
+							return false, nil
+						}
+						return true, nil
+					})
+					Expect(err).To(HaveOccurred())
```
mrbarge, line 209: Are we really expecting an error to have occurred here?

pkg/e2e/verify/namespace_webhook.go
```diff
@@ -159,6 +163,80 @@ var _ = ginkgo.Describe(namespaceWebhookTestName, func() {
...
+					prome2etestrule := e2eCreatePrometheusRule(prometheusName, namespace, privilegedUser, "", h)
+					err := wait.PollImmediate(5*time.Second, 3*time.Minute, func() (bool, error) {
+						_, err := h.Prometheusop().MonitoringV1().PrometheusRules(namespace).Create(context.TODO(), prome2etestrule, metav1.CreateOptions{})
+						if err != nil {
+							log.Printf("failed creating prometheus rules: %v", err)
+							return false, nil
+						}
+						return true, nil
+					})
+					Expect(err).To(HaveOccurred())
```
mrbarge, line 221: Likewise here, we'd expect these to both be not erroring, right?

pkg/e2e/verify/namespace_webhook.go
```diff
@@ -159,6 +163,80 @@ var _ = ginkgo.Describe(namespaceWebhookTestName, func() {
...
+				}
+				for _, namespace := range NONPRIV_NAMESPACES {
+					prome2etestrule := e2eCreatePrometheusRule(prometheusName, namespace, privilegedUser, "", h)
+					err := wait.PollImmediate(5*time.Second, 3*time.Minute, func() (bool, error) {
+						_, err := h.Prometheusop().MonitoringV1().PrometheusRules(namespace).Create(context.TODO(), prome2etestrule, metav1.CreateOptions{})
+						if err != nil {
+							log.Printf("failed creating prometheus rules: %v", err)
+							return false, nil
```
mrbarge, line 217: I kind-of feel that if `Create()` has a problem even once, we should be treating this as a failure.

```suggestion
							return false, err
```

pkg/e2e/verify/namespace_webhook.go
```diff
@@ -159,6 +163,80 @@ var _ = ginkgo.Describe(namespaceWebhookTestName, func() {
...
 		}, viper.GetFloat64(config.Tests.PollingTimeout))
+
+		util.GinkgoIt("dedicated admin can not create PrometheusRule", func() {
+			for privilegedNamespace := range PRIVILEGED_NAMESPACES {
+				prome2etestrule := e2eCreatePrometheusRule(prometheusName, privilegedNamespace, DUMMY_USER, "dedicated-admins", h)
+				err := wait.PollImmediate(5*time.Second, 3*time.Minute, func() (bool, error) {
+					_, err := h.Prometheusop().MonitoringV1().PrometheusRules(privilegedNamespace).Create(context.TODO(), prome2etestrule, metav1.CreateOptions{})
+					if err != nil {
+						log.Printf("failed creating prometheus rules: %v", err)
+						return false, nil
```
mrbarge, line 174: I'm feeling like maybe there's room for optimization in these "we expect an error to happen" tests. If I'm reading it right, we expect `Create()` to error in this scenario, and this test continues to try a `Create()` for three minutes and keeps the poll going until it eventually times out after three minutes, returning an error - which is then the criteria for test success.

Aside from that, if the `Create()` call does succeed, it's returning `true, nil`, so afterwards `err` is going to be nil anyway. Which seems not correct when it then goes to check if the error `HasOccurred()`. 

Like, I'm thinking this needs to be flipped on its head and be something like
```
err := wait.PollImmediate(5*time.Second, 3*time.Minute, func() (bool, error) {
   _, err := h.Prometheusop().MonitoringV1().PrometheusRules(privilegedNamespace).Create(context.TODO(), prome2etestrule, metav1.CreateOptions{})
   if err != nil {
      log.Printf("failed creating prometheus rules: %v", err)
      return true, nil
   }
   return false, err
})
Expect(err).NotTo(HaveOccurred(), "PrometheusRules could be created")
```

I am not sure what use we get from verifying that `Create()` continues to fail over and over for X minutes.

pkg/e2e/verify/namespace_webhook.go
```diff
@@ -159,6 +163,80 @@ var _ = ginkgo.Describe(namespaceWebhookTestName, func() {
...
+					return true, nil
+				})
+				Expect(err).To(HaveOccurred(), "Could not create PrometheusRules")
+			}
+		}, viper.GetFloat64(config.Tests.PollingTimeout))
+
+		util.GinkgoIt("Non-privileged users cannot create prometheusrule in privileged namespaces", func() {
+			for privilegedNamespace := range PRIVILEGED_NAMESPACES {
+				prome2etestrule := e2eCreatePrometheusRule(prometheusName, privilegedNamespace, DUMMY_USER, DUMMY_GROUP, h)
+				err := wait.PollImmediate(5*time.Second, 3*time.Minute, func() (bool, error) {
```
mrbarge, line 185: My same comment from earlier applies to this one too I think.

mrbarge:
Just one more note, which you may already be looking at given the WIP status of the PR - the PrometheusRules will need to be cleaned up at the end of each test - or else one test will go to `Create` them and they will already exist from a prior test, and fail. So somewhere in the `JustAfterEach()` code:

https://github.com/openshift/osde2e/blob/989f91d3d1e279728d598a46401ac9506c6f7824/pkg/e2e/verify/namespace_webhook.go#L93

there'll need to be some new functionality introduced that checks for the PrometheusRule(s) your tests can potentially create, and removes them if they're there.

pkg/e2e/verify/namespace_webhook.go
mrbarge: Some more quick thoughts

- We could check and see what error the API will actually return in the scenario where the user doesn't have the RBAC. Only treat the specific RBAC-related error you expect from the API as the "success" criteria. And if the error is _not_ that, then that is cause to go back and have the poller retry.
- And on the topic of the poller, the three minute wait is a bit too generous. This should succeed very quickly. One minute max is probably enough.

pkg/e2e/verify/namespace_webhook.go
```diff
@@ -124,6 +128,13 @@ var _ = ginkgo.Describe(namespaceWebhookTestName, func() {
 				return true, nil
 			})
 		})
+		
+		ginkgo.JustAfterEach(func() {
+			h.Impersonate(rest.ImpersonationConfig{})
+			err := deletePrometheusRule(namespace, prometheusName, h)
```
mrbarge, line 134: So the goal of this function will be to clean up any PrometheusRules the tests might have created. In the tests currently, they are attempting to create PrometheusRules in a whole bunch of different namespaces (anything in `PRIVILEGED_NAMESPACES` or `NONPRIV_NAMESPACES`

Which means that the cleanup function should attempt to clean up any PromRule named `prometheusName` in every namespace in those lists.

Now naturally it is not going to exist in some of them, and that's fine - nothing to clean up! But that will be raised as an error by the delete.

So rather than treating every non-nil err returned as a test failure, we should be checking if the error is a "good" error. In this case, something like:

```
if err != nil and !errors.IsNotFound(err) {
   // this is not good and means that it was an error and wasn't due to the object not existing
}
```

(in this case, the `errors` package is `k8s.io/apimachinery/pkg/api/errors` which would need to be added to the imports.

mrbarge: Another thing: because this `JustAfterEach` is only relevant to the tests doing PromRule stuff, maybe all the PromRule tests can be packaged inside a nested `ginkgo.Describe("xxx", func(){  }` block .. along with this.. that way it should only execute for the promrule tests.

pkg/e2e/verify/namespace_webhook.go
```diff
@@ -159,6 +170,80 @@ var _ = ginkgo.Describe(namespaceWebhookTestName, func() {
 				Expect(err).NotTo(HaveOccurred())
 			}
 		}, viper.GetFloat64(config.Tests.PollingTimeout))
+
+		util.GinkgoIt("dedicated admin can not create PrometheusRule", func() {
+			for privilegedNamespace := range PRIVILEGED_NAMESPACES {
+				prome2etestrule := e2eCreatePrometheusRule(prometheusName, privilegedNamespace, DUMMY_USER, "dedicated-admins", h)
+				err := wait.PollImmediate(5*time.Second, 1*time.Minute, func() (bool, error) {
+					_, err := h.Prometheusop().MonitoringV1().PrometheusRules(privilegedNamespace).Create(context.TODO(), prome2etestrule, metav1.CreateOptions{})
+					if err != nil {
```
mrbarge, line 179: Do you know specifically what type of error you expect this to generate? This test may be best to explicitly check for that type of error. One of the helper functions in `k8s.io/apimachinery/pkg/api/errors` may then be able to help. 

ie. I wonder if it would be a `Forbidden` type of error? In which case:
```suggestion
					if err != nil && errors.IsForbidden(err) { 
```

mrbarge: The same will apply for the other places in the tests where we expect an RBAC failure to occur, ie the one right below this one.

pkg/e2e/verify/namespace_webhook.go
```diff
@@ -125,6 +130,24 @@ var _ = ginkgo.Describe(namespaceWebhookTestName, func() {
 			})
 		})
 
+		ginkgo.Describe("e2eCreatePrometheusRule", func() {
```
jbpratt, line 133: This is already inside another `Describe` block, it can be moved out. And please note that this test suite already has a `JustAfterEach` and `JustBeforeEach` set up on lines 76-94

pkg/e2e/verify/namespace_webhook.go
```diff
@@ -159,6 +182,80 @@ var _ = ginkgo.Describe(namespaceWebhookTestName, func() {
...
+			for _, nonPrivilegedNamespace := range NONPRIV_NAMESPACES {
+				prome2etestrule := e2eCreatePrometheusRule(prometheusName, nonPrivilegedNamespace, DUMMY_USER, "dedicated-admins", h)
+				err := wait.PollImmediate(5*time.Second, 3*time.Minute, func() (bool, error) {
+					_, err := h.Prometheusop().MonitoringV1().PrometheusRules(nonPrivilegedNamespace).Create(context.TODO(), prome2etestrule, metav1.CreateOptions{})
+					if err != nil && errors.IsForbidden(err) {
+						log.Printf("failed creating prometheus rules: %v", err)
+						return false, nil
+					}
+					return true, nil
+				})
```
jbpratt, line 255: all of the polling is the same except for the namespace you poll, can we please move this inside of the `e2eCreatePrometheusRule` function and just return an error?

jbpratt (changes requested):

pkg/e2e/verify/namespace_webhook.go
jbpratt: sorry, just noticed the differences in the error handling of the creation, disregard this comment!

reetika-vyas:
@mrbarge  Could you please review it once again?

reetika-vyas:
> Just one more note, which you may already be looking at given the WIP status of the PR - the PrometheusRules will need to be cleaned up at the end of each test - or else one test will go to `Create` them and they will already exist from a prior test, and fail. So somewhere in the `JustAfterEach()` code:
> 
> https://github.com/openshift/osde2e/blob/989f91d3d1e279728d598a46401ac9506c6f7824/pkg/e2e/verify/namespace_webhook.go#L93
> 
> there'll need to be some new functionality introduced that checks for the PrometheusRule(s) your tests can potentially create, and removes them if they're there.

added mentioned changes

reetika-vyas:
@jbpratt  Would you mind taking a look one more time?

pkg/e2e/verify/namespace_webhook.go
```diff
@@ -81,6 +86,18 @@ var _ = ginkgo.Describe(namespaceWebhookTestName, func() {
...
+				if manageNamespace {
+					err := e2eCreatePrometheusRule(prometheusName, privilegedNamespace, "system:admin", "", h)
+					Expect(err).NotTo(HaveOccurred())
+				}
+			}
+			for _, namespace := range NONPRIV_NAMESPACES {
+				err := e2eCreatePrometheusRule(prometheusName, namespace, DUMMY_USER, DUMMY_GROUP, h)
+				Expect(err).NotTo(HaveOccurred())
+			}
+
```
mrbarge, line 100: This whole block doesn't feel like it is necessary - the tests are the ones that are testing the creation of the rule. 

The `JustBeforeEach()` can be used to ensure that some test pre-requisites are performed or satisfied before the tests run (https://onsi.github.io/ginkgo/#separating-creation-and-configuration-justbeforeeach). But the `PrometheusRules` don't need to be created there, the actual tests will do that.

However, it is fine to have the cleanup occur in the `JustAfterEach()` as we want to ensure that if any PrometheusRules were created, they need to have been deleted at the end of the test / before any other test runs.

pkg/e2e/verify/namespace_webhook.go
```diff
@@ -254,6 +357,15 @@ func deleteNamespace(ctx context.Context, namespace string, waitForDelete bool,
 	return err
 }
 
+func deletePrometheusRule(namespace, prometheusName string, h *helper.H) error {
+	log.Printf("Deleting prometheusrule for prometheusrule validation webhook (%s)", prometheusName)
+	err := h.Prometheusop().MonitoringV1().PrometheusRules(namespace).Delete(context.TODO(), prometheusName, metav1.DeleteOptions{})
+	if err != nil {
```
mrbarge, line 363: This will need to account for the situation that the PrometheusRule never existed in order to be deleted (which I'm _guessing_ should return an error). 

So something like (untested..):
```suggestion
	if err != nil && !errors.IsNotFound(err) {
```

pkg/e2e/verify/namespace_webhook.go
```diff
@@ -157,6 +186,80 @@ var _ = ginkgo.Describe(namespaceWebhookTestName, func() {
...
 			}
 		}, viper.GetFloat64(config.Tests.PollingTimeout))
+
+		util.GinkgoIt("dedicated admin can not create PrometheusRule", func(ctx context.Context) {
+			for privilegedNamespace := range PRIVILEGED_NAMESPACES {
+				prome2etestrule := e2eCreatePrometheusRule(prometheusName, privilegedNamespace, DUMMY_USER, "dedicated-admins", h)
+				err := wait.PollImmediate(5*time.Second, 1*time.Minute, func() (bool, error) {
+					_, err := h.Prometheusop().MonitoringV1().PrometheusRules(privilegedNamespace).Create(ctx, prome2etestrule, metav1.CreateOptions{})
+					if err != nil && errors.IsForbidden(err) {
+						log.Printf("failed creating prometheus rules: %v", err)
```
mrbarge, line 196: This log could end up being misleading if someone's reviewing raw test logs. The "Failure" in this case is good and indicative of test success. I'd lean towards removing this log.

pkg/e2e/verify/namespace_webhook.go
```diff
@@ -157,6 +186,80 @@ var _ = ginkgo.Describe(namespaceWebhookTestName, func() {
...
+		util.GinkgoIt("Non-privileged users cannot create prometheusrule in privileged namespaces", func(ctx context.Context) {
+			for privilegedNamespace := range PRIVILEGED_NAMESPACES {
+				prome2etestrule := e2eCreatePrometheusRule(prometheusName, privilegedNamespace, DUMMY_USER, DUMMY_GROUP, h)
+				err := wait.PollImmediate(5*time.Second, 1*time.Minute, func() (bool, error) {
+					_, err := h.Prometheusop().MonitoringV1().PrometheusRules(privilegedNamespace).Create(ctx, prome2etestrule, metav1.CreateOptions{})
+					if err != nil && errors.IsForbidden(err) {
+						log.Printf("failed creating prometheus rules: %v", err)
+						return true, nil
+					}
+					return false, fmt.Errorf("prometheusrule could be created")
```
mrbarge, line 214: Return the actual error here rather than creating a new one, that way if `err` is nil it will actually trigger the re-poll.

```suggestion
					return false, err
```

mrbarge:
@reetika-vyas - sorry for the delay - left some more feedback.

Have you also tried running these tests on a cluster with `osde2e` and can you confirm they are passing ok?

ie. following https://github.com/openshift/ops-sop/blob/master/v4/knowledge_base/run-osde2e-tests.md 
and you can use `--focus-tests` to hone in on this file.

ie.
```
export OCM_TOKEN="..."
osde2e test -i $CLUSTERID -e stage --kubeconfig $KUBECONFIG --focus-tests "dedicated admin can not create PrometheusRule"
```

Maybe start with one of the tests you're introducing, make sure that's all working fine, then try another.

mrbarge:
None of the tests appear to be passing because they're not currently set up to do any user impersonation.

If we have a look at the `updateNamespaces` function, right at the start we can see how it sets up user impersonation before it does the actual update of the namespace. What will need to happen is similar logic for this PR.

I recommend taking advantage of the existing `e2eCreatePrometheusRule` function that has been written for this PR - which has the desired to-be-impersonated user and group being passed in, just not being used - to do that. That will also clean up a lot of the tests and make them tidier.

In that function, let's keep it simple. Let's get rid of the `wait` loop which seems to just be adding confusion and complexity. Call `Create`, pass the error back up.

Then, change all the new tests to instead call this function. Depending on what is being tested, they will either expect an error, or they will not. If they expect an error, the test should verify what sort of error was raised, so that we know it was due to the user being forbidden.

To get you going, here is what the `e2eCreatePrometheusRule` function _might_ look like (I've not really given this more than a moment's thought)

```
func e2eCreatePrometheusRule(ctx context.Context, prometheusrulename, namespace string, asUser string, userGroup string, h *helper.H) error {

	// reset impersonation upon return
	defer h.Impersonate(rest.ImpersonationConfig{})

	// reset impersonation at the beginning just-in-case
	h.Impersonate(rest.ImpersonationConfig{})

	// we need to add these groups for impersonation to work
	userGroups := []string{"system:authenticated", "system:authenticated:oauth"}
	if userGroup != "" {
		userGroups = append(userGroups, userGroup)
	}

	// update the namespace as our desired user
	h.Impersonate(rest.ImpersonationConfig{
		UserName: asUser,
		Groups:   userGroups,
	})

	e2eprometheusrule := &monv1.PrometheusRule{
		TypeMeta: metav1.TypeMeta{},
		ObjectMeta: metav1.ObjectMeta{
			Name:      prometheusrulename,
			Namespace: namespace,
		},
		Spec: monv1.PrometheusRuleSpec{
			Groups: []monv1.RuleGroup{
				{
					Name: "example",
					Rules: []monv1.Rule{
						{
							Alert: "VersionAlert",
							Expr: intstr.IntOrString{
								StrVal: "version{job==\"prometheus-example-app\"} == 0",
							},
						},
					},
				},
			},
		},
	}

	_, err := h.Prometheusop().MonitoringV1().PrometheusRules(namespace).Create(ctx, e2eprometheusrule, metav1.CreateOptions{})
	return err
}
```

And then one test, refactored, would look like this:

```
		util.GinkgoIt("dedicated admin can not create PrometheusRule", func(ctx context.Context) {
			for privilegedNamespace := range PRIVILEGED_NAMESPACES {
				err := e2eCreatePrometheusRule(ctx, prometheusName, privilegedNamespace, DUMMY_USER, "dedicated-admins", h)
				Expect(apierrors.IsForbidden(err)).To(BeTrue())
			}
		}, viper.GetFloat64(config.Tests.PollingTimeout))
```

(the rest would all also need to change)

Additionally, please ensure that we've tested this on a cluster before calling for further reviews. You can do this by:
- Creating a staging cluster and getting the kubadmin creds
- Setting `OCM_TOKEN` environment variable to be your OCM token
- Running osde2e as follows: `out/osde2e test -e stage -k /path/to/cluster/kubeadmin-cred-file --focus-tests "namespace validating webhook"`

You can substitute `namespace validating webhook` for the specific name of one test, ie. `dedicated admin can not create PrometheusRule`

pkg/e2e/verify/namespace_webhook.go
```diff
@@ -254,6 +305,15 @@ func deleteNamespace(ctx context.Context, namespace string, waitForDelete bool,
 	return err
 }
 
+func deletePrometheusRule(namespace, prometheusName string, h *helper.H) error {
+	log.Printf("Deleting prometheusrule for prometheusrule validation webhook (%s)", prometheusName)
+	err := h.Prometheusop().MonitoringV1().PrometheusRules(namespace).Delete(context.TODO(), prometheusName, metav1.DeleteOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		return fmt.Errorf("failed to delete prometheusrule from managed namespace '%s': %v", prometheusName, err)
+	}
+	return err
```
mrbarge, line 314: This should just return nil as it is fine for the function to attempt to delete a PrometheusRule that does not exist.
```suggestion
	return nil
```

reetika-vyas:
@mrbarge I have tested it on stage cluster for 3 tests, adding the result below for one test. For last one I was not able to test, my cluster deleted everytime I ran tests. 
```
2022/12/05 10:06:08 namespace_webhook.go:230: Deleting group for namespace validation webhook (random-group-name)
2022/12/05 10:06:08 namespace_webhook.go:289: Deleting namespace for namespace validation webhook (redhat-ocm-addon-test-operator)
2022/12/05 10:06:08 namespace_webhook.go:289: Deleting namespace for namespace validation webhook (mykube-admin)
2022/12/05 10:06:08 namespace_webhook.go:289: Deleting namespace for namespace validation webhook (open-shift)
2022/12/05 10:06:09 namespace_webhook.go:289: Deleting namespace for namespace validation webhook (oopenshift)
2022/12/05 10:06:09 namespace_webhook.go:289: Deleting namespace for namespace validation webhook (ops-health-monitoring-foo)
2022/12/05 10:06:09 namespace_webhook.go:289: Deleting namespace for namespace validation webhook (default-user)
2022/12/05 10:06:09 namespace_webhook.go:289: Deleting namespace for namespace validation webhook (logger)
2022/12/05 10:06:09 namespace_webhook.go:310: Deleting prometheusrule for prometheusrule validation webhook (prometheus-example-app)
2022/12/05 10:06:10 namespace_webhook.go:310: Deleting prometheusrule for prometheusrule validation webhook (prometheus-example-app)
2022/12/05 10:06:10 namespace_webhook.go:310: Deleting prometheusrule for prometheusrule validation webhook (prometheus-example-app)
2022/12/05 10:06:10 namespace_webhook.go:310: Deleting prometheusrule for prometheusrule validation webhook (prometheus-example-app)
2022/12/05 10:06:10 namespace_webhook.go:310: Deleting prometheusrule for prometheusrule validation webhook (prometheus-example-app)
2022/12/05 10:06:10 namespace_webhook.go:310: Deleting prometheusrule for prometheusrule validation webhook (prometheus-example-app)
2022/12/05 10:06:11 namespace_webhook.go:310: Deleting prometheusrule for prometheusrule validation webhook (prometheus-example-app)
•SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS SUCCESS! 16.090721375s 2022/12/05 10:06:20 e2e.go:513: Skipping metrics upload for local osde2e run.
2022/12/05 10:06:21 helper.go:126: Setting project name to osde2e-qz2hm
2022/12/05 10:06:21 helper.go:265: ServiceAccount is now set to ``
2022/12/05 10:06:21 e2e.go:567: Running Must Gather...
2022/12/05 10:06:21 helper.go:265: ServiceAccount is now set to `system:serviceaccount:osde2e-qz2hm:cluster-admin`
2022/12/05 10:06:22 runner.go:111: Using 'quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:f9adc9c73195bac0a4eaed2dd39fe00c717db19898e5eee16f6330aab032375b' as image for runner
```

mrbarge:
Thanks for working on those changes @reetika-vyas 
/lgtm

PR #1380, opened 2022-11-09
support setting ginkgo loglevel

jbpratt:
allow for controlling the ginkgo reporter output which can be set to one
of `succinct` (default), `v` (verbose, printing GinkgoWriter output) and
`vv` (very-verbose, including skipped and pending tests). This will
allow tests to then use the `GinkgoWriter` and improve the output of
osde2e.

https://onsi.github.io/ginkgo/#controlling-ginkgos-output


MrSantamaria:
I'm trying to look at the artifacts for this one but the gcp bucket seems to be down? 
Did you check if this had an impact on the output of the junit file? 

Ginkgo 1 didn't support xml output and I'm not sure if this affected that flow at all.

jbpratt:
> I'm trying to look at the artifacts for this one but the gcp bucket seems to be down? Did you check if this had an impact on the output of the junit file?
> 
> Ginkgo 1 didn't support xml output and I'm not sure if this affected that flow at all.

no behavior should be changed here for any jobs, this is just allowing to set the log level (still defaulting to succinct)

PR #1383, opened 2022-11-09
providers/ocmprovider: remove ccs overwrite logic

jbpratt:
remove the ccs.overwrite setting and logic as it is no longer needed/used

### DIFF
--- a/pkg/common/providers/ocmprovider/cluster.go
+++ b/pkg/common/providers/ocmprovider/cluster.go
@@ -202,18 +202,9 @@ func (o *OCMProvider) LaunchCluster(clusterName string) (string, error) {
 					}
 				}
 			} else {
-				var err error
-				var ccsUser string
-				ccsUser, err = aws.VerifyCCS()
+				_, err = aws.VerifyCCS()
 				if err != nil {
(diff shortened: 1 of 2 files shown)


MrSantamaria:
We can remove this whole logic. 
This is a remnant attempt of mine trying to figure out multi clusters in a single account back when osdccsadmin was a thing.

jbpratt:
> We can remove this whole logic. This is a remnant attempt of mine trying to figure out multi clusters in a single account back when osdccsadmin was a thing.

Thanks for the note, updated to remove all the overwrite logic!

FWIW I still do multiple ccs clusters (and rosa sts) in the same account, osdCcsAdmin is still a thing that I use :smile_cat:

MrSantamaria:
It is something we can use for sure. 
In the olden days anytime `rosa init` was called it cycled keys and it was near impossible to keep aws credentials for the osd flavors.

jbpratt:
> It is something we can use for sure. In the olden days anytime `rosa init` was called it cycled keys and it was near impossible to keep aws credentials for the osd flavors.

:smirk: https://github.com/openshift/rosa/issues/637

PR #1387, opened 2022-11-10
bug fix: rosa provider not loading stage environment  

ritmun:
We are currently using "rosa,stage" config files to load sage environemnt for rosa pipeline. However, rosa provider is not picking up the environment variable set in stage.yaml file. 

The approach here uses the var set in stage.yaml as is, which is `ocm.env`. rosaprovider is actually a type of ocmprovider, so it seems appropriate to use ocm.env as the environment for rosaprovider. That is the fix applied here.

A different approach to this fix was to define `rosa.env`  in stage.yaml file, which I tested locally, however, viper is refusing to read that variable.

I've tested this locally and it picks up the correct environment per the stage.yaml config file.

### DIFF
--- a/configs/rosa-stage.yaml
+++ b/configs/rosa-stage.yaml
@@ -1,0 +1,1 @@
+provider: rosa
--- a/configs/rosa-stage.yaml
+++ b/configs/rosa-stage.yaml
@@ -1,3 +1,3 @@
 provider: rosa
 rosa:
-  Env: stage
+  env: stage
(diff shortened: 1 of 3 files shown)


MrSantamaria:
This isn't a bug so much as a badly implemented feature. 

The Rosa and ocm flows use different variables to define the environment. 
The env variable ROSA_ENV sets the appropriate env when using Rosa. 
I want to get away from setting our variables in all these different places and focus on our config file.

PR #1388, opened 2022-11-10
[WIP] allow passing args as env vars to addon test harness

ritmun:
This change allows passing the following args as environment variables:

CONFIGS
SECRET_LOCATIONS
CLUSTER_ID
ENV
SKIP_HEALTH_CHECK
MUST_GATHER

[might need to make another pass at this since some of these vars might not need to be passed as env]

### DIFF
--- a/cmd/osde2e/test/cmd.go
+++ b/cmd/osde2e/test/cmd.go
@@ -4,6 +4,7 @@ import (
 	"log"
 	"math/rand"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/openshift/osde2e/cmd/osde2e/common"
@@ -55,7 +56,7 @@ func init() {
 	pfs.StringVar(
 		&args.configString,
 		"configs",
-		"",
+		os.Getenv("CONFIGS"),
 		"A comma separated list of built in configs to use",
 	)
 	Cmd.RegisterFlagCompletionFunc("configs", helpers.ConfigComplete)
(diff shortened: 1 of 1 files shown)


MrSantamaria:
I been thinking on something like this PR for a while. 
We as a team need to define what/where we want end users to set/grab these explicit flags/args/configs. 
CLUSTER_ID for example is passed as an args in the upstream clis so we shouldn't change the behavior in osde2e. 
CONFIGS and SECRET_LOCATIONS we have always passed as args, and for us to understand how osde2e is running it's best we force consumption through the command. 
ENV can be deprecated from this logic. In all our pipelines we pass the environment as a config, here we change our behavior from upstream since osde2e black boxes the logic from the consumer.

SKIP_HEALTH_CHECK and MUST_GATHER are up for debate. Personally those flags are set under special conditions and most likely for debugging, so passing a debug config that groups those together might be a better route.

cmd/osde2e/test/cmd.go
```diff
@@ -117,10 +123,15 @@ func init() {
 		"",
 		"Skip any Ginkgo tests whose names match the regular expression.",
 	)
+
+	var mg = false
+	if strings.EqualFold(os.Getenv( "MUST_GATHER" ),"true"){
+		mg = true
+	}
```
jbpratt, line 130:
```suggestion
	 mg :=  strings.EqualFold(os.Getenv("MUST_GATHER" ), "true")
```

jbpratt:
/close

stale, closing to stop jobs from running

PR #1404, opened 2022-11-17
[OSD-13651] add test for dedicated admin can patch console

a7vicky:
Add test case for dedicated admin user:
-  can patch **consoles.operator.openshift.io**

_Fix_ :  [OSD-13651](https://issues.redhat.com/browse/OSD-13651)

### DIFF
--- a/pkg/e2e/osd/dedicatedadmin.go
+++ b/pkg/e2e/osd/dedicatedadmin.go
@@ -17,6 +17,8 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/util/retry"
 )
@@ -267,6 +269,27 @@ var _ = ginkgo.Describe(dedicatedAdminTestName, func() {
 			}()
 
 			err := manageSubscriptions(ctx, namespaceList, h)
+			Expect(err).NotTo(HaveOccurred())
+		}, float64(viper.GetFloat64(config.Tests.PollingTimeout)))
+
+		util.GinkgoIt("dedicated-admin user can patch consoles.operator.openshift.io CR", func(ctx context.Context) {
+			// Impersonate ded-admin
+			h.Impersonate(rest.ImpersonationConfig{
+				UserName: "dummy-admin@redhat.com",
+				Groups: []string{
+					"dedicated-admin",
+				},
+			})
+			defer func() {
+				h.Impersonate(rest.ImpersonationConfig{})
+			}()
+
+			patchData := []byte(fmt.Sprintf(`{"spec":{"plugins":["test"]}}`))
+
+			_, err := h.Dynamic().Resource(schema.GroupVersionResource{
+				Group: "operator.openshift.io", Version: "v1",
+				Resource: "consoles",
+			}).Patch(ctx, "cluster", types.MergePatchType, patchData, metav1.PatchOptions{})
 			Expect(err).NotTo(HaveOccurred())
 		}, float64(viper.GetFloat64(config.Tests.PollingTimeout)))
 	})
--- a/pkg/e2e/osd/dedicatedadmin.go
+++ b/pkg/e2e/osd/dedicatedadmin.go
@@ -277,7 +277,7 @@ var _ = ginkgo.Describe(dedicatedAdminTestName, func() {
 			h.Impersonate(rest.ImpersonationConfig{
 				UserName: "dummy-admin@redhat.com",
 				Groups: []string{
-					"dedicated-admin",
+					"dedicated-admins",
 				},
 			})
 			defer func() {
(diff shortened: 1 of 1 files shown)


pkg/e2e/osd/dedicatedadmin.go
```diff
@@ -269,6 +271,34 @@ var _ = ginkgo.Describe(dedicatedAdminTestName, func() {
 			err := manageSubscriptions(ctx, namespaceList, h)
 			Expect(err).NotTo(HaveOccurred())
 		}, float64(viper.GetFloat64(config.Tests.PollingTimeout)))
+
+		// dedicated-admin user can patch consoles.operator.openshift.io CR
```
jbpratt, line 275: unneeded comment (prefer `It`, `By,`, or `Context` to inline comments)

pkg/e2e/osd/dedicatedadmin.go
```diff
@@ -269,6 +271,34 @@ var _ = ginkgo.Describe(dedicatedAdminTestName, func() {
...
+		util.GinkgoIt("ded-admin user can patch consoles.operator.openshift.io CR", func(ctx context.Context) {
```
jbpratt, line 276:
```suggestion
		util.GinkgoIt("dedicated-admin user can patch consoles.operator.openshift.io CR", func(ctx context.Context) {
```

this will be part of the test case name

jbpratt (changes requested):

pkg/e2e/osd/dedicatedadmin.go
```diff
@@ -269,6 +271,37 @@ var _ = ginkgo.Describe(dedicatedAdminTestName, func() {
...
+			_, err := h.Dynamic().Resource(schema.GroupVersionResource{
+				Group: "operator.openshift.io", Version: "v1",
+				Resource: "consoles",
+			}).Patch(ctx, "cluster", types.MergePatchType, patchData, metav1.PatchOptions{})
+			Expect(err).NotTo(HaveOccurred())
+
+			// revret the changes
+			patchEmpty := []byte(fmt.Sprintf(`{"spec":{"plugins":[""]}}`))
+
+			_, err1 := h.Dynamic().Resource(schema.GroupVersionResource{
```
bmeng, line 299: should this be named as `err` as well?

pkg/e2e/osd/dedicatedadmin.go
bmeng: this looks a bit weird, but it is the same for all the other test cases in this file...

pkg/e2e/osd/dedicatedadmin.go
```diff
@@ -269,6 +271,37 @@ var _ = ginkgo.Describe(dedicatedAdminTestName, func() {
...
+			patchData := []byte(fmt.Sprintf(`{"spec":{"plugins":["test"]}}`))
+
+			_, err := h.Dynamic().Resource(schema.GroupVersionResource{
+				Group: "operator.openshift.io", Version: "v1",
+				Resource: "consoles",
+			}).Patch(ctx, "cluster", types.MergePatchType, patchData, metav1.PatchOptions{})
+			Expect(err).NotTo(HaveOccurred())
+
+			// revret the changes
+			patchEmpty := []byte(fmt.Sprintf(`{"spec":{"plugins":[""]}}`))
```
bmeng, line 297: I am not sure if patch it the second time to revert is necessary.
As the patched console plugin is dummy, it should not have problem to just leave it there?
Or if we would like to check the permission only, it should also be ok if we do not patch the `.spec`

pkg/e2e/osd/dedicatedadmin.go
a7vicky: Thank you for the rebview. I have address this in my latest commit

pkg/e2e/osd/dedicatedadmin.go
a7vicky: Thank you @bmeng I have removed the revert patch process as this is harmless to be patch with any dummy data.

pkg/e2e/osd/dedicatedadmin.go
a7vicky: This has been addressed in my latest commit.

pkg/e2e/osd/dedicatedadmin.go
jbpratt: Yes, it does look a bit weird and to be honest every test case name in this file is redundant it should be fixed eventually. The full test case name for this test for example is `[Suite: informing] [OSD] dedicated-admin permissions dedicated-admin group permissions dedicated-admin user can patch consoles.operator.openshift.io CR` https://github.com/kubernetes/community/blob/master/contributors/devel/sig-testing/writing-good-e2e-tests.md#some-concrete-tools-at-your-disposal

```
     In particular be aware that every test is uniquely identified and described (e.g. in test reports) by the concatenation of its Describe clause and nested It clauses. So for example Describe("Pods",...).... It(""should be scheduled with cpu and memory limits") produces a sane test identifier and descriptor Pods should be scheduled with cpu and memory limits, which makes it clear what's being tested, and hence what's not working if it fails. Other good examples include:

   CAdvisor should be healthy on every node

and

   Daemon set should run and stop complex daemon

On the contrary (these are real examples), the following are less good test descriptors:

   KubeProxy should test kube-proxy

and

Nodes [Disruptive] Network when a node becomes unreachable
[replication controller] recreates pods scheduled on the
unreachable node AND allows scheduling of pods on a node after
it rejoins the cluster

An improvement might be

Unreachable nodes are evacuated and then repopulated upon rejoining [Disruptive]
```

pkg/e2e/osd/dedicatedadmin.go
mrbarge: I'd generally recommend that tests should be leaving the cluster in the state that the cluster originally started in, and therefore it should be doing the removal. 

If a test is re-run on a cluster and the previous execution of the test has left the cluster in a different state, it's testing a different situation.

pkg/e2e/osd/dedicatedadmin.go
```diff
@@ -269,6 +271,27 @@ var _ = ginkgo.Describe(dedicatedAdminTestName, func() {
 			err := manageSubscriptions(ctx, namespaceList, h)
 			Expect(err).NotTo(HaveOccurred())
 		}, float64(viper.GetFloat64(config.Tests.PollingTimeout)))
+
+		util.GinkgoIt("dedicated-admin user can patch consoles.operator.openshift.io CR", func(ctx context.Context) {
+			// Impersonate ded-admin
+			h.Impersonate(rest.ImpersonationConfig{
+				UserName: "dummy-admin@redhat.com",
+				Groups: []string{
+					"dedicated-admin",
```
mrbarge, line 280: The group name needs to be corrected here for the test to pass

```suggestion
					"dedicated-admins",
```

a7vicky: Thanks @mrbarge  for the review. I have updated this in my latest commit.

mrbarge:
Thanks for the changes @a7vicky , feel free to unhold if you're good for it to go.

PR #1413, opened 2022-11-21
common: load kubeconfig on helper setup

jbpratt:
loading kubeconfig on helper setup allows for running the tests directly with ginkgo and an existing cluster (outside of the framework) which is easier for development

also add additional check to no-op `LoadKubeconfig` after the first call (we call helper.Setup() a lot)


MrSantamaria:
Since we get a kubeconfig from OCM, I'd even go the route of creating a sync.Once.

ryankwilliams (approved):

PR #1419, opened 2022-11-22
Add support for using ginkgo labels

ryankwilliams:
# Change

This PR adds the initial ground work for enabling the ability to use ginkgo labels. It provides the ability for osde2e to accept a label filter query to pass to ginkgo. Where ginkgo can use that to decide which test cases it should execute.

# Why

Labels are going to play a large role as we work towards defining standardization and consistency across all OSD operator tests. Labels are a way to easily tag test cases.

With the use of labels, we can fine tune new tests and existing tests cases based on a standard set of supported labels (list will be provided next week for review/discussion).

Labels can be applied to both ginkgo describe nodes (test suites) and ginkgo it nodes (test cases).

*Example*: Test suites can define a label such as `operators` or `<specific-operator>`, then within the suite, test cases can be tagged according to their represented category (e.g. `smoke`, `e2e`, `upgrade`). Allowing to build label filter queries such as:
   * `--label-filter "smoke"` << run everything tagged as smoke tests
   * `--label-filter "operators && smoke"` << run all operators smoke tests
   * `--label-filter "<specific-operator> && smoke"` << run the specific operator smoke tests
   * `--label-filter "e2e"` << run everything tagged as e2e tests
   * `--label-filter "e2e && smoke"` << run all smoke tests that are part of the e2e suite

These are just some basic examples above to demonstrate the ability of using labels.

# Additional Info

* With the introduction to support labels, focus strings still works. They can work together with labels.
* Separate PR's will be created to tag test cases as we work towards a few selected OSD operators to start with.

### DIFF
--- a/cmd/osde2e/test/cmd.go
+++ b/cmd/osde2e/test/cmd.go
@@ -49,6 +49,7 @@ var args struct {
 	mustGather       bool
 	focusTests       string
 	skipTests        string
+	labelFilter      string
 }
 
 func init() {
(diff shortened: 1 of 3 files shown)


MrSantamaria:
This would be a good replacement for the ginkgo focus we have been using.

PR #1425, opened 2022-11-25
Update and trim dependencies

cblecker:
This PR updates and trims off some dependencies, in order to resolve some security alerts.

Additionally, I've trimmed down the replace stanza to only those that are strictly required.


cblecker:
/hold
This is a bit of a bigger PR, so I'd like extra eyes on it before merge.

jbpratt:
/lgtm

tested locally, thanks for the updates!

PR #1433, opened 2022-12-01
Introduce e2e label

jbpratt:
Amend the `label.E2E` to each `Describe` within the `Suite: e2e` to
allow executing these tests based on label rather than focus-filter.

Focus filter will continue working as is. Eventually (once the initial
label introduction is complete), the configs will be updated to use
labels and/or focus where applicable.

depends on #1432

/hold


ryankwilliams:
For https://issues.redhat.com/browse/SDCICD-859

ryankwilliams:
PR depends on #1447

PR #1434, opened 2022-12-01
common/helper: remove dedicated-admin service account deletion

jbpratt:
deletion is happening after the namespace deletion which takes this service account with it, an error is always printed for this failed:

"Error deleting dedicated-admin serviceaccount for 'osde2e-omwhi': serviceaccounts "osde2e-omwhi" not found"

### DIFF
--- a/pkg/common/helper/helper.go
+++ b/pkg/common/helper/helper.go
@@ -171,10 +171,6 @@ func (h *H) Cleanup(ctx context.Context) {
 			if err != nil {
 				log.Printf("Error deleting project `%s` in Cleanup(): %s", project.Name, err.Error())
 			}
-			err = h.Kube().CoreV1().ServiceAccounts("dedicated-admin").Delete(ctx, project.Name, metav1.DeleteOptions{})
-			if err != nil {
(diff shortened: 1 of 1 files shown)


MrSantamaria:
The goal of the cleanup func precedes my times but I think the goal was to have osde2e attempt to remove any created resources in a recovery function. 
Should we outright remove it from here or add a "if exist then delete" type of logic?

jbpratt:
> The goal of the cleanup func precedes my times but I think the goal was to have osde2e attempt to remove any created resources in a recovery function. Should we outright remove it from here or add a "if exist then delete" type of logic?

the service account is namespace scoped and we delete the namespace right above it so it takes all resources within the namespace with it, so that when we hit this logic the SA already is deleted with the namespace delete.

MrSantamaria:
Hmm it seems to have been implemented for the use of reusing Hibernating Clusters. 
I'd assume in the case a "completed-" cluster didn't fully clean up. 

Either ways you are right the sa is tied to the ns, and we also don't support hibernating clusters anymore. 
Maybe we will see that issue arise once we start using long running clusters? But that's a problem for that time. 
/lgtm

PR #1443, opened 2022-12-02
Sdcicd 708 addon metadata

ritmun:
Jira: https://issues.redhat.com/browse/SDCICD-708

Addon test suite now writes addon-metadata.json file with addon id and version

To test locally, run the example addon harness per https://github.com/openshift/osde2e-example-test-harness/blob/65ad69888c0a71eaf6dfb9ac7fb48a4d84dbbdcd/README.md#locally-running-this-example 

log should include a line `writing addon metadata to <filepath-filename>`


pkg/common/metadata/addon.go
```diff
@@ -0,0 +1,55 @@
...
+	Version string `json:"version,string"`
+	ID      string `json:"id,string"`
+}
+
+// Instance is the global metadata instance
+var AddonInstance *AddonMetadata
+
+func init() {
+	AddonInstance = &AddonMetadata{}
+}
```
jbpratt, line 26: I don't see this global being used, can it be removed?

pkg/common/metadata/addon.go
```diff
@@ -0,0 +1,51 @@
...
+// addonMetadata houses metadata to be written out to the additional-metadata.json
+type AddonMetadata struct {
+	// Whether the CRD was found. Typically Spyglass seems to have issues displaying non-strings, so
+	// this will be written out as a string despite the native JSON boolean type.
+	Version string `json:"version,string"`
+	ID      string `json:"id,string"`
+}
+
+// Instance is the global metadata instance
+var AddonInstance *AddonMetadata
```
jbpratt, line 22: I think this global can be removed

jbpratt (changes requested):

go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	github.com/kyleconroy/sqlc v1.16.0
 	github.com/kylelemons/godebug v1.1.0
 	github.com/lib/pq v1.10.7
+	github.com/markbates/pkger v0.15.1
```
jbpratt, line 25: Hmm, not sure why this is being added, perhaps an accident? I removed pkger in #1057

ritmun: thanks for pointing out! It stayed in my gomod because pkged.go file didn't get deleted since it's ignored in git, so it kept pulling the dep. Just manually cleaned up and it went away.

ritmun:
> /approve

Thanks! Do you mind adding lgtm as well? Looks like tide needs both approved and lgtm.

PR #1447, opened 2022-12-06
Generate junit report with customized settings after tests finished

ryankwilliams:
# Change

The current implementation will generate the junit report during the
RunSpecs function call using the standard junit report config settings
defined.

With the introduction of labels "test tagging", the labels are now
going to be appended to the test case name within the junit xml. Every
time a label is either added/modified/deleted, it will alter the test
case name. Presenting different test case names when viewing results
from local runs, runs in prow or results in test grid.

This commit modifies osde2e to generate the junit report with config
settings after the test have finished. Allowing for customizing the way
the junit report is built.

Initial junit report config settings are:
 1. Omit labels from displaying within test case names

Needed to tag ginkgo to a specific commit to get the ability to alter
test case names. Can be removed once a new ginkgo version is released. 

https://github.com/onsi/ginkgo/commit/de44005c2cdb08a64a7efbb0fd59392a7eb05b61

# Verification

*Before omit labels from test case name*

```
cat junit_07zj7.xml | grep prune | head -n 1
          <testcase name="[It] [Suite: operators] [OSD] Prune jobs pruner jobs should works builds-pruner should run successfully [operators]" classname="OSD e2e suite" status="passed" time="10.217130325">
```

*After omit labels from test case name*

```
cat junit_0uc1x.xml | grep prune | head -n 1
          <testcase name="[It] [Suite: operators] [OSD] Prune jobs pruner jobs should works builds-pruner should run successfully" classname="OSD e2e suite" status="passed" time="10.262170541">
```

# For [SDCICD-863](https://issues.redhat.com//browse/SDCICD-863)

### DIFF
--- a/pkg/e2e/e2e.go
+++ b/pkg/e2e/e2e.go
@@ -36,6 +36,7 @@ import (
 	viper "github.com/openshift/osde2e/pkg/common/concurrentviper"
 	"github.com/openshift/osde2e/pkg/db"
 
+	"github.com/onsi/ginkgo/v2/reporters"
 	"github.com/openshift/osde2e/pkg/common/aws"
 	"github.com/openshift/osde2e/pkg/common/cluster"
 	clusterutil "github.com/openshift/osde2e/pkg/common/cluster"
@@ -741,7 +742,6 @@ func runTestsInPhase(
 		}
 	}
 	suffix := viper.GetString(config.Suffix)
-	reporterConfig.JUnitReport = filepath.Join(phaseDirectory, fmt.Sprintf("junit_%v.xml", suffix))
 	ginkgoPassed := false
 
 	if !suiteConfig.DryRun {
@@ -750,6 +750,18 @@ func runTestsInPhase(
 			return false, testCaseData
 		}
 	}
+
+	// Generate JUnit report once all tests have finished with customized settings
+	_ = ginkgo.ReportAfterSuite("OSDE2E", func(report ginkgo.Report) {
+		err := reporters.GenerateJUnitReportWithConfig(
+			report,
+			filepath.Join(phaseDirectory, fmt.Sprintf("junit_%v.xml", suffix)),
+			reporters.JunitReportConfig{OmitSpecLabels: true},
+		)
+		if err != nil {
+			log.Printf("error creating junit report file %s", err.Error())
+		}
+	})
 
 	// We need this anonymous function to make sure GinkgoRecover runs where we want it to
 	// and will still execute the rest of the function regardless whether the tests pass or fail.
--- a/go.mod
+++ b/go.mod
@@ -23,7 +23,7 @@ require (
 	github.com/kylelemons/godebug v1.1.0
 	github.com/lib/pq v1.10.7
 	github.com/mitchellh/mapstructure v1.5.0
-	github.com/onsi/ginkgo/v2 v2.5.1
+	github.com/onsi/ginkgo/v2 v2.5.2-0.20221206015719-de44005c2cdb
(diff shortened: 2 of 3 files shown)


pkg/e2e/e2e.go
```diff
@@ -751,6 +751,15 @@ func runTestsInPhase(
 		}
 	}
 
+	// Generate JUnit report once all tests have finished with customized settings
+	ginkgo.ReportAfterSuite("OSDE2E", func(report ginkgo.Report) {
+		reporters.GenerateJUnitReportWithConfig(
```
jbpratt, line 756: need to check err https://cs.github.com/onsi/ginkgo/blob/880485988b4b99c965b2192865e2bbe69a58dd01/reporters/junit_report_test.go?q=reporters.GenerateJUnitReportWithConfig#L218-L223

MrSantamaria:
Devil's advocate, but would it be a good idea to post the labels? 
In regards to the current logic some test flip a switch based on something like CCS. 
The [Suite: operators]...operators example would for sure be redundant, but maybe having the labels to show the configurations the test is taking in might be useful? 
This set of labels could also get recorded somewhere else.

ryankwilliams:
> Devil's advocate, but would it be a good idea to post the labels? In regards to the current logic some test flip a switch based on something like CCS. The [Suite: operators]...operators example would for sure be redundant, but maybe having the labels to show the configurations the test is taking in might be useful? This set of labels could also get recorded somewhere else. I'm not sure on your full image yet.

@MrSantamaria Thank you for these points.

One of the main reasons I was suggesting to remove labels from being appended to the test case names is to eliminate the test case names changing every time a test case's labels are altered. If a test case is altered a variety of times being reclassified, it will affect the test case names when viewing results when running locally, in a prow job and results in test grid. Below is an example showing labels being altered, these would represent test case names in the JUnit report file:

```
# Initial label to existing test
[It] [Suite: operators] [OSD] Prune jobs pruner jobs should works builds-pruner should run successfully [operators]"

# Additional labels added after updated classification
[It] [Suite: operators] [OSD] Prune jobs pruner jobs should works builds-pruner should run successfully [operators, sts, rosa]"

# Updated labels to support hypershift
[It] [Suite: operators] [OSD] Prune jobs pruner jobs should works builds-pruner should run successfully [operators, sts, rosa, hypershift]"
```

Of course this test case name has the redundant text in it but eventually all test cases will want to be revised.

We also will know what type of tests are being run when reviewing the [periodic job names in Prow](https://prow.ci.openshift.org/?repo=openshift%2Fosde2e&type=periodic). From there each job we would be able to see the tests that were ran based on the job configuration.

The label filter will also show directly within the JUnit report as properties for the test suite based on what was provided to run:

```xml
<testsuites tests="237" disabled="229" errors="0" failures="0" time="36.547932581">
<testsuite name="OSD e2e suite" package="/home/rywillia/osd/osde2e" tests="237" disabled="0" skipped="229" errors="0" failures="0" time="36.547932581" timestamp="2022-12-07T13:54:59">
<properties>
<property name="SuiteSucceeded" value="true"/>
<property name="SuiteHasProgrammaticFocus" value="false"/>
<property name="SpecialSuiteFailureReason" value=""/>
<property name="SuiteLabels" value="[]"/>
<property name="RandomSeed" value="1670439291"/>
<property name="RandomizeAllSpecs" value="false"/>
>>>> <property name="LabelFilter" value="Operators"/>
>>>> <property name="FocusStrings" value="Must Gather"/>
```

MrSantamaria:
Didn't even think about the actual test names being different in prow/testgrid! Thanks for clearing that up. 
lgtm!

PR #1446, opened 2022-12-06
e2e/webhooks: refactor Managed Cluster Validating Webhooks tests

jbpratt:
Refactor the Managed Cluster Validating Webhooks tests as a proof of
concept for applying a consistent set of standards, heavily utilizing
Ginkgo, Gomega, and the e2e-framework, to each test to increase
maintainability and ease the process of expanding feature coverage by
better understanding what currently exists. This consolidates each
tested webhook into a single suite, covering the following webhooks:

- pod
- regularuser
- techpreviewnoupgrade
- hiveownership
- scc (new)
- namespace
- prometheusrule

and install tests. Two webhooks are not covered, `clusterlogging` which
is not deployed to >4.6 clusters ~~and `prometheusrule` which currently
has tests in development.~~

Many tests use varying level of Ginkgo functionality like
`BeforeEach`/`AfterEach`, some falling back to `defer` or varying
methods of creating and waiting for resources rather than a consistent
set of standard helpers/layout. While each test is different, a common
pattern can be devised to provide consistentcy across the tests. Having
this common pattern applied across tests makes it easy to work with
existing tests and add new tests.

Test case naming has been improved to provide a readable description of
the test. `Describe` and nested `It` clauses are combined to build the
name of the test used in the final report which can lead to poorly named
tests if the author is not aware. More information on test naming can
come from sig-testing's [document on writing good e2e
tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-testing/writing-good-e2e-tests.md).

Within the tests, the kubernetes-sigs/e2e-framework is proposed to be
used for interacting with the cluster. It provides a client ("klient")
exposing a controller runtime client that schemas can be registered
with, allowing us to dynamically interact with objects using similar
crud operations. On top of the client, there are many helpers developed
that meet many of the needs within the tests. This allows osde2e to be
focused more on Ginkgo and a thin layer of test helpers. Opportunities
can be taken to contribute functionality developed back to the
e2e-framework.

A handful of tests have been removed due to being duplicated or
consolidated into a single test such as the install tests. A few new
tests have been added to cover missing features/branches within the
webhooks such as the SCC test to validate modifications made.

Signed-off-by: Brady Pratt <bpratt@redhat.com>
Co-authored-by: Ryan Williams <3375653+ryankwilliams@users.noreply.github.com>

for a list of tests added and removed:

<details>

before (24):

```
[Suite: service-definition] [OSD] NodeLabels Modifying nodeLabels is not allowed node-label cannot be added
[...]

PR #1453, opened 2022-12-08
Introduce the service definition label to the existing tests

ryankwilliams:
This change is amending to the existing tests describe nodes. Which allows the suite of tests to be executed using ginkgos label filter feature over using the focus filter.

Focus filter will continue to work as the string Suite: service-definition still exists within the test case names. It will eventually be removed in replacement of the label identifiers.

Closes [SDCICD-861](https://issues.redhat.com//browse/SDCICD-861)

### DIFF
--- a/pkg/common/label/label.go
+++ b/pkg/common/label/label.go
@@ -67,4 +67,8 @@ var (
 
 	// Operator tests supported on all cluster types
 	Operators = ginkgo.Label("Operators")
+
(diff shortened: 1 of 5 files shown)


MrSantamaria:
Looks like file: regularuser_webhook.go
Will be removed as part of this other PR:
https://github.com/openshift/osde2e/pull/1446

PR #1478, opened 2022-12-21
sdcicd-799 utilize the ensurePassingXML functionality in operator suite

ritmun:
This change will use the newer RetrieveTestResults function instead of the older RetrieveResults function which did not utilize ensurePassingXML which validates junit xml. 

This function is used in some newer suites but not retrofitted into existing ones, such as this (Operators) one. We should probably do that. 

jira https://issues.redhat.com/browse/SDCICD-799


pkg/e2e/operators/operators.go
```diff
@@ -391,7 +391,7 @@ Loop:
 			return err
 		default:
 			if elapsed < timeoutDuration {
-				log.Printf("Waiting %v for %s clusterRoleBinding to exist", (timeoutDuration - elapsed), clusterRoleBindingName)
+				log.Printf("Waiting %v for %s clusterRoleBinding to exist", timeoutDuration-elapsed, clusterRoleBindingName)
```
MrSantamaria, line 394: Could we change this back to having the space? 
Removing the space between variables makes it hard to read, I thought it was one word.

pkg/e2e/operators/operators.go
```diff
@@ -432,7 +432,7 @@ Loop:
 			return err
 		default:
 			if elapsed < timeoutDuration {
-				log.Printf("Waiting %v for %s roleBinding to exist", (timeoutDuration - elapsed), roleBindingName)
+				log.Printf("Waiting %v for %s roleBinding to exist", timeoutDuration-elapsed, roleBindingName)
```
MrSantamaria, line 435: Could we change this back to having the space? 
Removing the space between variables makes it hard to read, I thought it was one word.

pkg/e2e/operators/operators.go
```diff
@@ -474,7 +474,7 @@ Loop:
 			return err
 		default:
 			if elapsed < timeoutDuration {
-				log.Printf("Waiting %v for %s configMap to exist", (timeoutDuration - elapsed), operatorLockFile)
+				log.Printf("Waiting %v for %s configMap to exist", timeoutDuration-elapsed, operatorLockFile)
```
MrSantamaria, line 477: Could we change this back to having the space? 
Removing the space between variables makes it hard to read, I thought it was one word.

pkg/e2e/operators/operators.go
```diff
@@ -517,7 +517,7 @@ Loop:
 			return nil, err
 		default:
 			if elapsed < timeoutDuration {
-				log.Printf("Waiting %v for %s deployment to exist", (timeoutDuration - elapsed), deploymentName)
+				log.Printf("Waiting %v for %s deployment to exist", timeoutDuration-elapsed, deploymentName)
```
MrSantamaria, line 520: Could we change this back to having the space? 
Removing the space between variables makes it hard to read, I thought it was one word.

MrSantamaria (changes requested):

pkg/e2e/operators/operators.go
ritmun: My IDE's auto reformat must've done it - will do

MrSantamaria:
I also don't think we need this anymore. 
From what I validated locally the junit_*.xml is being outputted correctly. 
Refer to this [thread](https://redhat-internal.slack.com/archives/CMK13BP4J/p1673289666907439?thread_ts=1673287242.860939&cid=CMK13BP4J)

jbpratt:
/close

based on https://github.com/openshift/osde2e/pull/1478#issuecomment-1379517516 , please feel free to reopen if this is incorrect!

PR #1484, opened 2022-12-29
SDCICD-874 rework querying of operator catalogs

mrbarge:
# What is this change?

Reworks the `should upgrade from previous version` operator CSV upgrade tests to no longer perform a catalog GRPC query via a created pod, and instead do everything in-code.

# Why?

The `should upgrade from previous version` tests work by:
- determining what the N-1 installed ClusterServiceVersion of an operator should be, based on its catalog
- remove the current version of the operator and install version N-1
- wait and expect OLM to upgrade the operator back to version N successfully

To do step 1 above involves querying the operator catalog service, which is grpc-based. To do so, osde2e  would create a pod in the cluster that ran an image (which was an image in someone's personal quay repo), then interpret the log output from the pod.

Whilst it works, nothing above is really anything remotely ideal.

This rework instead swaps out that function with a new one that does the following:
- creates and exposes a service for doing GRPC (/ HTTP/2) querying of the catalog service
- performs the GRPC query in code to get the needed data
- cleans up the service/route afterwards

As a result:
:heavy_check_mark:  no more running unverified personal images!
:heavy_check_mark:  no more spinning up lots of different pods!
:heavy_check_mark:  should be faster!

Refs [SDCICD-874](https://issues.redhat.com//browse/SDCICD-874)


jbpratt:
/hold

looks awesome, want to take a closer look when I get back

go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	github.com/kyleconroy/sqlc v1.16.0
 	github.com/kylelemons/godebug v1.1.0
 	github.com/lib/pq v1.10.7
+	github.com/markbates/pkger v0.15.1
```
jbpratt, line 25: `pkger` has been removed, this may be because of the `pkged.go` file still be locally?

mrbarge: Ah good catch, thank you, that was exactly the issue. Addressed now!

PR #1500, opened 2023-01-10
add terraform for BYOVPC

jbpratt:
This is ripped straight from hackday. Eventually we can hook this up to
run on cluster creation/deletion to allow more dynamic hypershift
cluster provisioning. If the regions validation falls out of date, we can
remove it and depend on the input provided by osde2e to be correct/validated
ahead of time against the api. It'd probably be worth adding additional
variables to allow further customization but this works for now.

thanks @wgordon17 :)

### DIFF
--- a/assets/terraform/setup-vpc.tf
+++ b/assets/terraform/setup-vpc.tf
@@ -1,0 +1,66 @@
+variable "aws_region" {
+  type        = string
+  description = "The region to create the ROSA cluster in"
+  default     = "us-east-1"
+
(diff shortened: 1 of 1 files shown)


wgordon17:
Hey all 👋 Just to give you a heads up on a change to my TF script so that we're targeting AZ _IDs_ instead of _names_ (more reliable for now) https://gist.github.com/wgordon17/d88501a15204a2f8143bc275a3f64c80

PR #1504, opened 2023-01-11
SDCICD-889: e2e/osd: enable daemonsets on hypershift

jbpratt:
HyperShift only has worker nodes available so we need to skip the infra role test.

This change refactors the test to use the e2e-framework and fixes the behavior to properly schedule onto the correct node based on label. Based on the documentation[1], `nodeSelector.role: worker` should allow for scheduling a daemonset onto only worker nodes, but the node does not actually have a `role: "worker"` label so nothing gets scheduled.

Taking a look at the node labels, there is the standard kubernetes role label but not the simple `role` label:

```
$ oc get nodes ip-10-0-224-45.ec2.internal -o json | jq .metadata.labels
{
  "beta.kubernetes.io/arch": "amd64",
  "beta.kubernetes.io/instance-type": "m5.xlarge",
  "beta.kubernetes.io/os": "linux",
  "failure-domain.beta.kubernetes.io/region": "us-east-1",
  "failure-domain.beta.kubernetes.io/zone": "us-east-1a",
  "kubernetes.io/arch": "amd64",
  "kubernetes.io/hostname": "ip-10-0-224-45.ec2.internal",
  "kubernetes.io/os": "linux",
  "node-role.kubernetes.io/worker": "",
  "node.kubernetes.io/instance-type": "m5.xlarge",
  "node.openshift.io/os_id": "rhcos",
  "topology.ebs.csi.aws.com/zone": "us-east-1a",
  "topology.kubernetes.io/region": "us-east-1",
  "topology.kubernetes.io/zone": "us-east-1a"
}
```

[1]: https://docs.openshift.com/dedicated/osd_architecture/osd_policy/osd-service-definition.html#daemon-sets_osd-service-definition

### DIFF
--- a/pkg/e2e/osd/daemonsets.go
+++ b/pkg/e2e/osd/daemonsets.go
@@ -5,16 +5,17 @@ import (
 	"fmt"
 
 	"github.com/onsi/ginkgo/v2"
-	. "github.com/onsi/gomega"
 	"github.com/openshift/osde2e/pkg/common/alert"
 	viper "github.com/openshift/osde2e/pkg/common/concurrentviper"
 	"github.com/openshift/osde2e/pkg/common/config"
+	"github.com/openshift/osde2e/pkg/common/expect"
 	"github.com/openshift/osde2e/pkg/common/helper"
 	"github.com/openshift/osde2e/pkg/common/label"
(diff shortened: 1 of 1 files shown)


ryankwilliams:
Verified changes for both Rosa osd + hypershift locally.

PR #1520, opened 2023-01-16
SDCICD-848 - Fix OSD CCS Cluster Creation

MrSantamaria:
https://issues.redhat.com/browse/SDCICD-848

### DIFF
--- a/pkg/common/providers/ocmprovider/cluster.go
+++ b/pkg/common/providers/ocmprovider/cluster.go
@@ -159,46 +159,43 @@ func (o *OCMProvider) LaunchCluster(clusterName string) (string, error) {
 		}
 
 		if viper.GetString(config.CloudProvider.CloudProviderID) == "aws" && awsAccount != "" && awsAccessKey != "" && awsSecretKey != "" {
+			awsBuilder := v1.NewAWS().
+				AccountID(awsAccount).
+				AccessKeyID(awsAccessKey).
+				SecretAccessKey(awsSecretKey)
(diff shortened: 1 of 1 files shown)


pkg/common/providers/ocmprovider/cluster.go
```diff
@@ -159,46 +159,43 @@ func (o *OCMProvider) LaunchCluster(clusterName string) (string, error) {
...
-					}
-					if httpsProxy := viper.GetString(config.Proxy.HttpsProxy); httpsProxy != "" {
-						proxy = proxy.HTTPSProxy(httpsProxy)
-						newCluster = newCluster.Proxy(proxy)
-					}
+				if httpProxy := viper.GetString(config.Proxy.HttpProxy); httpProxy != "" {
+					proxy = proxy.HTTPProxy(httpProxy)
+					newCluster = newCluster.Proxy(proxy)
+				}
+				if httpsProxy := viper.GetString(config.Proxy.HttpsProxy); httpsProxy != "" {
```
dustman9000, line 196: I think we would only want the httpsProxy if there is a CA bundle, right?

MrSantamaria: The PR moved it out of an if block but that http proxy logic is untouched. 
It does check for the CA Bundle over here:
https://github.com/openshift/osde2e/pull/1520/files#diff-360361bd516bd9b8a5a85b5a4ce53db8fa5e5682353cf4791c44a6aaeb2f7eadR185

PR #1526, opened 2023-01-18
e2e/verify: wait for deployment to be created

jbpratt:
the test keeps failing because the deployment isn't created yet but it gets created within 30 seconds of the test failing. Let's wait for it to exist and retry the GET request as it should _eventually_ succeed.

### DIFF
--- a/pkg/e2e/verify/routes.go
+++ b/pkg/e2e/verify/routes.go
@@ -5,7 +5,6 @@ import (
 	"crypto/tls"
 	"fmt"
 	"net/http"
-	"time"
 
 	"github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


MrSantamaria:
lgtm but I'm confused on one aspect. When removing the 3 minute timeout what becomes the timeout? 
I believe the wait package has a 2 min default timeout,  and the eventually is every second(I think) 
So is this just checking every second for the default 2 minutes?

jbpratt:
> lgtm but I'm confused on one aspect. When removing the 3 minute timeout what becomes the timeout? I believe the wait package has a 2 min default timeout, and the eventually is every second(I think) So is this just checking every second for the default 2 minutes?

I'm just falling back to the default timeouts https://github.com/kubernetes-sigs/e2e-framework/blob/main/klient/wait/wait.go#L25-L28

We have a big problem of setting timeouts all over the place without rhyme or reason so I'm just trying to minimize the arbitrary timeouts

MrSantamaria:
> I'm just falling back to the default timeouts https://github.com/kubernetes-sigs/e2e-framework/blob/main/klient/wait/wait.go#L25-L28
>
Oh it is 5 minutes, so it is an increase. 

> We have a big problem of setting timeouts all over the place without rhyme or reason so I'm just trying to minimize the arbitrary timeouts
> 

I wouldn't consider the different timeouts a problem. 
Personally I would say that the root cause for scattered timeouts are a lack of defined SLOs per test. 
That would be a great place for QE to define timeout strategies across Openshift. For a test like this it starts with giving it an actual owner. 

This would be a great topic of discussion for our standup. 
In the interim lgtm! Thanks for the explanation.

PR #1530, opened 2023-01-20
SDCICD-915: support fips through rosaprovider

jbpratt:
fips is supported on ROSA and should be tested. This adds support for deploying fips enabled clusters through osde2e

### DIFF
--- a/pkg/common/providers/rosaprovider/cluster.go
+++ b/pkg/common/providers/rosaprovider/cluster.go
@@ -144,6 +144,10 @@ func (m *ROSAProvider) LaunchCluster(clusterName string) (string, error) {
 		}
 	}
 
+	if viper.GetBool(config.Tests.EnableFips) {
+		createClusterArgs = append(createClusterArgs, "--fips")
+	}
+
 	err = callAndSetAWSSession(func() error {
 		// Retrieve AWS Account info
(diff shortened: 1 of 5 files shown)


pkg/common/providers/rosaprovider/cluster.go
```diff
@@ -144,6 +144,10 @@ func (m *ROSAProvider) LaunchCluster(clusterName string) (string, error) {
 		}
 	}
 
+	if viper.GetBool(config.Tests.EnableFips) {
```
ryankwilliams, line 147: I know this config setting was already [present](https://github.com/openshift/osde2e/blob/main/pkg/common/config/config.go#L246) under test configuration. Do we think this should be a cluster specific setting under [cluster config](https://github.com/openshift/osde2e/blob/main/pkg/common/config/config.go#L265)?

Looks like this config setting is only used within the `pkg/e2e/verify/fips.go` test and there are 2 configs (fedramp/fips).

This change looks good to me :+1:, just wanted to mention this.

jbpratt: updated in b558e8a

ryankwilliams: Thanks for making the change @jbpratt!

PR #1573, opened 2023-02-08
providers/rosa: use region set in config file

jbpratt:
This causes an error if trying to use govcloud because it is forcing us-east-1 instead of what is specified by the user


jbpratt:
/assign @MrSantamaria 

cc @tonytheleg @dustman9000

dustman9000:
thanks!
/lgtm

PR #1558, opened 2023-01-31
Add support for auto creating HyperShift cluster VPC

ryankwilliams:
# Change

This PR introduces support for auto creating an AWS VPC for HyperShift clusters. This feature is disabled by default to maintain the current model of BYOVPC. Where you provide the public and private subnet ids for your VPC. VPC operations are performed using terraform. Introduces the [terraform-exec](https://github.com/hashicorp/terraform-exec) go module to osde2e.

When the config setting is enabled for auto creating VPC subnet ids, osde2e will do the following:

1. Call terraform to create VPC and fetch subnet ids
2. Pass subnet ids into rosa create cluster command
3. Pass a new cluster property to rosa create cluster command to save the location where the terraform state file exists (used for later on to delete created VPC by terraform)

When osde2e goes to destroy the cluster, the following is performed:

1. Get the cluster property to know where on the file system it should look for the terraform state file (this is really needed when you run osde2e multiple times with destroyAfterTest set to false. Then on your final run you set it to true and need to know where the state file is.
2. Delete the cluster
3. Delete STS roles
4. Delete the ELB security group (workaround for open hosted control plane bug)
5. Call terraform to delete auto created VPC

# Example

Run osde2e using custom config below to deploy HyperShift cluster with auto created VPC.

```
dryRun: false
provider: rosa
mustGather: false
Hypershift: True
config:
  aws:
    createVPCSubnetIDs: true <<<<<<<<
    region: us-east-1
rosa:
  env: stage
  STS: true
cluster:
  hibernateAfterUse: false
  cleanCheckRuns: "3"
  name: rw-java
  version: 4.12.0
cloudProvider:
  providerId: aws
  region: us-east-1
```

Output

```
<compressed>
2023/01/30 17:01:48 vpc.go:53: Creating ROSA HyperShift aws vpc
2023/01/30 17:04:08 vpc.go:92: ROSA HyperShift aws vpc created!
2023/01/30 17:04:08 cluster.go:93: AWS VPC created at runtime, subnet ids: subnet-030c0eac66f95c93c,subnet-0f14a4a997259b0fa
2023/01/30 17:04:08 cluster.go:102: ROSA cluster version: 4.12.0
2023/01/30 17:04:08 cluster.go:231: Pruning `--compute-machine-type` and ``
2023/01/30 17:04:08 cluster.go:231: Pruning `--machine-cidr` and ``
2023/01/30 17:04:08 cluster.go:231: Pruning `--service-cidr` and ``
[...]

PR #1610, opened 2023-02-22
SDCICD-933: Auto create ROSA hosted control plane VPC when undefined

ryankwilliams:
# Change

This PR adds support to OSDE2E to auto create the VPC for ROSA hosted control plane clusters over requiring the user to BYOVPC.

When a user calls OSDE2E with the options for a HyperShift cluster and provides the VPC subnet ids, it will not create any VPC and use the one provided.

```
provider: rosa
Hypershift: True
config:
  aws:
    vpcSubnetIDs: "subnet-1,subnet-2"
    region: us-east-1
rosa:
  env: stage
  STS: true
cluster:
  name: hcp
  version: 4.12.4
```

When a user calls OSDE2E with the options for a HyperShift cluster and does not provided any VPC subnet ids, it will automatically create the VPC and destroy the VPC when finished.

```
provider: rosa
Hypershift: True
config:
  aws:
    region: us-east-1
rosa:
  env: stage
  STS: true
cluster:
  name: hcp
  version: 4.12.4
```

# Implementation Details

When OSDE2E goes to create the VPC it will:
1. Call terraform to create VPC and fetch subnet ids
2. Pass subnet ids into rosa create cluster command
3. Pass a new cluster property to rosa create cluster command to save the location where the terraform state file exists (used for later on to delete created VPC by terraform)

When OSDE2E goes to delete the VPC it will:
1. Get the cluster property to know where on the file system it should look for the terraform state file (this is really needed when you run osde2e multiple times with destroyAfterTest set to false. Then on your final run you set it to true and need to know where the state file is.
2. Delete the cluster
3. Delete STS roles
4. Call terraform to delete auto created VPC

# Status
* In order for this to work, Jira issue **HOSTEDCP-789** has to be done and available in newer OCP version. Without this, VPC deletion will fail due to ELB security group still existing.

For [SDCICD-933](https://issues.redhat.com/browse/SDCICD-933)

Triggered by #1558. That PR was the first implementation which was later own broken down into multiple PR's to introduce the functionality.

### DIFF
--- a/assets/terraform/setup-vpc.tf
+++ b/assets/terraform/setup-vpc.tf
@@ -1,30 +1,23 @@
 variable "aws_region" {
   type        = string
   description = "The region to create the ROSA cluster in"
-  default     = "us-east-1"
-
-  validation {
-    condition     = contains(["us-east-1", "eu-west-1"], var.aws_region)
-    error_message = "HyperShift is currently only availble in these regions: us-east-1, eu-west-1."
-  }
 }
 
 variable "az_ids" {
   type = object({
-    us-east-1 = list(string)
-    eu-west-1 = list(string)
+    us-east-2 = list(string)
+    us-west-2 = list(string)
   })
   description = "A list of region-mapped AZ IDs that a subnet should get deployed into"
   default = {
-    us-east-1 = ["use1-az1", "use1-az2"]
-    eu-west-1 = ["euw1-az1", "euw1-az2"]
+    us-east-2 = ["use2-az1", "use2-az2"]
+    us-west-2 = ["usw2-az1", "usw2-az2"]
   }
 }
 
 variable "cluster_name" {
   type        = string
   description = "The name of the ROSA cluster to create"
-  default     = "rosa-cluster"
 
   validation {
     condition     = can(regex("^[a-z][-a-z0-9]{0,13}[a-z0-9]$", var.cluster_name))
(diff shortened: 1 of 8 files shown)


ryankwilliams:
@jbpratt Regarding your comment https://github.com/openshift/osde2e/pull/1558#pullrequestreview-1277047090 see my responses below:

1. For logging, I did not reuse the existing logger because the files I was changing was using the import `log` package. I personally feel separate PR's should be opened to streamline/standardize on the logging. Let me know if that is okay, else I can adjust them in the files I modified.
2. For saving the state of where the terraform state file, that is a good idea to store in an s3 bucket over using cluster properties. My concern here is then, we would probably need to have a static s3 bucket out there where we can write to. It would also introduce needing to set up some policies to control who can write to it if using a different aws account. Though it would then be possible to push the terraform state file to the bucket in the event the state file was ever deleted by accident. Would just require more to think of, where the cluster properties would require less code to implement saving the state. Let me know your thoughts/how you want me to proceed.

ryankwilliams:
@jbpratt @MrSantamaria Regarding your comments https://github.com/openshift/osde2e/pull/1558#discussion_r1091843599, see my response below:

In this PR I changed when it would create the VPC. When user enables HyperShift setting and does not pass subnet ids, it will create it. Where if the subnet ids are passed, it would use those/not create the vpc.

I do like the idea of introducing an auto flag. If we went this approach, we just will want to brainstorm/define what exact operations are to be performed on the cloud provider when `--auto` is enabled and the alternative when its not passed.

PR #1624, opened 2023-02-27
common/cluster: catch x509 errors in healthcheck

jbpratt:
it seems that the certificate is not signed by the time the cluster comes up so the healthcheck is failing

https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-osde2e-main-hypershift-stage-e2e-default/1630176089026334720


jbpratt:
maybe this isn't the right solution if the cert never becomes valid..

PR #1630, opened 2023-03-01
Draft: [OSD-15129] Adding FedRAMP to osde2e

theautoroboto:


pkg/common/providers/ocmprovider/ocm.go
```diff
@@ -15,17 +16,21 @@ const (
...
-	TokenURL = "https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token"
+	// commercialTokenURL specifies the endpoint used to create access tokens in the commercial environment.
+	commercialTokenURL     = "https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token"
+	// The following TokenURLs are specific to the FedRAMP environment
+	govintegrationTokenURL = "https://rh-ocm-appsre-integration.auth-fips.us-gov-west-1.amazoncognito.com/oauth2/token"
+	govstageTokenURL       = "https://ocm-ra-stage-domain.auth-fips.us-gov-west-1.amazoncognito.com/oauth2/token"
+	govproductionTokenURL  = "https://ocm-ra-production-domain.auth-fips.us-gov-west-1.amazoncognito.com/oauth2/token"
 
 	// ClientID is used to identify the client to OSD.
 	ClientID = "cloud-services"
```
vkareh, line 27: For govcloud the client ID needs to be different for each environment: https://github.com/openshift/rosa/blob/v1.2.15/pkg/fedramp/config.go#L65-L70

PR #1665, opened 2023-03-14
Comment for hypershift config regarding aws region

MrSantamaria:


configs/hypershift.yaml
```diff
@@ -4,4 +4,5 @@ rosa:
   Env: stage
   STS: True
 tests:
-  ginkgoLabelFilter: HyperShift
\ No newline at end of file
+  ginkgoLabelFilter: HyperShift
+# aws-region: us-west-2 - Currently this is the only supported region.
```
ryankwilliams, line 8:
```suggestion
# aws-region: us-west-2 - Check for supported regions by using `rosa list regions````

PR #1669, opened 2023-03-14
Convert Addon Harness to Test Harness - First Pass

MrSantamaria:
[SDCICD-717](https://issues.redhat.com/browse/SDCICD-717)

This PR refactors the addon suite to be the new test harness suite. 
For this pass I didn't deprecate all the functionality that will be not be carried over.


pkg/e2e/harness_runner.go/harness_runner.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+var _ = ginkgo.Describe("[Suite: Tests] Test Harness", func() {
+	defer ginkgo.GinkgoRecover()
+	h := helper.New()
+
+	TimeoutInSeconds := float64(viper.GetFloat64(config.Tests.PollingTimeout))
+	util.GinkgoIt("should run until completion", func(ctx context.Context) {
+		log.Printf("Test harness timeout is %v", TimeoutInSeconds)
+		h.SetServiceAccount(ctx, viper.GetString(config.Tests.TestUser))
```
MrSantamaria, line 26: @jbpratt glad you are looking at this PR. In the old addon implementation there is some unused features. 
They had implemented cleanup logic and this variable which could be used inside the test harness. 
Do we want to port all of those features though they aren't being used or leverage some of the newer ones being set up such as the impersonator funcs yall have created?

MrSantamaria:
Our PR checks do not test for this change. 
If someone wants to try this branch out I'm using:
`            "env": {"ROSA_ENV": "stage",
                "SKIP_DESTROY_CLUSTER": "true",
                "GINKGO_SKIP":"" ,
                "MUST_GATHER": "false",
                "CLUSTER_ID": "",
                "HIBERNATE_AFTER_USE": "false",
                "ADDON_IDS": "reference-addon",
                "TEST_HARNESSES": "quay.io/rmundhe_oc/osde2e-example-test-harness,quay.io/rmundhe_oc/osde2e-example-test-harness",
            },
            "args": ["test",
                "--configs",
                "rosa,test-harness",`
                
These configs match this [example-addon](https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/osde2e/openshift-osde2e-main-periodics.yaml#L274-L342)

pkg/e2e/harness_runner/harness_runner.go
```diff
@@ -0,0 +1,38 @@
...
+	viper "github.com/openshift/osde2e/pkg/common/concurrentviper"
+
+	"github.com/openshift/osde2e/pkg/common/alert"
+	"github.com/openshift/osde2e/pkg/common/config"
+	"github.com/openshift/osde2e/pkg/common/helper"
+	"github.com/openshift/osde2e/pkg/common/prow"
+)
+
+var _ = ginkgo.Describe("[Suite: harnessess] Test Harness", func() {
+	defer ginkgo.GinkgoRecover()
```
jbpratt, line 19: https://pkg.go.dev/github.com/onsi/ginkgo/v2#GinkgoRecover not sure if this should still be there

pkg/e2e/harness_runner/harness_runner.go
```diff
@@ -0,0 +1,38 @@
...
+	h := helper.New()
```
jbpratt, line 20: helpers need to be created in `BeforeAll` or `BeforeEach` to prevent it from being created before Ginkgo actually runs the test

pkg/e2e/harness_runner/harness_runner.go
```diff
@@ -0,0 +1,38 @@
...
+
+	TimeoutInSeconds := float64(viper.GetFloat64(config.Tests.PollingTimeout))
+	ginkgo.It("should run until completion", func(ctx context.Context) {
+		log.Printf("Test harness timeout is %v", TimeoutInSeconds)
```
jbpratt, line 24: prefer `GinkgoWriter` to stdlib logger so this doesn't get printed outside of the test context

pkg/e2e/harness_runner/harness_runner.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/openshift/osde2e/pkg/common/config"
+	"github.com/openshift/osde2e/pkg/common/helper"
+	"github.com/openshift/osde2e/pkg/common/prow"
+)
+
+var _ = ginkgo.Describe("[Suite: harnessess] Test Harness", func() {
+	defer ginkgo.GinkgoRecover()
+	h := helper.New()
+
+	TimeoutInSeconds := float64(viper.GetFloat64(config.Tests.PollingTimeout))
```
jbpratt, line 22: nit: is there a reason we use float64 for polling timeout? It is commonly used all over osde2e but then below it is just converted to an int anyway

pkg/e2e/harness_runner/harness_runner.go
```diff
@@ -0,0 +1,38 @@
...
+		failed := h.RunTests(ctx, "test-harness", int(TimeoutInSeconds), harnesses, []string{})
+		if len(failed) > 0 {
+			message := fmt.Sprintf("Tests failed: %v", failed)
+			if url, ok := prow.JobURL(); ok {
+				message += "\n" + url
+			}
+			if err := alert.SendSlackMessage(viper.GetString(config.Tests.SlackChannel), message); err != nil {
+				log.Printf("Failed sending slack alert for test failure: %v", err)
+			}
+		}
```
jbpratt, line 36: Is there a reason we rely on this slack messaging mechanic versus the one we do in other files versus prow?

pkg/e2e/harness_runner/harness_runner.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/onsi/ginkgo/v2"
+	viper "github.com/openshift/osde2e/pkg/common/concurrentviper"
+
```
jbpratt, line 11:
```suggestion
```

jbpratt (changes requested):

pkg/e2e/harness_runner/harness_runner.go
MrSantamaria: Removed

pkg/e2e/harness_runner/harness_runner.go
MrSantamaria: Removed

pkg/e2e/harness_runner/harness_runner.go
MrSantamaria: Add the helper to a BeforeAll

pkg/e2e/harness_runner/harness_runner.go
MrSantamaria: Changed to ginkgo.GinkgoWriter.Write()
Though I'm not seeing this change in other test.

pkg/e2e/harness_runner/harness_runner.go
MrSantamaria: No actual reason here, there isn't any standard across the tests. 
The reason it turns into an int is just for the helper RunTests function.

jbpratt:
@ryankwilliams do we just want to skip this job and merge the PR? I don't think this job is even running the test and the job failed due to an uninstall bug

PR #1678, opened 2023-03-17
Exit with failure if version cannot be chosen

bparees:
not sure why the expectation would be a graceful exit, but it causes
false negatives like this in job runs:

https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-ocp-osd-aws-nightly-4.12/1636143394818363392

in which the job effectively failed (did not run), but reports success.

There is some weird history on this change. It was originally introduced deliberately(tor eturn success) here:
https://github.com/openshift/osde2e/commit/44503b8cc4b890eb9f756c83380aee0796d45a56

But the "reason" points to this jira ticket:
https://issues.redhat.com/browse/SDCICD-639

which contains no further context beyond pointing to this slack thread:
https://coreos.slack.com/archives/CCX9DB894/p1618515431158100

which appears to be deleted.

Hopefully another maintainer on this repo can provide context on why the change was made and if this revert isn't appropriate maybe we can discuss a better solution to the problem at hand.

### DIFF
--- a/pkg/e2e/e2e.go
+++ b/pkg/e2e/e2e.go
@@ -357,8 +357,8 @@ func runGinkgoTests() (int, error) {
 
 		// configure cluster and upgrade versions
 		if err = ChooseVersions(); err != nil {
-			// if we fail to choose versions, we should gracefully exit
-			return Success, err
+			// If we can't find a version to use, exit with an error code.
+			return Failure, err
 		}
 
 		switch {


bparees:
@ryankwilliams ptal or delegate as needed

MrSantamaria:
@bparees Before my time as well but some of the decisions for those passes are due to the old ci watcher role. 
Basically if a run wasn't providing an actual failure it was treated as a success in order to not generate alerting signal. 

This decision comes from Prow/Testgrid only supporting pass/fail status. If we switch the version to fail we would send an alert to pipeline owners that their pipeline failed, when in reality it was an inconclusive run. 
I do agree the pipeline shouldn't just pass in this case but the current jobs aren't gating just informing. As we swap over to gating though this needs to change. 

IMPO: We shouldn't cause the pipeline to fail if the version fails, but we could take an action on failing to set a version, to a channel such as #\sd-cicd-alerts.

bparees:
>  If we switch the version to fail we would send an alert to pipeline owners that their pipeline failed, when in reality it was an inconclusive run.

I sympathize with that concern, but this behavior means you risk getting into a situation where no jobs are actually running to completion, they are all reporting success, and no one notices unless they actually go look at the job output.  (e.g. some sort of config or infrastructure change takes this from a "occasionally times out, we don't care" situation to "fails in this way every single time")

our general approach to this problem in OCP has been to stop caring about individual job results and start caring about the overall pass rate trends (either the job pass rate, or individual test pass rates).  You can then alert when thresholds are breached.

Of course that doesn't alert you the moment a particular payload is problematic, but given that this job is as red as it is, i get the impression no one is taking action on the existing alert today anyway?  (see https://prow.ci.openshift.org/job-history/gs/origin-ci-test/logs/release-openshift-ocp-osd-aws-nightly-4.12 in which the job has not passed reliably since around march 6th, and even then it was sporadic so i don't know how much noise these additional failures would really be contributing)

MrSantamaria:
> our general approach to this problem in OCP has been to stop caring about individual job results and start caring about the overall pass rate trends (either the job pass rate, or individual test pass rates). You can then alert when thresholds are breached.

Yes this is a problem. I missed the TRT sync meeting yesterday but looking at:
https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/release/openshift-release-release-4.12-periodics.yaml#L929
I would say the first step should be adding the reporter config to point over to the channel. Looking at the failures for that job they are operator test related. 
```
  reporter_config:
    slack:
      channel: '#sd-cicd-alerts'
      job_states_to_report:
      - failure
      - success
      - error
      report_template: '{{if eq .Status.State "success"}} :white_check_mark: Job *{{.Spec.Job}}*
        ended with *{{.Status.State}}*. <{{.Status.URL}}|View logs> :white_check_mark:
        {{else}} :warning:  Job *{{.Spec.Job}}* ended with *{{.Status.State}}*. <{{.Status.URL}}|View
        logs> :warning: {{end}}'
```

I would also prefer a false negative to a false positive so this Pr change /lgtm

bparees:
@MrSantamaria who do i need to talk to in order to push this forward?

jbpratt:
/lgtm

CI will be failing, we are working on getting the hypershift job fixed

PR #1745, opened 2023-04-11
Add support to auto create OIDC config id for ROSA HCP clusters

ryankwilliams:
# Change
When a user calls osde2e for ROSA HCP cluster and does not provide an oidc config id, it will be automatically created/deleted. If the user provides an oidc config id, it will not be deleted by osde2e.

For [SDCICD-965](https://issues.redhat.com/browse/SDCICD-965)

### DIFF
--- a/pkg/common/providers/rosaprovider/oidcconfig.go
+++ b/pkg/common/providers/rosaprovider/oidcconfig.go
@@ -28,7 +28,7 @@ func (m *ROSAProvider) createOIDCConfig(prefix string) (string, error) {
 		return "", fmt.Errorf("error creating odic-config: %v", err)
 	}
 
-	oidcConfigID, err := m.getOIDCConfig(prefix)
+	oidcConfigID, err := m.getOIDCConfigID(prefix)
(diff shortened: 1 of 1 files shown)


pkg/common/providers/rosaprovider/oidcconfig.go
```diff
@@ -0,0 +1,75 @@
...
+	oidcConfigID, err := m.getOIDCConfig(prefix)
+	if err != nil {
+		return "", fmt.Errorf("OIDC config created, %v", err)
+	}
+
+	return oidcConfigID, nil
+}
+
+// getOIDCConfig gets an existing oidc config
+func (m *ROSAProvider) getOIDCConfig(prefix string) (string, error) {
```
jbpratt, line 40:
```suggestion
func (m *ROSAProvider) getOIDCConfigID(prefix string) (string, error) {
```

ryankwilliams: Thanks for catching this!

PR #1771, opened 2023-04-18
SDCICD-977 access AWS account ID through CCS session struct

ritmun:
https://issues.redhat.com/browse/SDCICD-977

Adds AWS account ID as a member of CcsAwsSession. checks for   AWS_ACCOUNT_ID env var . Renames previously "AWS_ACCOUNT" to "AWS_ACCOUNT_ID". Replaces direct calls to viper config from elsewhere in code. 

Secret key name and config name are still aws-account and. config.aws.account respectively. No change there.

### DIFF
--- a/pkg/common/aws/session.go
+++ b/pkg/common/aws/session.go
@@ -3,6 +3,7 @@ package aws
 import (
 	"fmt"
 	"log"
+	"os"
 	"sync"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -15,10 +16,11 @@ import (
 )
 
 type ccsAwsSession struct {
-	session *session.Session
-	iam     *iam.IAM
-	ec2     *ec2.EC2
-	once    sync.Once
+	session   *session.Session
+	accountId string
+	iam       *iam.IAM
+	ec2       *ec2.EC2
+	once      sync.Once
 }
 
 // CCSAWSSession is the global AWS session for interacting with AWS.
@@ -48,6 +50,13 @@ func (CcsAwsSession *ccsAwsSession) GetAWSSessions() error {
 		CcsAwsSession.session, err = session.NewSessionWithOptions(options)
 		CcsAwsSession.iam = iam.New(CcsAwsSession.session)
 		CcsAwsSession.ec2 = ec2.New(CcsAwsSession.session)
+		if viper.GetString(config.AWSAccount) != "" {
+			CcsAwsSession.accountId = viper.GetString(config.AWSAccount)
+		} else if os.Getenv("AWS_ACCOUNT_ID") != "" {
+			CcsAwsSession.accountId = os.Getenv("AWS_ACCOUNT_ID")
+		} else {
+			CcsAwsSession.accountId = ""
+		}
 	})
 	if err != nil {
 		log.Printf("error initializing AWS session: %v", err)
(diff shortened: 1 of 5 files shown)


pkg/common/aws/session.go
```diff
@@ -48,6 +50,13 @@ func (CcsAwsSession *ccsAwsSession) GetAWSSessions() error {
 		CcsAwsSession.session, err = session.NewSessionWithOptions(options)
 		CcsAwsSession.iam = iam.New(CcsAwsSession.session)
 		CcsAwsSession.ec2 = ec2.New(CcsAwsSession.session)
+		if viper.GetString(config.AWSAccount) != "" {
+			CcsAwsSession.accountId = viper.GetString(config.AWSAccount)
+		} else if os.Getenv("AWS_ACCOUNT_ID") != "" {
+			CcsAwsSession.accountId = os.Getenv("AWS_ACCOUNT_ID")
+		} else {
+			CcsAwsSession.accountId = ""
+		}
```
ryankwilliams, line 59:
```suggestion
		CcsAwsSession.accountId = viper.GetString(config.AWSAccount)
```

We should be able to just have one call to viper which will get the aws account id either from custom config input or environment variable. As viper will get the value from environment variable named `AWS_ACCOUNT` [here](https://github.com/openshift/osde2e/blob/main/pkg/common/config/aws_config.go#L29).

If we also want to take into account the environment variable named `AWS_ACCOUNT_ID`, we should just update the list of environment variables viper can get the value from:

```go
viper.BindEnv(AwsAccount, "AWS_ACCOUNT", "AWS_ACCOUNT_ID")
```

If none is set, viper will have the default value be an empty string.

pkg/common/aws/session.go
```diff
@@ -15,10 +16,11 @@ import (
 )
 
 type ccsAwsSession struct {
-	session *session.Session
-	iam     *iam.IAM
-	ec2     *ec2.EC2
-	once    sync.Once
+	session   *session.Session
+	accountId string
```
ryankwilliams, line 20: Should we rename anything account id to be just account or rename account to be account id. Just thinking to be consistent throughout the code. What do you think?

pkg/common/config/aws_config.go
```diff
@@ -6,8 +6,8 @@ import (
 )
 
 var (
-	// AWSAccount is the AWS account
-	AWSAccount = "config.aws.account"
+	// AWSAccountId is the AWS account (Env var: AWS_ACCOUNT_ID)
+	AWSAccountId = "config.aws.account.id"
```
ryankwilliams, line 10:
```suggestion
	AWSAccountId = "config.aws.accountID"
```

If we keep it as `config.aws.account.id`, a custom config would be:

```
config:
  aws:
    account:
      id: my-account-id
```

Where if we switch it to be `config.aws.accountID` a custom config would be:

```
config:
  aws:
    accountID: my-account-id
```

Let me know what you think!

pkg/common/aws/session.go
```diff
@@ -73,3 +75,8 @@ func (CcsAwsSession *ccsAwsSession) GetCredentials() (*credentials.Value, error)
 func (CcsAwsSession *ccsAwsSession) GetRegion() *string {
 	return CcsAwsSession.session.Config.Region
 }
+
+// GetAccount returns the aws account in session
+func (CcsAwsSession *ccsAwsSession) GetAccount() string {
+	return CcsAwsSession.accountId
+}
```
ryankwilliams, line 82:
```suggestion

// GetAccountID returns the aws account id in session
func (CcsAwsSession *ccsAwsSession) GetAccountID() string {
	return CcsAwsSession.accountId
}
```

PR #1818, opened 2023-05-04
sdcicd-955 added new test config

ritmun:
new test config contains same e2e tests  as e2e-suite + test harnesses.

### DIFF
--- a/configs/complete-e2e.yaml
+++ b/configs/complete-e2e.yaml
@@ -1,0 +1,3 @@
+tests:
+  ginkgoLabelFilter: >
+    (E2E || TestHarness || ServiceDefinition || AppBuilds) && !Informing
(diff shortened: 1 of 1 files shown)


configs/complete-e2e.yaml
```diff
@@ -0,0 +1,3 @@
+tests:
+  ginkgoLabelFilter: >
+    (E2E || TestHarness || ServiceDefinition || AppBuilds) && !Informing
```
ryankwilliams, line 3: @ritmun We will still want to include `Operators` in the label filter. As that label is applied to each operator test suite. Since we are looking to run existing operator tests as part of the mono repo.

Once a operator's tests are decentralized, we would need to submit a PR to osde2e to remove the `Operators` label from the test suite. Then the operator tests get run as part of the test harness.

PR #1901, opened 2023-06-07
checking app gets deployed before verifying deployment condition

ritmun:
Workload test fails on servers where deployment condition check runs too quickly after template is deployed. 

DeploymentConditionMatch isn't polling till timeout if deployment doesn't exist by the tiem the function is called. 

Following this K8s example, assessing the deployment resource first, before verifying deployment condition
https://github.com/kubernetes-sigs/e2e-framework/blob/c3b933fe4c7313a31cbc2ea3cad2d6a2a3c8962f/examples/wait_for_resources/wait_test.go#L50


ryankwilliams:
@ritmun Should we close this PR since #2006 was merged to remove the workloads tests from both the HyperShift and E2E suite? We removed it for now to better stabilize/improve the CI signal for TRT nightly jobs and this functionality of OpenShift should be expected to work (refer to the PR description in #2006).

PR #1974, opened 2023-07-13
use describeTable for better running multi harness jobs SDCICD-955

ritmun:
This change organizes test harnesses as describeTable specs, so that when more than one harness is provided, each runs as a separate spec. This can allow us to run them in parallel if needed. 

This also separates each spec into a different namespace and keeps logs from each harness separate from each other

https://issues.redhat.com/browse/SDCICD-955

### DIFF
--- a/pkg/e2e/harness_runner/harness_runner.go
+++ b/pkg/e2e/harness_runner/harness_runner.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"path/filepath"
 	"strings"
 
 	"github.com/onsi/ginkgo/v2"
@@ -29,13 +30,13 @@ var (
 	suffix            string
 )
 
-var _ = ginkgo.Describe("Test Harness", ginkgo.Ordered, ginkgo.ContinueOnFailure, label.TestHarness, func() {
+var _ = ginkgo.Describe("Test harness", ginkgo.Ordered, ginkgo.ContinueOnFailure, label.TestHarness, func() {
 	for _, harness := range harnesses {
 		HarnessEntries = append(HarnessEntries, ginkgo.Entry("should run "+harness+" successfully", harness))
 	}
 
 	ginkgo.BeforeEach(func(ctx context.Context) {
-		// Run harness in a new project
+		ginkgo.By("Setting up new namespace")
 		viper.Set(config.Project, "")
 		h = helper.New()
 		h.SetServiceAccount(ctx, serviceAccount)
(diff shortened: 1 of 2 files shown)


pkg/e2e/harness_runner/harness_runner.go
```diff
@@ -7,34 +7,114 @@ import (
...
+			Expect(err).NotTo(HaveOccurred(), "Harness job pods failed")
+
+			h.Cleanup(context.TODO())
+			ginkgo.By("======= FINISHED HARNESS: " + harness + " =======")
+		},
+		HarnessEntries)
 })
+
+// Generates templated command string to provide to test harness container
+func getCommandString(h *helper.H, harness string, r *runner.Runner, suffix string, jobName string) (string, error) {
```
jbpratt, line 74: Call `ginkgo.GinkgoHelper()` at the top of this function so that ginkgo properly handles the offset and get rid of the error return. That way we don't both check errors with gomega internally, and return an error as well

pkg/e2e/harness_runner/harness_runner.go
```diff
@@ -7,34 +7,114 @@ import (
...
-			if err := alert.SendSlackMessage(viper.GetString(config.Tests.SlackChannel), message); err != nil {
-				log.Printf("Failed sending slack alert for test failure: %v", err)
-			}
-		}
-	}, TimeoutInSeconds+30)
+	for _, harness := range harnesses {
+		HarnessEntries = append(HarnessEntries, ginkgo.Entry("should run "+harness+" successfully", harness))
+	}
+	ginkgo.DescribeTable("Executing Harness",
+		func(harness string) {
```
jbpratt, line 35:
```suggestion
		func(ctx context.Context, harness string) {
```
ginkgo can provide a context to the function that way you don't have to use `context.TODO()` below

jbpratt (commented):
Nice idea!

I think there is some opportunity to divide things up in a `BeforeEach` and `AfterEach` if desirable, but this is a great start!

pkg/e2e/harness_runner/harness_runner.go
```diff
@@ -7,34 +7,114 @@ import (
...
+			h = helper.New()
+			h.SetServiceAccount(context.TODO(), "system:serviceaccount:%s:cluster-admin")
+			harnessImageIndex := strings.LastIndex(harness, "/")
+			harnessImage := harness[harnessImageIndex+1:]
+			suffix := util.RandomStr(5)
+			jobName := fmt.Sprintf("%s-%s", harnessImage, suffix)
+			r := h.RunnerWithNoCommand()
+			testCommand, err := getCommandString(h, harness, r, suffix, jobName)
+			Expect(err).NotTo(HaveOccurred(), "Could not create test pod template")
+			r.Cmd = testCommand
```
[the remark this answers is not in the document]
jbpratt: I think you can get away with just doing
```
testCommand, err := getCommandString(h, harness, r, suffix, jobName)
Expect(err).NotTo(HaveOccurred(), "Could not create test pod template")
r := h.Runner(testCommand)
```

pkg/e2e/harness_runner/harness_runner.go
jbpratt: Still returning an error here

pkg/e2e/harness_runner/harness_runner.go
```diff
@@ -7,34 +7,123 @@ import (
...
+
+			// run tests
+			ginkgo.By("Running harness pod")
+			stopCh := make(chan struct{})
+			err = r.Run(int(TimeoutInSeconds), stopCh)
+			Expect(err).NotTo(HaveOccurred(), "Could not run pod")
+
+			// ensure job has not failed
+			_, err = h.Kube().BatchV1().Jobs(r.Namespace).Get(ctx, jobName, metav1.GetOptions{})
+			Expect(err).NotTo(HaveOccurred(), "Harness job pods failed")
```
jbpratt, line 77: let's drop this last check, it isn't doing anything for us

PR #1996, opened 2023-07-21
harness batch job related updates

ritmun:
* changed harness timeout to use shorter suite timeout per harness
* Moved result retrieval to test spec from afterEach to avoid execution for fail pods
* Adding harness list in complete-e2e config file instead of env var for batch job
* Skip migrated operator tests using new label - only for complete-e2e test config. This is a temporary stop gap till we can remove operator label and tests from osde2e

### DIFF
--- a/configs/complete-e2e.yaml
+++ b/configs/complete-e2e.yaml
@@ -1,3 +1,13 @@
 tests:
   ginkgoLabelFilter: >
-    (E2E || Operators || TestHarness || ServiceDefinition || AppBuilds) && !Informing
+    (E2E || Operators || TestHarness || ServiceDefinition || AppBuilds) && !(Informing || MigratedHarness)
+  testHarnesses:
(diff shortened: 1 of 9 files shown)


configs/complete-e2e.yaml
```diff
@@ -1,3 +1,13 @@
 tests:
   ginkgoLabelFilter: >
-    (E2E || Operators || TestHarness || ServiceDefinition || AppBuilds) && !Informing
\ No newline at end of file
+    (E2E || Operators || TestHarness || ServiceDefinition || AppBuilds) && !(Informing || MigratedHarness)
+  testHarnesses:
+  #    For migrated harnesses, remember to add "SkipHarness" label to existing suite in this repo, until they are removed from here.
```
ryankwilliams, line 5:
```suggestion
  #    For migrated harnesses, remember to add "MigratedHarness" label to existing suite in this repo, until they are removed from here.
```

I think this should be `MigratedHarness` over `SkipHarness`?

PR #2001, opened 2023-07-24
test/candidate_diff: initial commit of suite to diff two clusters

jbpratt:
automate the cluster-diff portion of gap analysis by provisioning two clusters and executing the script


test/candidate_diff/candidate_diff_test.go
```diff
@@ -0,0 +1,70 @@
...
+	// fail if there is a difference
+
+	var (
+		name1    = "osde2e-diff-1"
+		name2    = "osde2e-diff-2"
+		version1 = "4.13.5"
+		version2 = "4.14.0-ec.3"
+	)
+
+	BeforeAll(func(ctx context.Context) {
```
ryankwilliams, line 32: We should probably have an `AfterAll` to destroy the 2 clusters provisioned.

test/candidate_diff/candidate_diff_test.go
```diff
@@ -0,0 +1,70 @@
...
+	"golang.org/x/sync/errgroup"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var _ = Describe("Gap Analysis - Candidate Cluster Diff", Ordered, func() {
+	// provision two clusters: one Y-1, one Y
+	// run gap analysis script
+	// fail if there is a difference
+
+	var (
```
ryankwilliams, line 25: @jbpratt This [page](https://docs.ci.openshift.org/docs/architecture/ci-operator/#testing-with-an-existing-openshift-release) is what I was mentioning earlier about how a job triggered by ci-operator for a given nightly OCP build. Injects the build ID into environment as an environment variable (e.g. _RELEASE_IMAGE_LATEST_). Which from there osde2e current installer version selector picks up that and checks for it existing in OCM.

We should take this into account when this test suite runs from CI to pick up a nightly build to deploy as the second cluster. For the first cluster version, we could auto determine it based on the target version (Y-1) from the candidate channel/verify the version exists in OCM.

ryankwilliams:
For SDCICD-1062

PR #2013, opened 2023-07-27
Test Harness: support multiple reruns of osde2e against the same cluster

ryankwilliams:
# Change
Before this PR change, if a user was to run osde2e test harness, it creates a namespace for the secrets that the tests can access. The namespace continues to live on the cluster after osde2e finishes. New runs would fail before running test harness due to the namespace already exists.

This change deletes the secrets namespace as part of the clean up operations and also attempts to delete the secrets namespace prior to creating it.


ryankwilliams:
Test failure for `HyperShift` PR check is unaffected by these changes.

PR #2022, opened 2023-07-31
use HARNESS_TIMEOUT var to specify individual harness timeout, POLLING_TIMEOUT if unspecified

ritmun:
In a multi harness suite, individual harnesses must maintain short timeout limits, in order to allow other harnesses and other test suites to run within POLLING_TIMEOUT

Introducing HARNESS_TIMEOUT to use in such multi harness jobs. Current single harness jobs only specify POLLING_TIMEOUT. This PR allows that var to be used when HARNESS_TIMEOUT is not defined.

### DIFF
--- a/pkg/e2e/harness_runner/harness_runner.go
+++ b/pkg/e2e/harness_runner/harness_runner.go
@@ -29,8 +29,11 @@ var (
 
 var _ = ginkgo.Describe("Test harness", ginkgo.Ordered, ginkgo.ContinueOnFailure, label.TestHarness, func() {
 	harnesses := viper.GetStringSlice(config.Tests.TestHarnesses)
-	timeoutInSeconds = 3600 * viper.GetInt(config.Tests.SuiteTimeout)
-
+	if viper.IsSet(config.Tests.HarnessTimeout) {
+		timeoutInSeconds = viper.GetInt(config.Tests.HarnessTimeout)
+	} else {
+		timeoutInSeconds = viper.GetInt(config.Tests.PollingTimeout)
+	}
 	fmt.Println("Harnesses to run: ", harnesses)
 	for _, harness := range harnesses {
 		HarnessEntries = append(HarnessEntries, ginkgo.Entry("should run "+harness+" successfully", harness))
--- a/pkg/e2e/harness_runner/harness_runner.go
+++ b/pkg/e2e/harness_runner/harness_runner.go
@@ -73,7 +73,6 @@ var _ = ginkgo.Describe("Test harness", ginkgo.Ordered, ginkgo.ContinueOnFailure
 			//	today
 			ginkgo.By("Running harness pod")
 			stopCh := make(chan struct{})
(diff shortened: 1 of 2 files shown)


pkg/e2e/harness_runner/harness_runner.go
```diff
@@ -70,6 +73,7 @@ var _ = ginkgo.Describe("Test harness", ginkgo.Ordered, ginkgo.ContinueOnFailure
 			//	today
 			ginkgo.By("Running harness pod")
 			stopCh := make(chan struct{})
+			fmt.Println("Will timeout in", timeoutInSeconds, " seconds")
```
ryankwilliams, line 76: We should be able to remove this print statement, another print statement already prints out the timeout duration when waiting for the harness pod ([pkg/common/runner/runner.go#130](https://github.com/openshift/osde2e/blob/b6ae4ef29f843efd4080c2ebf15ceb728ddc1f97/pkg/common/runner/runner.go#L130)):

```
2023/08/01 13:10:59 runner.go:130: Waiting for endpoints of osde2e-runner runner Pod with a timeout of 21600 seconds...
```

pkg/e2e/harness_runner/harness_runner.go
```diff
@@ -29,8 +29,11 @@ var (
 
 var _ = ginkgo.Describe("Test harness", ginkgo.Ordered, ginkgo.ContinueOnFailure, label.TestHarness, func() {
 	harnesses := viper.GetStringSlice(config.Tests.TestHarnesses)
-	timeoutInSeconds = 3600 * viper.GetInt(config.Tests.SuiteTimeout)
-
+	if viper.IsSet(config.Tests.HarnessTimeout) {
```
ryankwilliams, line 32: I am okay with the new setting for a harness timeout making it specific/easy to associate to the other test harness settings. I do think we should have a follow up action to deprecate the `POLLING_TIMEOUT` setting. Just to reduce any confusion on which setting to use. It could be taken in a phased approach:

Phase 1: Remove using `POLLING_TIMEOUT` for test harness runner and remove this setting from all test harness jobs. Removal means PR's would need to be opened to adjust them to use the new setting `HARNESS_TIMEOUT`. We could even do it in this PR (https://github.com/openshift/release/pull/41828) I have opened.

Phase 2: Review osde2e and drop support for `POLLING_TIMEOUT` anywhere else its required. I think most of its use is passed to the ginkgo it nodes for timeouts but I believe those are not valid anymore now that we directly call `ginkgo.It()` over using the wrapper we used to have in osde2e. Of course this one would be a `low` priority. :)

This way osde2e has 2 primary timeout settings:

1. `SUITE_TIMEOUT` controls the overall duration ginkgo will run the test suite.
2. `HARNESS_TIMEOUT` controls the timeout to wait for a test harness pod to be finished.

These are just some of my thoughts, I am interested to hear your thoughts/the rest of the team!

ryankwilliams: Sounds good @ritmun

PR #2090, opened 2023-08-28
osde2e-common to different dependabot schedule

MrSantamaria:
Test PR
If this works it should check on a daily interval if osde2e-common has been updated


jbpratt:
/close

closing these out

PR #2130, opened 2023-09-21
debug: fix getLastJobID search for job ID

jbpratt:
the page has changed since this code previously worked, it is now dynamically rendered on the client side so the value is not available in the HTML directly. Parse the last script block on the page for the JSON object used to render the page. Then look for the second item in the list, assuming it is the previously ran job.

Yes, this is extremely fragile.

```
    x, err := debug.GetLastJobID("https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com", "periodic-ci-openshift-osde2e-main-nightly-4.14-rosa-classic-sts")
    if err != nil {
        panic(err)
    }

    fmt.Println(x)
```

```
❯ go run main.go
2023/09/21 09:47:47 Looking up job history from https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/job-history/gs/origin-ci-test/logs/periodic-ci-openshift-osde2e-main-nightly-4.14-rosa-classic-sts
1704653830467096576
```

https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/job-history/gs/origin-ci-test/logs/periodic-ci-openshift-osde2e-main-nightly-4.14-rosa-classic-sts contains a JSON object that we can parse to find the last build ID. (Surely there is a better way to do this :( but lets go with it)

### DIFF
--- a/pkg/debug/diff.go
+++ b/pkg/debug/diff.go
@@ -2,11 +2,13 @@ package debug
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"log"
 	"net/http"
 	"os"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


jbpratt:
/retest

I cleaned up the s3 buckets in the pr check account

PR #2131, opened 2023-09-22
misc: remove all db and pd support

jbpratt:
this functionality is not used and is rotting away, if we desire a signal from it we should revive it in a new form


jbpratt:
> @jbpratt I still use the DB and have plans to bring it back. Could you split this PR to be only the PagerDuty functionality. Let me know if you want a side session on the DB. You can also access it using the credentials in Vault to see what is currently in there.

Hey @MrSantamaria , I think it'd be great to get an overview of the DB and how you are currently using it. I'd be interested to see if it could help in cleaning up the pipelines.

I've pushed a new commit just pruning pagerduty support. PTAL

PR #2134, opened 2023-09-22
sdcicd-1075 added generic test harness resource template for saas pipeline

ritmun:
moving resource template back to osde2e repo for easier maintenance. The template **at this stage** doesn't have operator specific things in it. 

The change will be run through once this PR and the following one are merged

https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/81595/diffs


jbpratt:
Can we fix this file to be how you desire it instead? https://github.com/openshift/osde2e/blob/main/ci/acceptance-tests-template.yaml or delete this one

templates/test-harness-template.yaml
```diff
@@ -0,0 +1,64 @@
+# THIS FILE IS GENERATED BY BOILERPLATE. DO NOT EDIT.
```
jbpratt, line 1: is this still true? :smile_cat:

jbpratt:
merging this since rosa-pr-check is broken and this doesn't have any effect on code

PR #2183, opened 2023-10-19
e2e/operators: stop checking for olm clusterroles

jbpratt:
the rolename scheme has changed in 4.15 so it makes testing that these are there more difficult. That brings the question of value in testing that the clusterroles exist, which is just something managed by OLM.

remove any checks for OLM created clusterroles

[SDCICD-1151](https://issues.redhat.com//browse/SDCICD-1151)

### DIFF
--- a/pkg/e2e/operators/dvo.go
+++ b/pkg/e2e/operators/dvo.go
@@ -46,19 +46,12 @@ var _ = ginkgo.Describe(deploymentValidationOperatorTestName, label.Informing, g
 		defaultDesiredReplicas int32 = 1
 	)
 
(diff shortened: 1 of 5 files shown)


Gaikanomer9 (approved):
lgtm

jbpratt:
> /lgtm as we're getting away from these anyway (except DVO)
> 
> Imo, checking basic resources of operator exist on deployment is more of an acceptance criteria not e2e feature testing.
> 
> Does this mean the harnesses with this assertion will also be failing it? Should we remove those as well?

Yes, the harnesses that test these roles will also need to be updated :)

PR #2202, opened 2023-10-23
change runner to run pods in any given namespace

ritmun:
jira: https://issues.redhat.com/browse/SDCICD-1070

What:
- project creation is decoupled from helper setup
- A project can be initialized with given suffix and a runner instance can be set to use that project
- Implemented this in harness runner. Initialize project, assign namespace to it and run job in it, gather results and remove namespace

Why:
- This does not touch the main osde2e runner namespace, removes dependency from it, 
- does not use or overwrite viper configs for execution, as it is doing now, only uses the defined project instance

### DIFF
--- a/pkg/common/helper/helper.go
+++ b/pkg/common/helper/helper.go
@@ -111,35 +111,13 @@ func (h *H) Setup() error {
 
 	project := viper.GetString(config.Project)
 	if project == "" {
-		// setup project and dedicated-admin account to run tests
-		// the service account is provisioned but only used when specified
-		// also generates a unique name for the osde2e test run project
 		suffix := util.RandomStr(5)
-		project = "osde2e-" + suffix
-
-		viper.Set(config.Project, project)
-		ginkgo.GinkgoLogr.Info("Setup called for %s", project)
-
-		h.proj, err = h.createProject(ctx, suffix)
+		h.proj, err = h.SetupNewProject(ctx, suffix)
 		if h.OutsideGinkgo && err != nil {
-			return fmt.Errorf("failed to create project: %s", err.Error())
+			return fmt.Errorf("failed to set up project: %s", err.Error())
 		}
-		Expect(err).ShouldNot(HaveOccurred(), "failed to create project")
-		Expect(h.proj).ShouldNot(BeNil())
-
-		h.CreateServiceAccounts(ctx)
-
-		// Quick fix for Hypershift pipelines failing. Currently the RBAC is not being created in the projects.
-		if !viper.GetBool(config.Hypershift) {
-			err = wait.PollImmediate(1*time.Second, 1*time.Minute, func() (bool, error) {
-				_, err = h.Kube().RbacV1().RoleBindings(h.CurrentProject()).Get(ctx, "dedicated-admins-project-dedicated-admins", metav1.GetOptions{})
-				if apierrors.IsNotFound(err) {
-					return false, nil
-				}
-				return true, err
-			})
-			Expect(err).NotTo(HaveOccurred())
-		}
+		Expect(err).ShouldNot(HaveOccurred(), "failed to set up project")
+		viper.Set(config.Project, h.proj.Name)
 	} else {
 		_ = wait.PollImmediate(5*time.Second, 30*time.Second, func() (bool, error) {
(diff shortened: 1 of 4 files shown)


pkg/common/helper/helper.go
```diff
@@ -162,6 +136,36 @@ func (h *H) Setup() error {
...
+	// Quick fix for Hypershift pipelines failing. Currently the RBAC is not being created in the projects.
+	if !viper.GetBool(config.Hypershift) {
+		err = wait.PollImmediate(1*time.Second, 1*time.Minute, func() (bool, error) {
+			_, err = h.Kube().RbacV1().RoleBindings(v1project.Name).Get(ctx, "dedicated-admins-project-dedicated-admins", metav1.GetOptions{})
+			if apierrors.IsNotFound(err) {
+				return false, nil
+			}
+			return true, err
+		})
+		Expect(err).NotTo(HaveOccurred())
```
jbpratt, line 164: This error check needs to be gated like line 147 to not error check with gomega when we are outside of ginkgo

pkg/e2e/harness_runner/harness_runner.go
```diff
@@ -82,12 +79,14 @@ var _ = ginkgo.Describe("Test harness", ginkgo.Ordered, ginkgo.ContinueOnFailure
 			Expect(err).NotTo(HaveOccurred(), "Could not read results")
 			ginkgo.By("Writing results")
 			h.WriteResults(results)
+
+			ginkgo.By("Deleting harness namespace")
```
jbpratt, line 83: both the setup and teardown should be kept in `BeforeEach` and `AfterEach`

pkg/e2e/operators/dvo.go
```diff
@@ -64,14 +64,16 @@ var _ = ginkgo.Describe(deploymentValidationOperatorTestName, label.Informing, g
 		// Create and check test deployment
 		h.CreateProject(ctx, "dvo-test")
 		h.SetProjectByName(ctx, "osde2e-dvo-test")
+		proj, err := h.Project().ProjectV1().Projects().Get(ctx, "osde2e-dvo-test", metav1.GetOptions{})
+		Expect(err).NotTo(HaveOccurred(), "error fetching project")
```
jbpratt, line 68: Can this just be replaced with your new `SetupNewProject` function?

jbpratt (changes requested):

pkg/common/helper/helper.go
jbpratt: Yeah, let's just go ahead and add the same conditional. It will blow up if we ever run with this specific configuration (hypershift + outside of ginkgo)

pkg/e2e/harness_runner/harness_runner.go
```diff
@@ -39,33 +40,29 @@ var _ = ginkgo.Describe("Test harness", ginkgo.Ordered, ginkgo.ContinueOnFailure
...
 		func(ctx context.Context, harness string) {
 			log.Printf("======= RUNNING HARNESS: %s =======", harness)
+			suffix = "h-" + util.RandomStr(5)
+			h := helper.New()
 			harnessImageIndex := strings.LastIndex(harness, "/")
 			harnessImage := strings.Split(harness[harnessImageIndex+1:], ":")[0]
 			jobName := fmt.Sprintf("%s-%s", harnessImage, suffix)
 
+			ginkgo.By("Setting up new namespace")
+			// var err error
```
jbpratt, line 53:
```suggestion
```
outdated comment

pkg/common/helper/helper.go
```diff
@@ -111,35 +111,9 @@ func (h *H) Setup() error {
...
-			err = wait.PollImmediate(1*time.Second, 1*time.Minute, func() (bool, error) {
-				_, err = h.Kube().RbacV1().RoleBindings(h.CurrentProject()).Get(ctx, "dedicated-admins-project-dedicated-admins", metav1.GetOptions{})
-				if apierrors.IsNotFound(err) {
-					return false, nil
-				}
-				return true, err
-			})
-			Expect(err).NotTo(HaveOccurred())
-		}
+		h.proj, err = h.SetupNewProject(ctx, suffix)
```
jbpratt, line 115: need to handle this err

PR #2228, opened 2023-11-08
load token in harness pod

ritmun:
Load ci-secrets secret in harness namespace

Load the OCM_TOKEN from ci-secrets secret as env var on harness pod. 

Removes the need to query secrets from osde2e-ci-secrets namespace. 

Tested locally with osde2e / mnmo harness

Triggered by [SDCICD-1142](https://issues.redhat.com//browse/SDCICD-1142)


pkg/e2e/harness_runner/harness_runner.go
```diff
@@ -51,6 +51,8 @@ var _ = ginkgo.Describe("Test harness", ginkgo.Ordered, ginkgo.ContinueOnFailure
 		suffix = "h-" + util.RandomStr(5)
 		subProject, err = h.SetupNewProject(ctx, suffix)
 		Expect(err).NotTo(HaveOccurred(), "Could not set up harness namespace")
+		h.SetPassthruSecretInProject(ctx, subProject)
```
jbpratt, line 54: need to check the return value

PR #2298, opened 2024-01-17
sdcicd-1201 Create blocking cluster healthcheck test spec 

ritmun:
What?:
Created test suite for blocking osd-cluster-ready check

Why:
To enable running osd-cluster-ready as a ginkgo test spec instead of set-up  

Jira:
[sdcicd-1201](https://issues.redhat.com//browse/sdcicd-1201)

[x] tested locally on a stage cluster

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -35,7 +35,7 @@ require (
 	github.com/openshift/custom-domains-operator v0.0.0-20221118201157-bd1052dac818
 	github.com/openshift/managed-upgrade-operator v0.0.0-20230525042514-a9b8c1d2571c
 	github.com/openshift/must-gather-operator v0.1.2-0.20221011152618-7805956e1ded
-	github.com/openshift/osde2e-common v0.0.0-20240111205449-2d6094e0e389
+	github.com/openshift/osde2e-common v0.0.0-20240122091653-ac2a77b4ad4d
 	github.com/openshift/route-monitor-operator v0.0.0-20221118160357-3df1ed1fa1d2
 	github.com/openshift/splunk-forwarder-operator v0.0.0-20230525060151-2dc403aa8ff9
 	github.com/operator-framework/api v0.17.7
--- a/go.sum
+++ b/go.sum
@@ -679,6 +679,8 @@ github.com/openshift/must-gather-operator v0.1.2-0.20221011152618-7805956e1ded h
 github.com/openshift/must-gather-operator v0.1.2-0.20221011152618-7805956e1ded/go.mod h1:LHntdQf5jrwxpmyVMbwXUK491COBJ7buHvyi4YLlHho=
 github.com/openshift/osde2e-common v0.0.0-20240111205449-2d6094e0e389 h1:RuqZWven2AKLwOCWGBagVEv/4adWbmYIyZNE8eUPxCs=
 github.com/openshift/osde2e-common v0.0.0-20240111205449-2d6094e0e389/go.mod h1:lzmkYjtdf0EgPnoujLNAzorq/7vygGCcgceo/dQI1d0=
+github.com/openshift/osde2e-common v0.0.0-20240122091653-ac2a77b4ad4d h1:QKnQTXXOF/fHSr1vyY1TEmag6X3Hk1XErllqQQVxYqs=
+github.com/openshift/osde2e-common v0.0.0-20240122091653-ac2a77b4ad4d/go.mod h1:lzmkYjtdf0EgPnoujLNAzorq/7vygGCcgceo/dQI1d0=
 github.com/openshift/route-monitor-operator v0.0.0-20221118160357-3df1ed1fa1d2 h1:s97F2fQ5r+XGVfWDy5wgH611iIvIDusQcHtv5jJ36uw=
 github.com/openshift/route-monitor-operator v0.0.0-20221118160357-3df1ed1fa1d2/go.mod h1:I/8znbaxMGRub27brm57UnkR48QZEWO26j2HRscy5bQ=
 github.com/openshift/splunk-forwarder-operator v0.0.0-20230525060151-2dc403aa8ff9 h1:u7XXBdFB+uoA+YcvDjqmfhAJsZTGGGQOBqIXSBT4EUQ=


jbpratt (changes requested):
I think simply checking the job like this should work for now. The mcscupgrade suite kicks the job to watch it succeed again, after the case of an upgrade which we don't need to worry about for now. 

https://github.com/openshift/osde2e/blob/2378d6040935ed86ba15bed2a498b1742064c9a2/test/mcscupgrade/mcscupgrade_test.go#L365

pkg/e2e/openshift/osdclusterready.go
```diff
@@ -0,0 +1,41 @@
...
+		k8s, err = openshift.New(ginkgo.GinkgoLogr)
+		Expect(err).ShouldNot(HaveOccurred(), "Unable to setup k8s client")
+	})
+
+	ginkgo.It("should verify cluster is ready", func(ctx context.Context) {
+		clientSet, err := kubernetes.NewForConfig(k8s.GetConfig())
+		Expect(err).ShouldNot(HaveOccurred(), "failed to create kubernetes clientset")
+		duration, err := time.ParseDuration(viper.GetString(config.Tests.ClusterHealthChecksTimeout))
+		ctx, cancel := context.WithTimeout(context.Background(), duration)
+		defer cancel()
```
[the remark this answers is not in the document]
jbpratt: You can also use ginkgo timeout instead of setting up a new context to timeout

pkg/e2e/openshift/osdclusterready.go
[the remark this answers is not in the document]
jbpratt: While I understand it wasn't mention, this has been something we have been pushing for a few months (hence the directory existing to begin with). The direction we want to move is to be able to execute things with ginkgo directly. Adding this to e2e means we can't do that. The test runner is ginkgo IMO, we are orchestrating everything else in prow.

pkg/common/cluster/healthchecks/healthcheckjob.go
```diff
@@ -42,19 +40,20 @@ func CheckHealthcheckJob(k8sClient *kubernetes.Clientset, ctx context.Context, l
...
-					data, err := k8sClient.CoreV1().Pods(namespace).GetLogs(pod.Name, &v1.PodLogOptions{}).DoRaw(context.TODO())
-					if err != nil {
-						log.Printf("failed getting logs for pod %s: %s", pod.Name, err.Error())
-					}
-					if err = os.WriteFile(filepath.Join(viper.GetString(config.ReportDir), fmt.Sprintf("%s.log", pod.Name)), data, os.FileMode(0o644)); err != nil {
-						log.Printf("unable to output container logfile %s.log: %s", pod.Name, err.Error())
+			} else {
+				for _, pod := range pods.Items {
+					if strings.Contains(pod.Name, name) {
+						data, err := k8sClient.CoreV1().Pods(namespace).GetLogs(pod.Name, &v1.PodLogOptions{}).DoRaw(context.TODO())
```
jbpratt, line 46:
```suggestion
						data, err := k8sClient.CoreV1().Pods(namespace).GetLogs(pod.Name, &v1.PodLogOptions{}).DoRaw(ctx)
```

pkg/e2e/openshift/osdclusterready.go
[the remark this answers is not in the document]
mmazur: If the code scaffolding doesn't already exist to easily do it in the new way, please do it in the current way. The point of the epic was to do it with minimal code changes and we're sticking to that.

If I understand this discussion correctly, I'm fairly certain we'll want to do the changes Brady mentions as part of the followup epics, but I'll  want us to have a longer architectural discussion before that happens.

pkg/e2e/openshift/osdclusterready.go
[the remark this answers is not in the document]
jbpratt: Unresolving this, no changes have been made

pkg/e2e/openshift/osdclusterready.go
```diff
@@ -0,0 +1,57 @@
...
+		duration, err := time.ParseDuration(timeout)
+		Expect(err).ShouldNot(HaveOccurred(), "failed to parse duration")
+		ctx, cancel := context.WithTimeout(context.Background(), duration)
+		defer cancel()
+
+		job := new(batchv1.Job)
+		err = k8s.WithNamespace(namespace).Get(ctx, jobname, namespace, job)
+		Expect(err).ShouldNot(HaveOccurred(), "could not list ")
+
+		joberr := WatchJob(k8s, ctx, *job)
```
jbpratt, line 50: this can be simplified with https://github.com/openshift/osde2e-common/pull/61

pkg/e2e/openshift/osdclusterready.go
```diff
@@ -0,0 +1,57 @@
...
+		if joberr != nil {
+			logs, err = k8s.GetJobLogs(ctx, jobname, namespace)
```
jbpratt, line 52: need to check err

pkg/e2e/openshift/osdclusterready.go
```diff
@@ -0,0 +1,57 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var suiteName = "OSD Cluster Ready"
+var timeout = "30m"
+var namespace = "openshift-monitoring"
+var jobname = "osd-cluster-ready"
```
jbpratt, line 23: please make these constants and I feel like it should be formatted differently via gofmt

pkg/e2e/openshift/osdclusterready.go
```diff
@@ -0,0 +1,57 @@
...
+		log.SetLogger(ginkgo.GinkgoLogr)
+		var err error
+		k8s, err = openshift.New(ginkgo.GinkgoLogr)
+		Expect(err).ShouldNot(HaveOccurred(), "Unable to setup k8s client")
+	})
+
+	ginkgo.It("should verify cluster is ready", func(ctx context.Context) {
+		var logs string
+		duration, err := time.ParseDuration(timeout)
+		Expect(err).ShouldNot(HaveOccurred(), "failed to parse duration")
```
jbpratt, line 42: Is there a reason we don't just make a `var timeout = 30 * time.Minute` object? It is already a constant at the top of the file

PR #2300, opened 2024-01-21
osde2e cleanup command for iam resources

ritmun:
Note to reviewers: Please view diff with w=1 (by default the diff is space-based and shows code replaced rather moved/indented) 
https://github.com/openshift/osde2e/pull/2300/files?w=1

- Added following iam cleanup related flags to existing osde2e cleanup command
```
      --iam-dry-run                   Show dry run log of deleting iam resources (default true)
      --iam                          Cleanup iam resources
      --iam-older-than int         Cleanup iam resources older than this number of days (default 1)
```

 `./osde2e cleanup --iam`   deletes 

- oidc provider
- role (and detaches any policies) 
- policy 

 older than 24 hours
 
  `./osde2e cleanup --iam --iam-older-than 0`   deletes 

all the above osde2e-related resources without time filter

  `./osde2e cleanup --iam --iam-dry-run`   shows what will be deleted 

https://issues.redhat.com/browse/SDCICD-1206

### DIFF
--- a/cmd/osde2e/cleanup/cmd.go
+++ b/cmd/osde2e/cleanup/cmd.go
@@ -28,8 +28,8 @@ var args struct {
 	secretLocations string
 	clusterID       string
 	iam             bool
-	olderThan       int
-	dryrun          bool
+	iamOlderThan    int
+	iamDryRun       bool
 }
 
 func init() {
@@ -66,14 +66,14 @@ func init() {
 		"Cleanup iam resources",
 	)
 	flags.IntVar(
-		&args.olderThan,
-		"older-than",
+		&args.iamOlderThan,
+		"iam-older-than",
 		1,
 		"Cleanup iam resources older than this number of days",
 	)
 	flags.BoolVar(
-		&args.dryrun,
-		"dry-run",
+		&args.iamDryRun,
+		"iam-dry-run",
 		true,
 		"Show dry run log of deleting iam resources",
 	)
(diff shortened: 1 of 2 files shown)


cmd/osde2e/cleanup/cmd.go
```diff
@@ -54,54 +59,90 @@ func init() {
...
+		"older-than",
+		1,
+		"Cleanup iam resources older than this number of days",
+	)
+	flags.BoolVar(
+		&args.dryrun,
+		"dry-run",
+		true,
+		"Show dry run log of deleting iam resources",
+	)
```
jbpratt, line 79: these flags should be named more specifically if they apply to only `iam` (`--iam-dry-run`, `--iam-older-than`)

jbpratt: though it doesn't seem either of these are being used

pkg/common/aws/iam.go
```diff
@@ -0,0 +1,209 @@
...
+		}
+
+		// Get the provider
+		result, err := CcsAwsSession.iam.GetOpenIDConnectProvider(input)
+		if err != nil {
+			return err
+		}
+
+		// If provider name contains "cloudfront" and is older than given days, delete it
+		if strings.Contains(*result.Url, providersubstr) && time.Since(*result.CreateDate) > olderthan {
```
jbpratt, line 42: What is the reasoning for waiting? If the cluster is deleted, I would assume this resource is no longer valuable. I don't check for dates when I do my cleanup.

pkg/common/aws/iam.go
```diff
@@ -0,0 +1,209 @@
...
+			fmt.Printf("Provider will be deleted: %s\n", *provider.Arn)
+
+			if !dryrun {
+				// Create DeleteOpenIDConnectProviderInput struct
```
jbpratt, line 46: out of date comment

pkg/common/aws/iam.go
```diff
@@ -0,0 +1,209 @@
...
+	input := &iam.ListRolesInput{
+		MaxItems: aws.Int64(1000),
+	}
+	result, err := CcsAwsSession.iam.ListRoles(input)
+	if err != nil {
+		return err
+	}
+
+	for _, role := range result.Roles {
+		if strings.Contains(*role.Arn, rolesubstr) && time.Since(*role.CreateDate) > olderthan {
```
jbpratt, line 77: same for this, I would like to understand why we want to leave resources up for longer than the cluster is running

pkg/common/aws/iam.go
```diff
@@ -0,0 +1,209 @@
...
+
+			if errInstanceProfiles != nil {
+				return fmt.Errorf("error getting instance profiles for role: %s", errInstanceProfiles)
+			}
+
+			for _, instanceProfile := range instanceProfiles.InstanceProfiles {
+				fmt.Println("Removing role from instance profile: ", *instanceProfile.InstanceProfileName)
+
+				if !dryrun {
+					// Create RemoveRoleFromInstanceProfileInput struct
```
jbpratt, line 94: out of date comment

pkg/common/aws/iam.go
jbpratt: Oh, I guess that makes a bit of sense. I would prefer the implementation of this to filter out any cluster resources that are running but we can go with this.

PR #2311, opened 2024-01-22
SDCICD-1201 run osd-cluster ready as a test spec and apply blocking label

ritmun:
What?:
Created test suite for blocking osd-cluster-ready check
moved jobwatcher to osde2e common
 
Why:
To enable running osd-cluster-ready as a ginkgo test spec instead of set-up

Jira:
[sdcicd-1201](https://issues.redhat.com//browse/sdcicd-1201)

[x] tested locally on a stage cluster


go.mod
```diff
@@ -73,6 +72,8 @@ require (
 	sigs.k8s.io/e2e-framework v0.3.0
 )
 
+require github.com/openshift/osde2e-common v0.0.0-20240122091653-ac2a77b4ad4d
```
jbpratt, line 75: /hold

you should bump osde2e-common by doing a `go get github.com/openshift/osde2e-common@latest` then a `go mod tidy`

jbpratt:
https://github.com/openshift/osde2e-common/pull/61 we should pull this change in as well

pkg/common/cluster/healthchecks/healthcheckjob.go
```diff
@@ -6,110 +6,47 @@ import (
...
-			}
-			if !more {
-				return fmt.Errorf("cluster watch result channel closed prematurely with event: %T %v", event, event)
+
+	err = k8s.WatchJob(ctx, namespace, name)
+
+	if err == nil {
+		logger.Println("Healthcheck job passed")
+		return nil
+	} else {
```
jbpratt, line 37: This else block can be dedented since we return early in the above block

pkg/e2e/openshift/osdclusterready.go
```diff
@@ -0,0 +1,46 @@
...
+	})
+
+	ginkgo.It("should verify cluster is ready", func(ctx context.Context) {
+		joberr := k8s.WatchJob(ctx, namespace, jobname)
+		if joberr != nil {
+			logs, err := k8s.GetJobLogs(ctx, jobname, namespace)
+			if err != nil {
+				fmt.Print("could not get osd-cluster-ready logs", err)
+			} else {
+				fmt.Print("job log:", logs)
```
jbpratt, line 41: We should use ginkgologr to log messages in test cases

PR #2323, opened 2024-02-03
reuse existing osd-cluster-ready loop from osde2e-common

ritmun:
Reusing existing logic in osde2e-common for osdClusterReady check

Rewrite of this logic in osde2e is redundant.


YiqinZhang (commented):
/lgtm

PR #2329, opened 2024-02-07
Accept duration format string instead of number of days for iam resource cleanup command. 

ritmun:
`osde2e cleanup` will accept a time format string instead of number of days (eg 1h30m)

`osde2e cleanup --iam --iam-older-than 6h`

[SDCICD-1206](https://issues.redhat.com//browse/SDCICD-1206)


YiqinZhang (commented):
/lgtm

PR #2330, opened 2024-02-08
load secret locations from vault when running on prow jobs

ritmun:
[sdcicd-1206](https://issues.redhat.com//browse/sdcicd-1206)

osdctl cleanup --iam  command should load configs so that AWS creds can be read into  aws session


YiqinZhang (commented):
/lgtm

PR #2340, opened 2024-02-13
added s3 bucket cleanup to osde2e cleanup command

ritmun:
`osde2e cleanup --s3 --dry-run=false` 

 finds buckets with prefix "osde2e-" or "managed-velreo" older than given duration, then deletes bucket objects and then buckets

repurposes dry-run and older-than flags from existing command by removing prefix "iam" 

[SDCICD-1225](https://issues.redhat.com//browse/SDCICD-1225)

### DIFF
--- a/pkg/common/aws/s3.go
+++ b/pkg/common/aws/s3.go
@@ -6,12 +6,15 @@ import (
 	"log"
 	"net/url"
 	"strings"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/aws/aws-sdk-go/service/s3/s3manager"
 )
+
+const velerosubstr = "managed-velero"
 
 // ReadFromS3Session reads a key from S3 using given AWS context.
 func ReadFromS3Session(session *session.Session, inputKey string) ([]byte, error) {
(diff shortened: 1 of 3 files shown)


pkg/common/aws/s3.go
```diff
@@ -85,3 +88,50 @@ func ParseS3URL(s3URL string) (string, string, error) {
...
+					input := &s3.DeleteBucketInput{
+						Bucket: bucket.Name,
+					}
+					// Delete the policy
+					_, err := CcsAwsSession.s3.DeleteBucket(input)
+					if err != nil {
+						return err
+					}
+					fmt.Println("Deleted bucket")
+
```
jbpratt, line 128: nit: extra newline cleanup

pkg/common/aws/s3.go
```diff
@@ -85,3 +88,50 @@ func ParseS3URL(s3URL string) (string, string, error) {
...
+				})
+				if err := s3manager.NewBatchDeleteWithClient(CcsAwsSession.s3).Delete(aws.BackgroundContext(), iter); err != nil {
+					return err
+				} else {
+					fmt.Println("Deleted object(s) from bucket")
+
+					input := &s3.DeleteBucketInput{
+						Bucket: bucket.Name,
+					}
+					// Delete the policy
```
jbpratt, line 122: out of date comment

pkg/common/aws/s3.go
```diff
@@ -85,3 +88,50 @@ func ParseS3URL(s3URL string) (string, string, error) {
...
+					fmt.Println("Deleted bucket")
+
+				}
+			}
+		} else {
+			fmt.Printf("Bucket will NOT be deleted: %s\n", *bucket.Name)
+		}
+	}
+
+	return nil
```
jbpratt, line 136: I think we can simplify the conditions just a bit like this:

```suggestion
    if err := CcsAwsSession.GetAWSSessions(); err != nil {
        return err
    }

    result, err := CcsAwsSession.s3.ListBuckets(&s3.ListBucketsInput{})
    if err != nil {
        return err
    }

    batchDeleteClient := s3manager.NewBatchDeleteWithClient(CcsAwsSession.s3)

    for _, bucket := range result.Buckets {
        if !((strings.Contains(*bucket.Name, rolesubstr) || strings.Contains(*bucket.Name, velerosubstr)) && time.Since(*bucket.CreationDate) > olderthan) {
            fmt.Printf("Skipping bucket: %s\n", bucket)
            continue
        }

        if !dryrun {
            iter := s3manager.NewDeleteListIterator(CcsAwsSession.s3, &s3.ListObjectsInput{Bucket: bucket.Name})
            if err := batchDeleteClient.Delete(aws.BackgroundContext(), iter); err != nil {
                return err
            }
            fmt.Printf("Deleted object(s) from bucket: %s\n", bucket)
            if _, err := CcsAwsSession.s3.DeleteBucket(&s3.DeleteBucketInput{Bucket: bucket.Name}); err != nil {
                return err
            }
        }
        fmt.Printf("Deleted bucket: %s\n", bucket)
    }

    return nil

```

jbpratt: Please clean up the logs as well

pkg/common/aws/s3.go
```diff
@@ -85,3 +88,47 @@ func ParseS3URL(s3URL string) (string, string, error) {
...
+	for _, bucket := range result.Buckets {
+		if (strings.Contains(*bucket.Name, rolesubstr) || strings.Contains(*bucket.Name, velerosubstr)) && time.Since(*bucket.CreationDate) > olderthan {
+			fmt.Printf("Bucket will be deleted: %s\n", bucket)
+			if !dryrun {
+				// Setup BatchDeleteIterator to iterate through a list of objects.
+				iter := s3manager.NewDeleteListIterator(CcsAwsSession.s3, &s3.ListObjectsInput{
+					Bucket: bucket.Name,
+				})
+				if err := batchDeleteClient.Delete(aws.BackgroundContext(), iter); err != nil {
+					log.Printf("error deleting objects from bucket %s, skipping: %s", *bucket.Name, err)
```
jbpratt, line 116: Can we log or fmt, but not both? It leads to really messy output

cmd/osde2e/provision/cmd.go
```diff
@@ -0,0 +1,102 @@
...
+	_ "github.com/openshift/osde2e/pkg/e2e/proxy"
+	_ "github.com/openshift/osde2e/pkg/e2e/scale"
+	_ "github.com/openshift/osde2e/pkg/e2e/state"
+	_ "github.com/openshift/osde2e/pkg/e2e/verify"
+	_ "github.com/openshift/osde2e/pkg/e2e/webhooks"
+	_ "github.com/openshift/osde2e/pkg/e2e/workloads"
+)
+
+var Cmd = &cobra.Command{
+	Use:   "provision",
```
jbpratt, line 34: /hold

Is this in the wrong commit?

pkg/common/aws/s3.go
jbpratt: In my suggestion, I've updated the log messages to include the bucket name to make it clearer

jbpratt:
/lgtm

please follow up and fix the help messages for the flags that were changed

PR #2368, opened 2024-03-04
wait until ocm reports cluster install "ready" before proceeding with tests

ritmun:
1. Polling cluster ocm status "ready" is moved immediately after provision, outside of healthcheck.

There should be no use case of osde2e that skips cluster becoming "ready" prior to any processing. 

Therefore, adding basic ocm ready wait immediately after provision, outside of old healthcheck logic. Healthcheck can still be skipped.  

2. Function cleanup: removing unused arg from ClusterHEalthcheck function
3. Function cleanup: Moving "install Timeout" initializing inside  WaitForOCMProvisioning instead of 4 duplicate lines everytime it's invoked. 

 https://issues.redhat.com/browse/SDCICD-1192

### DIFF
--- a/pkg/common/cluster/clusterutil.go
+++ b/pkg/common/cluster/clusterutil.go
@@ -91,14 +91,7 @@ func WaitForClusterReadyPostInstall(clusterID string, logger *log.Logger) error 
 		return fmt.Errorf("error getting cluster provisioning client: %v", err)
 	}
 
-	installTimeout := viper.GetInt64(config.Cluster.InstallTimeout)
-	if viper.GetBool(config.Hypershift) {
-		// Install timeout 30 minutes for hypershift
-		installTimeout = 30
-	}
-	logger.Printf("Waiting %v minutes for cluster '%s' to be ready...\n", installTimeout, clusterID)
-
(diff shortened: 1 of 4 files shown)


pkg/common/cluster/clusterutil.go
```diff
@@ -236,7 +223,14 @@ func WaitForClusterReadyPostWake(clusterID string, logger *log.Logger) error {
...
 }
 
-func waitForOCMProvisioning(provider spi.Provider, clusterID string, installTimeout int64, logger *log.Logger, isUpgrade bool) (becameReadyAt time.Time, err error) {
+func WaitForOCMProvisioning(provider spi.Provider, clusterID string, logger *log.Logger, isUpgrade bool) (becameReadyAt time.Time, err error) {
+	installTimeout := viper.GetInt64(config.Cluster.InstallTimeout)
+	if viper.GetBool(config.Hypershift) {
+		// Install timeout 30 minutes for hypershift
+		installTimeout = 30
+	}
+	fmt.Printf("Waiting %v minutes for cluster to be ready...\n", installTimeout)
```
jbpratt, line 232: Please use the `logger` that way the logs are consistent

PR #2369, opened 2024-03-05
remove kubeconfig from conformance command

ritmun:
Conformance test errors with
`error converting to options: stat /tmp/kubeconfig: no such file or directoryerror: stat /tmp/kubeconfig: no such file or directory` 

This var is set after provision and doesn't need to be overridden here. 

[sdcicd-1192](https://issues.redhat.com//browse/sdcicd-1192)


YiqinZhang (commented):
/lgtm

PR #2423, opened 2024-04-09
SDCICD-1198 provide required gcp args to ocp tests

ritmun:
gcp conformance tests require "gce" as cloud type and a gcp project id.

### DIFF
--- a/pkg/e2e/openshift/config.go
+++ b/pkg/e2e/openshift/config.go
@@ -9,6 +9,7 @@ import (
 	. "github.com/onsi/gomega"
 	viper "github.com/openshift/osde2e/pkg/common/concurrentviper"
 	"github.com/openshift/osde2e/pkg/common/config"
+	"github.com/openshift/osde2e/pkg/common/providers/ocmprovider"
 )
 
 const testCmd = `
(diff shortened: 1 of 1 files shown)


pkg/e2e/openshift/config.go
```diff
@@ -99,9 +102,19 @@ func region() string {
...
 func cloud() string {
+	if viper.GetString(config.CloudProvider.CloudProviderID) == "gcp" {
+		return "gce"
+	}
 	return viper.GetString(config.CloudProvider.CloudProviderID)
 }
 
+func gcpproject() string {
+	if viper.GetString(config.CloudProvider.CloudProviderID) == "gcp" {
+		return ",\"projectid\":\"" + viper.GetString(ocmprovider.GCPProjectID) + "\""
```
jbpratt, line 113:
```suggestion
		return fmt.Sprintf(`,"projectid":%q`,  viper.GetString(ocmprovider.GCPProjectID) 
```

this may be a bit clearer

jbpratt: that is what the `%q` directive does for us! `%s` for strings `%q` for quoted strings https://pkg.go.dev/fmt

jbpratt:
```
❯ go run main.go
""test""
❯ cat -p -p main.go
package main

import "fmt"

func main() {
	fmt.Printf(`"%q"`, "test")
}
```

PR #2429, opened 2024-04-12
SDCICD-1240 skip waiting if cluster already uninstalling

ritmun:
1. skip waiting if cluster already uninstalling - this causes additional 15 min wait per cluster with no action 
2. replace pollImmediate with PollUntilContextTimeout ([SDCICD-1266](https://issues.redhat.com//browse/SDCICD-1266))


YiqinZhang (commented):
/lgtm

PR #2435, opened 2024-04-17
SDCICD-1266 deprecate more instances of pollImmediate

ritmun:
These instances directly affect prow jobs terminating abruptly while polling test pod, in [sdcicd-1198](https://issues.redhat.com//browse/sdcicd-1198) for example:

```"level"=0 "msg"="Waiting for test results using Endpoint 'osde2e-ueg2v/openshift-conformance-2zj9z'..."
  "level"=0 "msg"="Waiting for test results using Endpoint 'osde2e-ueg2v/openshift-conformance-2zj9z'..."
  "level"=0 "msg"="Waiting for test results using Endpoint 'osde2e-ueg2v/openshift-conformance-2zj9z'..."
  "msg"="unable to get pod osde2e-ueg2v/openshift-conformance-oz8kx" "error"="pods \"openshift-conformance-oz8kx\" not found"
  "level"=0 "msg"="Collecting logs from containers on openshift-conformance runner Pod..."
  [38;5;9m[FAILED][0m in [It] - /go/src/github.com/openshift/osde2e/pkg/e2e/openshift/conformance.go:103 [38;5;243m@ 04/17/24 05:46:46.086[0m```


hence updating these.

PR #2438, opened 2024-04-18
SDCICD-1266: pass context down through the runner

jbpratt:
pass down a context from the entrypoint of runner to each place it is used to allow taking advantage of context management with ginkgo in the future (cancelation/timeout on failure of test for example)

also resolve Int64Ptr/Int32Ptr deprecations while here

[SDCICD-1266](https://issues.redhat.com//browse/SDCICD-1266)

https://go.dev/blog/context


jbpratt:
this is just tacking onto the lint issue and further development on #2435

PR #2505, opened 2024-05-30
spin up a osde2e runner job as owner of test pod

ritmun:
currently test pod doesn't have a specific owner. This causes garbage collector to clean the pod and related objects abruptly during test runs. Instead, create a job to own the pod and namespace objects.

Discovered in conformance job for [sdcicd-1198](https://issues.redhat.com//browse/sdcicd-1198)

### DIFF
--- a/pkg/common/runner/pod.go
+++ b/pkg/common/runner/pod.go
@@ -7,6 +7,9 @@ import (
 	"path/filepath"
 	"time"
 
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+
 	"github.com/openshift/osde2e/pkg/common/util"
 	kubev1 "k8s.io/api/core/v1"
 	kerror "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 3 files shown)


pkg/common/runner/service_test.go
```diff
@@ -6,14 +6,31 @@ import (
...
 )
 
 func TestResultsService(t *testing.T) {
 	// setup mock client
-	client := fake.NewSimpleClientset()
+	client := fake.NewSimpleClientset(&corev1.Pod{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Pod",
+			APIVersion: "v1",
+		},
```
jbpratt, line 22: `TypeMeta` shouldn't be needed

pkg/common/runner/pod.go
```diff
@@ -152,15 +156,40 @@ func (r *Runner) createPod(ctx context.Context) (pod *kubev1.Pod, err error) {
...
+		if job, err = r.Kube.BatchV1().Jobs(r.Namespace).Create(ctx, job, metav1.CreateOptions{}); err != nil {
+			r.Error(err, fmt.Sprintf("error creating %s/%s runner Job", r.Namespace, r.Name))
 		}
 		return err == nil, nil
-	})
-	return createdPod, err
+	}); err != nil {
+		return nil, err
+	}
+	r.UID = job.UID
```
jbpratt, line 178: Does the job have a UID after creation or do we need to create a new UUID for it? https://pkg.go.dev/k8s.io/apimachinery/pkg/util/uuid#NewUUID

I'd also be surprised if we still need the ownerreference stuff since we are creating a job now. Were you able to test this out locally?

PR #2538, opened 2024-06-11
SDCICD-1289 - Enable OSDe2e to use Keycloack Client credentials

MrSantamaria:
Implements this change: 
https://github.com/openshift/osde2e-common/commit/b4f22da714553429cda714861dc2c2e0024ffe3c

As part of supporting the Fedramp project OSDe2e will need to support Keycloack client for login. 
This PR adds the ability to use clientID and secret to create an OCM/Rosa connection.

### DIFF
--- a/pkg/common/providers/ocmprovider/config.go
+++ b/pkg/common/providers/ocmprovider/config.go
@@ -39,21 +39,6 @@ const (
 	// CCS defines whether the cluster should expect cloud credentials or not
 	CCS = "ocm.ccs"
 
-	// GCP CCS Credential json
-	// Env: GCP_CREDS_JSON, OCM_GCP_CREDS_JSON
-	GCPCredsJSON = "ocm.gcp.credsJSON"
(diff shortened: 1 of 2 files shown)


pkg/common/providers/rosaprovider/rosa.go
```diff
@@ -33,21 +34,36 @@ type ROSAProvider struct {
...
 // New will create a new ROSAProvider.
 func New() (*ROSAProvider, error) {
+	fedramp := viper.GetBool(config.Cluster.FedRamp)
 	rosaEnv := viper.GetString(Env)
+	var ocmEnv ocmclient.Environment
+	var ocmProvider *ocmprovider.OCMProvider
+	var err error
 
-	ocmProvider, err := ocmprovider.NewWithEnv(rosaEnv)
+	ocmProvider, err = ocmprovider.NewWithEnv(rosaEnv)
```
jbpratt, line 43: is this change needed?

PR #2541, opened 2024-06-14
SDCICD-1306: add new EC compatible Dockerfile

jbpratt:
when trying to build the existing Dockerfile, it fails the enterprise-contract policy due to the base image not being allowed. We should migrate to using an image built from the allowed registries

```
[report]   - metadata:
[report]       code: base_image_registries.base_image_permitted
[report]       collections:
[report]       - minimal
[report]       - redhat
[report]       depends_on:
[report]       - base_image_registries.base_image_info_found
[report]       - base_image_registries.allowed_registries_provided
[report]       description: Verify that the base images used when building a container image
[report]         come from a known set of trusted registries to reduce potential supply chain
[report]         attacks. By default this policy defines trusted registries as registries that
[report]         are fully maintained by Red Hat and only contain content produced by Red Hat.
[report]         The list of permitted registries can be customized by setting the `allowed_registry_prefixes`
[report]         list in the rule data. To exclude this rule add "base_image_registries.base_image_permitted"
[report]         to the `exclude` section of the policy configuration. To exclude this rule
[report]         add "base_image_registries.base_image_permitted" to the `exclude` section
[report]         of the policy configuration.
[report]       solution: Make sure the image used in each task comes from a trusted registry.
[report]         The list of trusted registries is a configurable xref:ec-cli:ROOT:configuration.adoc#_data_sources[data
[report]         source].
[report]       title: Base image comes from permitted registry
[report]     msg: Base image "registry.ci.openshift.org/openshift/release" is from a disallowed
[report]       registry
```

### DIFF
--- a/.tekton/osde2e-main-pull-request.yaml
+++ b/.tekton/osde2e-main-pull-request.yaml
@@ -19,7 +19,7 @@ metadata:
 spec:
   params:
   - name: dockerfile
-    value: Dockerfile.osde2e
+    value: osde2e.Dockerfile
   - name: git-url
     value: '{{source_url}}'
   - name: image-expires-after
--- a/.tekton/osde2e-main-push.yaml
+++ b/.tekton/osde2e-main-push.yaml
@@ -18,7 +18,7 @@ metadata:
 spec:
   params:
   - name: dockerfile
-    value: Dockerfile.osde2e
+    value: osde2e.Dockerfile
   - name: git-url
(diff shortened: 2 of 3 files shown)


jbpratt:
/override "Red Hat Konflux / osde2e-e2e / osde2e-main" "ci/prow/rosa-pr-check" "ci/prow/hypershift-pr-check"

PR #2542, opened 2024-06-14
change random region selector to provider-agnostic ocm client

ritmun:
allows region selection for random region for any cloud provider

GCP was previously excluded from random region, and aws region was found using aws creds. This isn't necessary since ocm client returns region list without provider creds.

### DIFF
--- a/pkg/common/providers/ocmprovider/cluster.go
+++ b/pkg/common/providers/ocmprovider/cluster.go
@@ -407,36 +407,21 @@ func (o *OCMProvider) DetermineRegion(cloudProvider string) (string, error) {
 	if region == "random" {
 		var regions []*v1.CloudRegion
 		// We support multiple cloud providers....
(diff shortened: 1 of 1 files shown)


pkg/common/providers/ocmprovider/cluster.go
```diff
@@ -407,36 +407,21 @@ func (o *OCMProvider) DetermineRegion(cloudProvider string) (string, error) {
...
 			}
-			regions = response.Items().Slice()
-		}
-
-		// But we don't support passing GCP credentials yet :)
-		if cloudProvider == "gcp" {
-			log.Println("Random GCP region not supported yet. Setting region to us-east1")
-			return "us-east1", nil
+			if cloudProvider == "gcp" {
+				log.Println("Defaulting to us-east1")
```
jbpratt, line 418: log can be moved on to like 421 and formatted with `region` variable

PR #2544, opened 2024-06-14
e2e/webhooks: use dynamic client

jbpratt:
use the dynamic client to create CustomDomain CRs and MustGather CRs so that we don't pull in the dependency of both operators


jbpratt:
/override ci/prow/hypershift-pr-check 

hypershift doesn't run these tests, no point in running this job

PR #2545, opened 2024-06-14
move version selection to beforeSuite

ritmun:
move version selection to beforeSuite

it is only applicable to cluster launch, moving it with the same

### DIFF
--- a/pkg/e2e/e2e.go
+++ b/pkg/e2e/e2e.go
@@ -79,6 +79,7 @@ func beforeSuite() bool {
 		log.Printf("Error getting cluster provider: %s", err.Error())
 		return false
 	}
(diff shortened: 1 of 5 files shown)


pkg/e2e/e2e.go
```diff
@@ -188,6 +198,34 @@ func beforeSuite() bool {
 	return true
 }
 
+func configureVersion(provider2 *spi.Provider) (int, error) {
```
jbpratt, line 201: this `provider2` var isn't used, looks like we are using the global one

jbpratt:
/lgtm

@ritmun feel free to manually merge once the hypershift and rosa job pass!

PR #2558, opened 2024-06-20
SDCICD-1306: use els base image

jbpratt:
when using the ubi-minimal image build from the golang-builder, we are getting a `glibc not found` error when trying to run osde2e. Swap to the rhel9-2-els image that is used by other konflux builds


jbpratt:
/override ci/prow/rosa-pr-check ci/prow/hypershift-pr-check 
ci/prow/code-quality-checks 

should have no changes to these jobs

PR #2572, opened 2024-06-24
use CLOUD_PROVIDER_REGION when AWS_REGION is not specified

ritmun:
CLOUD_PROVIDER_REGION will be passed to jobs, and should be seen by AWS session as AWS_REGION

Bind it to viper config as a backup for AWS_REGION env var


YiqinZhang (commented):
/lgtm

PR #2592, opened 2024-07-01
SDCICD-885 remove migrated CIO test file

ritmun:
- removing migrated rh-api test to operator in https://github.com/openshift/cloud-ingress-operator/pull/332
- adding CIO image to e2e-suite to be included in nightlies

First installment of cio migrations. CIO has several e2e test files in osde2e, which will be moved subsequently


YiqinZhang (commented):
/lgtm

PR #2611, opened 2024-07-11
SDCICD-1330 EC2 instance cleanup

varunraokadaparthi:
- `--ec2` parameter terminates EC2 instances with tag `Name: osde2e*` and older than 24 hrs.
- Ignores EC2 instances with tag `Name: osde2e-proxy*`
- Notifcation sample: https://redhat-internal.slack.com/archives/C06HQR8HN0L/p1721926378565619

### DIFF
--- a/cmd/osde2e/cleanup/cmd.go
+++ b/cmd/osde2e/cleanup/cmd.go
@@ -2,6 +2,8 @@ package cleanup
 
 import (
 	"bytes"
+	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
@@ -38,6 +40,16 @@ var args struct {
 	dryRun          bool
 	clusters        bool
 	sendSummary     bool
+	ec2             bool
+}
+
+type Message struct {
+	Summary   string `json:"summary"`
+	BuildFile string `json:"buildfile"`
+	S3Errors  string `json:"s3"`
+	IAMErrors string `json:"iam"`
+	IPErrors  string `json:"ip"`
+	EC2Errors string `json:"ec2"`
 }
 
 func init() {
@@ -113,6 +125,13 @@ func init() {
 		"Send cleanup summary to webhook",
 	)
 
+	flags.BoolVar(
+		&args.ec2,
+		"ec2",
+		false,
+		"Terminate ec2 instances",
+	)
+
 	Cmd.RegisterFlagCompletionFunc("output-format", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 		return []string{"json", "prom"}, cobra.ShellCompDirectiveDefault
 	})
@@ -133,9 +152,10 @@ func run(cmd *cobra.Command, argv []string) error {
 	var iamErrorBuilder strings.Builder
 	var s3ErrorBuilder strings.Builder
 	var ipErrorBuilder strings.Builder
+	var ec2ErrorBuilder strings.Builder
 
 	if args.dryRun {
-		summaryBuilder.WriteString("-- Cleanup dry run --\n")
+		summaryBuilder.WriteString("-- Cleanup dry run -- \n")
 	}
 	if os.Getenv("JOB_NAME") != "" {
 		r, _ := regexp.Compile("cleanup-([a-z]+)-aws")
@@ -219,9 +239,24 @@ func run(cmd *cobra.Command, argv []string) error {
 		s3BucketDeletedCounter := 0
 		s3BucketFailedCounter := 0
 		err = aws.CcsAwsSession.CleanupS3Buckets(fmtDuration, args.dryRun, args.sendSummary, &s3BucketDeletedCounter, &s3BucketFailedCounter, &s3ErrorBuilder)
-		summaryBuilder.WriteString("s3 Buckets: " + strconv.Itoa(s3BucketDeletedCounter) + "/" + strconv.Itoa(s3BucketFailedCounter) + "\n")
+		summaryBuilder.WriteString("S3 Buckets: " + strconv.Itoa(s3BucketDeletedCounter) + "/" + strconv.Itoa(s3BucketFailedCounter) + "\n")
 		if err != nil {
 			return fmt.Errorf("could not delete s3 buckets: %s", err.Error())
+		}
+	}
+
+	if args.ec2 {
+		instancesDeleted, instancesFailedToDelete, err := aws.CcsAwsSession.TerminateEC2Instances(fmtDuration, args.dryRun)
+		summaryBuilder.WriteString("EC2 Instances: " + strconv.Itoa(instancesDeleted) + "/" + strconv.Itoa(instancesFailedToDelete) + "\n")
+		if err != nil {
+			if !errors.Is(err, aws.ErrTerminateEC2Instances) {
+				return fmt.Errorf("could not terminate ec2 instances: %s", err.Error())
+			}
+			ec2ErrorMessage := err.Error()
+			if len(ec2ErrorMessage) > config.SlackMessageLength {
+				ec2ErrorMessage = ec2ErrorMessage[:config.SlackMessageLength]
+			}
+			ec2ErrorBuilder.WriteString(ec2ErrorMessage)
 		}
 	}
 
@@ -241,7 +276,7 @@ func run(cmd *cobra.Command, argv []string) error {
 			fmt.Println("Slack Webhook is not set, skipping notification.")
 			return nil
 		}
-		buildFile := "Build file: "
+		buildFile := ""
 		if strings.Contains(viper.GetString(config.JobName), "rehearse") {
 			basePRJobURL := "https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_release"
 			buildFile += basePRJobURL + "/" + os.Getenv("PULL_NUMBER")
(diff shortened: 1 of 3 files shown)


cmd/osde2e/cleanup/cmd.go
```diff
@@ -225,6 +234,16 @@ func run(cmd *cobra.Command, argv []string) error {
 		}
 	}
 
+	if args.ec2 {
+		ec2InstanceTerminatedCounter := 0
+		ec2InstanceFailedCounter := 0
+		err = aws.CcsAwsSession.TerminateEC2Instances(fmtDuration, args.dryRun, args.sendSummary, &ec2InstanceTerminatedCounter, &ec2InstanceFailedCounter, &ec2ErrorBuilder)
```
jbpratt, line 240: I know everything else does it, so we can go with it but it is a bit odd to pass pointers to ints as parameters rather than just returning the counters as return values from `TerminateEC2Instances`. Just FWIW

cmd/osde2e/cleanup/cmd.go
```diff
@@ -255,9 +274,10 @@ func run(cmd *cobra.Command, argv []string) error {
 		errorMessage := `"full":"` + buildFile + `",`
 		s3ErrorMessage := `"s3":" S3 Errors: \n ` + s3ErrorBuilder.String() + `",`
 		iamErrorMessage := `"iam":" IAM Errors: \n` + iamErrorBuilder.String() + `",`
-		ipErrorMessage := `"ip":" ElasticIP Errors: \n` + ipErrorBuilder.String() + `"}`
+		ipErrorMessage := `"ip":" ElasticIP Errors: \n` + ipErrorBuilder.String() + `",`
+		ec2ErrorMessage := `"ec2":" EC2 Errors: \n` + ec2ErrorBuilder.String() + `"}`
```
jbpratt, line 278: Another: while things are like this currently, it is fine and we can proceed. I don't recommend doing something like this on a the future project though, this is super fragile. Just create a struct and json encode it.

pkg/common/aws/ec2.go
```diff
@@ -79,3 +85,49 @@ func (CcsAwsSession *ccsAwsSession) ReleaseElasticIPs(dryrun bool, sendSummary b
 
 	return nil
 }
+
+// TerminateEC2Instances finds EC2 instances with tag Name "osde2e*",
+// older than given duration, then terminates these EC2 instances.
+// Ignores EC2 instances with tag Name "osde2e-proxy*".
+func (CcsAwsSession *ccsAwsSession) TerminateEC2Instances(olderthan time.Duration, dryrun bool, sendSummary bool,
+	terminatedCounter *int, failedCounter *int, errorBuilder *strings.Builder,
+) error {
```
jbpratt, line 94: Please format this so it is one param per line like so or the entire definition on a single line. The rest of the file needs cleaner formatting in a later PR.

```suggestion
func (CcsAwsSession *ccsAwsSession) TerminateEC2Instances(
        olderthan time.Duration,
        dryrun bool,
        sendSummary bool,
	terminatedCounter *int,
	failedCounter *int,
	errorBuilder *strings.Builder,
) error {
```

pkg/common/aws/ec2.go
```diff
@@ -79,3 +85,49 @@ func (CcsAwsSession *ccsAwsSession) ReleaseElasticIPs(dryrun bool, sendSummary b
...
+	}
+	result, err := CcsAwsSession.ec2.DescribeInstances(&ec2.DescribeInstancesInput{})
+	if err != nil {
+		return err
+	}
+	for _, reservation := range result.Reservations {
+		for _, instance := range reservation.Instances {
+			if time.Since(*instance.LaunchTime) > olderthan {
+				for _, tag := range instance.Tags {
+					if *tag.Key == "Name" && !strings.Contains(*tag.Value, osde2eproxysubstr) && strings.Contains(*tag.Value, osde2esubstr) {
```
jbpratt, line 107: Why aren't we cleaning up _everything_ regardless of it containing this tag? This only cleans up instances that were created by osde2e but realistically it should delete everything regardless (someone comes in and creates 10 instances manually)

pkg/common/aws/ec2.go
```diff
@@ -79,3 +85,49 @@ func (CcsAwsSession *ccsAwsSession) ReleaseElasticIPs(dryrun bool, sendSummary b
...
+						if !dryrun {
+							input := &ec2.TerminateInstancesInput{
+								InstanceIds: []*string{instance.InstanceId},
+							}
+							_, err := CcsAwsSession.ec2.TerminateInstances(input)
+							if err != nil {
+								*failedCounter++
+								errorMsg := fmt.Sprintf("Instance %s failed to terminate: %s\n", *instance.InstanceId, err.Error())
+								fmt.Println(errorMsg)
+								if sendSummary && errorBuilder.Len() < config.SlackMessageLength {
```
jbpratt, line 118: is there a case where this is true?

pkg/common/aws/ec2.go
```diff
@@ -79,3 +85,49 @@ func (CcsAwsSession *ccsAwsSession) ReleaseElasticIPs(dryrun bool, sendSummary b
...
+								continue
+							}
+							*terminatedCounter++
+							fmt.Println("Terminated")
+						}
+						break
+					}
+				}
+			}
+		}
```
jbpratt, line 130: Parts of this can be simplified by flipping your logic around. Rather than `if instance is older than N let's delete it`, you can think `if instance is not older than N let's skip it`. Same for the tag key/value containing the substring. After that, I recommend inlining any variables that you can and only printing once (unless we fail).

```suggestion
    for _, reservation := range result.Reservations {
        for _, instance := range reservation.Instances {
            if time.Since(*instance.LaunchTime) < olderthan {
                continue
            }
            for _, tag := range instance.Tags {
                if *tag.Key == "Name" && strings.Contains(*tag.Value, osde2eproxysubstr) {
                    continue
                }
                fmt.Printf("Terminating instance: %s\n", instance)
                if !dryrun {
                    _, err = CcsAwsSession.ec2.TerminateInstances(&ec2.TerminateInstancesInput{InstanceIds: aws.StringSlice([]string{*instance.InstanceId})})
                    if err != nil {
                        *failedCounter++
                        errorMsg := fmt.Sprintf("Instance %s failed to terminate: %s\n", *instance.InstanceId, err.Error())
                        fmt.Println(errorMsg)
                        if sendSummary && errorBuilder.Len() < config.SlackMessageLength {
                            errorBuilder.WriteString(errorMsg)
                        }
                        break
                    }
                    *terminatedCounter++
                }
            }
        }
```

jbpratt (changes requested):
Few things we can change. I know a lot of this was replicated from existing code so I'm fine with a few of the quirks. Let me know if you have any questions. Thanks for working on this!

/cc @ritmun @YiqinZhang @MrSantamaria

cmd/osde2e/cleanup/cmd.go
varunraokadaparthi: I didn't want to change the function's return type. As it would have to return an error along with counter. Please do let me know if that way is more intuitive.

pkg/common/aws/ec2.go
varunraokadaparthi: As long as sendSummary is enabled and there are few to no errors.  The total slack notification length has to be under limits.

pkg/common/aws/ec2.go
jbpratt: I think it is fine to proceed with removing this part of condition. There is no reason to restrict something we know that we want to get rid of anyway for the sake of consistency

pkg/common/aws/ec2.go
varunraokadaparthi:
```
              if *tag.Key == "Name" && strings.Contains(*tag.Value, osde2eproxysubstr) {
                    continue
                }
```
The tags are a list and they seem to be in random order. If an EC2 instance does belong to osde2e-proxy but the Name tag is not the first in tag list then that EC2 instance would be deleted. This shouldn't happen I think. 
There should be an if condition(not the if not) here right?

pkg/common/aws/ec2.go
varunraokadaparthi: I tried to follow approach similar to s3 buckets where only `osde2e` and `managed-velereo` are removed. But EC2 instances seem to be of only 2 types, `osde2e-proxy` and `osde2e`. Is it safe to assume we won't have any other types?

jbpratt: I would default to deleting everything besides the proxy.

pkg/common/aws/ec2.go
jbpratt: Let's sync tomorrow and think about rewriting this. I think we should try something like looping over to build up a list of instances to delete and provide that as one slice to the terminateinstances param

jbpratt:
Nice! @varunraokadaparthi this looks solid. Can you post the output when executing the command? You can add the `//nolint:gocyclo` comment to run function to resolve the lint issue. I think we've simplified it enough overall that it isn't a problem.

jbpratt (commented):
nice, looking good! Ready to merge after one small nit

pkg/common/aws/ec2.go
```diff
@@ -3,13 +3,20 @@ package aws
...
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
 )
 
+const (
+	tagKeyForExemptEC2Instances = "osde2e-proxy"
+)
+
+var ErrOutOfTerminateEC2Instance = fmt.Errorf("Unable to terminate EC2 instances: \n")
```
jbpratt, line 18:
```suggestion
var ErrTerminateEC2Instances = fmt.Errorf("unable to terminate EC2 instances")
```
drop the `OutOf` and treat the error string as a fragment of a larger string. It should be possible to join together errors to make a legible sentence (`failed to run osde2e cleanup: unable to terminate EC2 instances: authentication failed` for example

varunraokadaparthi: Done

varunraokadaparthi:
//nolint:gocyclo

PR #2685, opened 2024-09-09
fix param name for specific nightly in parameterized jenkins job template 

ritmun:
SREs frequently need to spin up specific nighlty clusters for testing. 

To trigger specific nightly, we are now using INSTALL_LATEST_NIGHTLY instead of previous INSTALL_LATEST_XY 

Update parameterized job to accept it


jbpratt (approved):
/lgtm

PR #2701, opened 2024-09-24
ci: remove unused scripts

jbpratt:
these are no longer in use and haven't been for a few years


jbpratt:
/override "Red Hat Konflux / osde2e-enterprise-contract / osde2e-main"

PR #2713, opened 2024-09-27
enable passing user ca bundle data as file to rosa

ritmun:
- rosa accepts file path for additional user ca bundle
- jobs provide user ca bundle data
This change allows storing ca bundle in temp file and passing that path to rosa

[ Gap analysis 4.17 ]

### DIFF
--- a/pkg/common/config/config.go
+++ b/pkg/common/config/config.go
@@ -955,7 +955,7 @@ func init() {
 
 func WriteSecret(name string) string {
 	tmpDir, _ := os.MkdirTemp("", "")
-	tmpFile, _ := os.CreateTemp(tmpDir, "*")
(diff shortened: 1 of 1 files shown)


pkg/common/config/config.go
```diff
@@ -945,6 +953,14 @@ func init() {
 	InitGCPViper()
 }
 
+func WriteSecret(name string) string {
+	tmpDir, _ := os.MkdirTemp("", "")
+	tmpFile, _ := os.CreateTemp(tmpDir, "*")
```
jbpratt, line 958:
```suggestion
	tmpFile, _ := os.CreateTemp("", "osde2e-userca-bundle-data-")
```
I think just writing to base tmp is fine but I'd give it a prefixed name

PR #2721, opened 2024-10-02
Updates to Fedramp test Dockerfile

MrSantamaria:
Made some changes that were causing the container from being built.

### DIFF
--- a/test/fedramp/Dockerfile
+++ b/test/fedramp/Dockerfile
@@ -7,7 +7,7 @@ WORKDIR ${PKG}
 # Copy go.mod and go.sum to the working directory
 COPY ../../go.mod ../../go.sum ./
 COPY ../../LICENSE .
-COPY . .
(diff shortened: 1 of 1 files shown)


test/fedramp/Dockerfile
```diff
@@ -24,7 +24,7 @@ COPY --from=builder /go/src/github.com/openshift/osde2e/test/fedramp/osde2e-fedr
 COPY --from=builder /go/src/github.com/openshift/osde2e/test/fedramp/LICENSE /licenses/.
 
 # Set the entry point to 'ginkgo'
-ENTRYPOINT ["ginkgo"]
+ENTRYPOINT ["/usr/local/bin/ginkgo"]
 USER 65532:65532
 
 # Provide default arguments
```
jbpratt, line 30: I noticed L31 last argument needs to be updated as well, the file name is wrong

PR #2729, opened 2024-10-08
pkg/common/versions: increase triggered nightly timeout

jbpratt:
it still times out at 30 minutes sometimes


jbpratt:
/override "Red Hat Konflux / osde2e / osde2e-main"

PR #2732, opened 2024-10-10
update validate-sdn-ovn-migration-jobs.sh vars

reedcort:
Update validate-sdn-ovn-migration-jobs.sh env vars to match env vars in https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/resources/jenkins/osde2e/secrets.yaml?ref_type=heads#L29 so the script can run as a Jenkins job


joshbranham (commented):
/lgtm

PR #2773, opened 2024-11-06
SDCICD-1411 add ability to alternate between two aws accounts 

ritmun:
[SDCICD-1411](https://issues.redhat.com//browse/SDCICD-1411) add ability to alternate between two aws accounts when AWS_SHARED_CREDENTIALS are provided in env var

- AWSSharedCredentials is the base64 encoded AWS credentials file content.
- Variable should contain two named profiles: "profile-a" and "profile-b"
If provided, supersedes AWS secret set in env.
- This is used to optimize AWS resource spending by nuking each of the 2 accounts alternately.

-  Since simply checking whether current week is odd or even
 within current year may result in unexpected outage if a year has
 odd number of weeks, use odd/even based on a constant start date.

### DIFF
--- a/pkg/common/config/aws_config.go
+++ b/pkg/common/config/aws_config.go
@@ -2,7 +2,19 @@
 package config
 
 import (
+	"encoding/base64"
+	"fmt"
+	"math"
+	"os"
+	"path/filepath"
+	"time"
+
 	viper "github.com/openshift/osde2e/pkg/common/concurrentviper"
+)
+
+const (
+	profileA = "profile-a"
+	profileB = "profile-b"
 )
 
 var (
@@ -12,8 +24,20 @@ var (
 	// AWSAccessKey is the AWS access key
 	AWSAccessKey = "config.aws.accessKey"
 
-	// AWSSecretKey is the AWS secret access key
+	// AWSSecretAccessKey is the AWS secret access key
 	AWSSecretAccessKey = "config.aws.secretAccessKey"
+
+	// AWSSharedCredentials is the base64 encoded AWS credentials file content.
+	// This is used to optimize AWS resource spending
+	// by nuking each of the 2 accounts alternately.
+	// Should contain two named profiles: "profile-a" and "profile-b"
+	// If provided, supersedes AWS secret set in env.
+	AWSSharedCredentials = "config.aws.sharedCredentials"
+
+	// AWSCredentialsFile is the custom AWS credentials full file path including filename
+	// This is where the provided credentials will be saved
+	// AWS client uses this as a custom shared credentials file
+	AWSCredentialsFile = "config.aws.sharedCredentialsFilePath"
 
 	// AWSProfile is the AWS profile to use
 	AWSProfile = "config.aws.profile"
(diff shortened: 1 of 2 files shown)


pkg/common/config/aws_config.go
```diff
@@ -26,6 +46,59 @@ var (
 )
 
 func InitAWSViper() {
+	_ = viper.BindEnv(AWSSharedCredentials, "AWS_SHARED_CREDENTIALS")
+	RegisterSecret(AWSSharedCredentials, "aws-shared-credentials")
+
+	viper.SetDefault(AWSCredentialsFile, "/tmp/.aws/osde2e/credentials")
```
jbpratt, line 52: Is this default created automatically somewhere? `/tmp` is cleared on every restart so it may be better to put this in the `$HOME/.aws` directory

pkg/common/config/aws_config.go
```diff
@@ -26,6 +46,59 @@ var (
...
+		// use profile based on week
+		week := getWeekSince2024()
+		const profileA = "profile-a"
+		const profileB = "profile-b"
+		currentProfile := ""
+		if week%2 == 0 {
+			currentProfile = profileA
+		} else {
+			currentProfile = profileB
+		}
```
jbpratt, line 87:
```suggestion
		currentProfile := profileA
		if week%2 != 0 {
			currentProfile = profileB
		}
```

pkg/common/config/aws_config.go
```diff
@@ -26,6 +46,59 @@ var (
...
+		os.Setenv("AWS_SECRET_ACCESS_KEY", "")
+		os.Setenv("AWS_ACCOUNT_ID", "")
+		os.Setenv("AWS_SHARED_CREDENTIALS_FILE", viper.GetString(AWSCredentialsFile))
+		os.Setenv("AWS_PROFILE", currentProfile)
+		_ = viper.BindEnv(AWSProfile, "AWS_PROFILE")
+		RegisterSecret(AWSProfile, "aws-profile")
+	} else {
+		_ = viper.BindEnv(AWSProfile, "AWS_PROFILE")
+		RegisterSecret(AWSProfile, "aws-profile")
+	}
```
jbpratt, line 100:
```suggestion
        }
	_ = viper.BindEnv(AWSProfile, "AWS_PROFILE")
	RegisterSecret(AWSProfile, "aws-profile")
```

it seems these are set to the same thing in both cases

pkg/common/config/aws_config.go
```diff
@@ -35,12 +108,22 @@ func InitAWSViper() {
...
+
+// Since simply checking whether current week is odd or even
+// within current year may result in unexpected outage if a year has
+// odd number of weeks, use odd/even based on a constant start date.
+func getWeekSince2024() int {
+	timeFormat := "2006-01-02"
+	t, _ := time.Parse(timeFormat, "2024-01-01")
+	now := time.Now()
+	duration := now.Sub(t)
+	fmt.Println(duration)
```
jbpratt, line 126:
```suggestion
```

pkg/common/config/aws_config.go
```diff
@@ -26,6 +46,59 @@ var (
 )
 
 func InitAWSViper() {
```
jbpratt, line 48: also +1 on returning errors so we can exit gracefully if any of this fails

pkg/common/config/aws_config.go
```diff
@@ -26,6 +46,59 @@ var (
...
+	RegisterSecret(AWSSharedCredentials, "aws-shared-credentials")
+
+	viper.SetDefault(AWSCredentialsFile, "/tmp/.aws/osde2e/credentials")
+	_ = viper.BindEnv(AWSCredentialsFile, "AWS_CREDENTIAL_FILE")
+
+	if viper.GetString(AWSSharedCredentials) != "" {
+		// If shared credntials file is provided in env vars, it should contain two profiles named "profile-a" and profile-b"
+		// Osde2e will use one of them based on current week.
+		// While one profile is in use, the other is cleaned up using AWS nuke
+		err := os.MkdirAll(filepath.Dir(viper.GetString(AWSCredentialsFile)), os.FileMode(0o755))
```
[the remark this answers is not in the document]
jbpratt: Sorry, I just meant the result. You call the get function numerous times

pkg/common/config/aws_config.go
jbpratt: a log.Fatal should be fine

pkg/common/config/aws_config.go
```diff
@@ -25,7 +49,49 @@ var (
...
+	viper.SetDefault(AWSCredentialsFile, os.Getenv("HOME")+"/.aws/osde2e/credentials")
+	_ = viper.BindEnv(AWSCredentialsFile, "AWS_CREDENTIAL_FILE")
+	customCredsPath := viper.GetString(AWSCredentialsFile)
+
+	if sharedCreds != "" {
+		// If shared credntials file is provided in env vars, it should contain two profiles named "profile-a" and profile-b"
+		// Osde2e will use one of them based on current week.
+		// While one profile is in use, the other is cleaned up using AWS nuke
+		if err := os.MkdirAll(filepath.Dir(customCredsPath), os.FileMode(0o755)); err != nil {
+			return fmt.Errorf("Could not write given shared credentials file: %v", err)
```
jbpratt, line 66: Last thing and I'm good to go:

```suggestion
			return fmt.Errorf("Could not write given shared credentials file: %w", err)
```

nit, be sure to use the `%w` directive when you can to wrap errors

jbpratt: (There are multiple instances of this to fix)

jbpratt (approved):
/lgtm

PR #2774, opened 2024-11-07
Add ClusterOperator health check

reedcort:
# What is being added?
This PR increases the wait timeout so clusters with 20+ replicas are given enough time to complete the upgrade. Also it adds a clusterOperatorHealthCheck function to ensure operators are healthy and all nodes are in a schedulable state before proceeding to the next step

### DIFF
--- a/test/sdn_migration/sdn_migration_test.go
+++ b/test/sdn_migration/sdn_migration_test.go
@@ -3,8 +3,8 @@ package sdn_migration_test
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
-	corev1 "k8s.io/api/core/v1"
 	"os"
 	"strconv"
 	"strings"
@@ -20,6 +20,7 @@ import (
 	osdprovider "github.com/openshift/osde2e-common/pkg/openshift/osd"
 	rosaprovider "github.com/openshift/osde2e-common/pkg/openshift/rosa"
 	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/e2e-framework/klient/k8s"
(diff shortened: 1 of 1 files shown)


test/sdn_migration/sdn_migration_test.go
```diff
@@ -156,19 +157,18 @@ var _ = Describe("SDN migration", Ordered, func() {
 	})
 
 	It("rosa cluster is healthy post 4.15.8 upgrade", postUpgradeCheck, func(ctx context.Context) {
-		err := osdClusterReadyHealthCheck(ctx, testRosaCluster.client, "post-upgrade", testRosaCluster.reportDir)
+		err := cluterOperatorHealthCheck(ctx, testRosaCluster.client, logger)
```
joshbranham, line 160: I understand the reasoning behind validating all cluster operators are healthy, but can you explain why we no longer want to call `osdClusterReadyHealthCheck` as part of this change?

abyrne55: An "all COs ≟ healthy" check effectively includes (or is at least more comprehensive than) the checks performed by osd-cluster-ready. @reedcort please correct me if I'm wrong

abyrne55 (approved):
Nice work, @reedcort!

/lgtm

test/sdn_migration/sdn_migration_test.go
reedcort: Yeah, that's correct. The reason I'm not using the 'osdClusterReadyHealthCheck' is that after each upgrade, the nodes will roll and temporarily become unschedulable, which causes certain COs to go from 'Ready' to 'Progressing.' When there are many nodes, the health check job will time out.

test/sdn_migration/sdn_migration_test.go
```diff
@@ -181,11 +181,75 @@ var _ = Describe("SDN migration", Ordered, func() {
 		Expect(err).ShouldNot(HaveOccurred(), "Rosa Cluster failed to patch network")
 	})
 	It("rosa cluster has no critical alerts firing post sdn to ovn migration", postMigrationCheck, func(ctx context.Context) {
-		err := osdClusterReadyHealthCheck(ctx, testRosaCluster.client, "post-upgrade", testRosaCluster.reportDir)
+		err := cluterOperatorHealthCheck(ctx, testRosaCluster.client, logger)
```
jbpratt, line 184: This err is never checked

test/sdn_migration/sdn_migration_test.go
```diff
@@ -181,11 +181,75 @@ var _ = Describe("SDN migration", Ordered, func() {
...
+		} else {
+			for _, co := range coList.Items {
+				// Check if the "Progressing" condition exists and is set to "false"
+				progressingCondition := findConditionByType(co.Status.Conditions, "Progressing")
+				availableCondition := findConditionByType(co.Status.Conditions, "Available")
+				if progressingCondition != nil && progressingCondition.Status == "True" || availableCondition != nil && availableCondition.Status == "False" {
+					logger.Info("waiting for cluster operators to be healthy")
+					state = "unhealthy"
+					break
+
```
jbpratt, line 216:
```suggestion
```
formatting

(it may help to run https://github.com/mvdan/gofumpt against your file to format it and fix a few of these issues)

jbpratt (changes requested):

PR #2785, opened 2024-11-15
OSD-26414- Refactor Proxy OVN to SDN osde2e

reedcort:
Updates the SDN migration e2e test to deploy a multi-az proxy with terraform when 'DefaultBuildWithProxy' is selected.

### DIFF
--- a/test/sdn_migration/terraform/assets/proxy-setup.sh.tftpl
+++ b/test/sdn_migration/terraform/assets/proxy-setup.sh.tftpl
@@ -1,0 +1,92 @@
+#!/bin/bash
+set -x
+
+# Use IP Address for Squid configuration and certs
+export IPADDRESS=$(hostname -i)
+
+# Update DNF packages and install packages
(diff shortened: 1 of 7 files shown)


test/sdn_migration/terraform/assets/proxy-setup.sh.tftpl
joshbranham, line 1: Since the repo is public, can we just clone `osd-network-verifier` in CI and reference the Terraform that way versus copying it all over? Then if we fix or update it, CI has the benefit of getting those fixes. If we feel that is a risk, we could lock the clone to a commit?

reedcort: We can't because the Terraform dir is too different from the osd-network-verifier one. First example the proxy in `osd-network-verifier` was only for single az clusters this is for multi-az. Also the

abyrne55:
Nice work @reedcort!

/lgtm

PR #2822, opened 2025-01-22
SDE-4270 - Update osde2e ocm access to ClientID

MrSantamaria:
Follow up to https://github.com/openshift/osde2e-common/pull/206
Follow up to https://github.com/openshift/osde2e/pull/2824 and https://github.com/openshift/osde2e-common/commit/0a3a3902b38f64bef745b1ad9a38d03395fb87a3

This PR refactors the logic that we were using for FedRamp's Keycloack to enable ServiceAccount login through OSDe2e. 
This change will be enabled by using the credentials currently stored under /ocm-sd-cicada-sa in vault.
This also bumps the latest osde2e-common dependancies

Added the OCM_CONFIG as part of the viper inits, this also sets the env variable for the container. 

Removed the ProdProvider from the old OCM Provider. The new SA does not support Prod, this is by design.

### DIFF
--- a/pkg/common/providers/ocmprovider/ocm.go
+++ b/pkg/common/providers/ocmprovider/ocm.go
@@ -3,9 +3,9 @@ package ocmprovider
 
 import (
 	"fmt"
+	"strings"
 
 	viper "github.com/openshift/osde2e/pkg/common/concurrentviper"
-	"github.com/openshift/osde2e/pkg/common/config"
 	"github.com/openshift/osde2e/pkg/common/spi"
 
 	ocm "github.com/openshift-online/ocm-sdk-go"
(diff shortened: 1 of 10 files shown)


pkg/common/providers/ocmprovider/ocm.go
```diff
@@ -79,12 +77,11 @@ func OCMConnection(token, clientID, clientSecret, env string, debug bool) (*ocm.
...
+	// FedRamp uses a different tokenURL, so we need to check if url contains fr
+	if strings.Contains(url, "fr") {
 		connectionBuilder.Client(clientID, clientSecret).TokenURL(fRTokenURL)
 	} else {
-		// Default commercial flow
-		connectionBuilder.TokenURL(TokenURL).Client(ClientID, "").Tokens(token)
+		connectionBuilder.Client(clientID, clientSecret)
 	}
 
 	connection, err := connectionBuilder.Build()
```
jbpratt, line 87: Could we swap to use the common provider func here?

PR #2860, opened 2025-03-07
Fix: load INSTANCE_TYPE variable as machine type for cluster creation

ritmun:
This variable is passed from jenkins parameterized job and should be loaded to take effect

https://github.com/openshift/osde2e/blob/173723722a7da39d78f56c58b3272865429ad0aa/scripts/parameterized-job.sh#L19


jbpratt (approved):

YiqinZhang (commented):
/lgtm

PR #2910, opened 2025-04-01
misc: drop alert pkg/cmd

jbpratt:
it is unused, lets drop it!


jbpratt:
/override "Red Hat Konflux / osde2e-test-suites-enterprise-contract / fedramp" "Red Hat Konflux / osde2e-test-suites-enterprise-contract / sdn-migration"

PR #2973, opened 2025-06-25
SDCICD-1594: assets: remove unused files

jbpratt:
these are unused and should be removed


jbpratt:
> Could you double check if assets/terraform/setup-vpc.tf is being used by hypershift runs?

Yep, I doubled checked! https://github.com/openshift/osde2e-common/blob/4501a65fd98ff45726bffc4e1e5032ab5be87dc1/pkg/openshift/rosa/assets/setup-hcp-vpc.tf#L4 The file in this repo isn't used anywhere in the repository https://github.com/search?q=repo%3Aopenshift%2Fosde2e%20setup-vpc.tf%20&type=code

PR #3001, opened 2025-08-04
Add NodePool STS Permissions Test for Day 2 Operations

Amarthya-v:
This test validates STS permissions required for ROSA HCP NodePool operations:
- ec2:RunInstances (create EC2 instances for new nodes)
- ec2:CreateTags (tag new instances)  
- ec2:DescribeInstances (validate nodes exist in AWS)

The test performs day 2 operations by creating a NodePool, waiting for nodes 
to provision, and scheduling workloads. Test failures block releases until 
STS permissions are fixed, ensuring proper AWS access for cluster operations.

Fixes: [https://issues.redhat.com/browse/OSD-28896]


pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,232 @@
...
+			if len(parts) > 0 && len(parts[0]) > 10 {
+				return parts[0]
+			}
+		}
+	}
+	return ""
+}
+
+func createNodePool(clusterID, nodePoolName string) error {
+	if _, err := exec.Command("rosa", "whoami").CombinedOutput(); err != nil {
```
btoll, line 127: I don't think we should be shelling out to `rosa` using `exec.Command`.  Aren't there libraries to take advantage of that can do this?

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,232 @@
...
+		
+		if len(nodes.Items) == 0 {
+			ginkgo.Skip("No nodes found")
+		}
+
+		clusterName := extractClusterName(nodes.Items)
+		clusterID := getClusterID(nodes.Items)
+		
+		if clusterName == "" || clusterID == "" {
+			ginkgo.Skip("Could not determine cluster information")
```
btoll, line 52: I think it would be helpful if this string was more specific to what is missing.

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,232 @@
...
+		"--replicas", "1",
+		"--instance-type", "m5.large",
+		"--yes")
+	
+	_, err := cmd.CombinedOutput()
+	return err
+}
+
+func deleteNodePool(clusterID, nodePoolName string) {
+	exec.Command("rosa", "delete", "machinepool", nodePoolName, "--cluster", clusterID, "--yes").Run()
```
btoll, line 145: Same comment as above about using `exec.Command`.

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,232 @@
...
+			RestartPolicy: corev1.RestartPolicyNever,
+		},
+	}
+
+	if err := client.Create(ctx, pod); err != nil {
+		return fmt.Errorf("failed to create test pod: %v", err)
+	}
+	defer client.Delete(ctx, pod)
+
+	// Wait for pod success
```
btoll, line 218: Unnecessary comment.

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,232 @@
...
+		nodePoolName := fmt.Sprintf("osde2e-sts-test-%d", time.Now().Unix())
+		
+		ginkgo.By("Creating NodePool to test STS permissions for day 2 operations")
+		err := createNodePool(clusterID, nodePoolName)
+		
+		if err != nil {
+			expect.NoError(fmt.Errorf("NodePool creation failed - STS permissions missing %v.", err))
+		}
+
+		// Cleanup
```
btoll, line 64: Unnecessary comment.

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,232 @@
...
+	"sigs.k8s.io/e2e-framework/klient/wait"
+)
+
+// NodePool STS Permissions Test
+// This test validates day 2 operations for ROSA HCP clusters by creating NodePools.
+// It ensures required STS permissions (ec2:RunInstances, ec2:CreateTags, ec2:DescribeInstances) 
+
+var nodePoolTestName = "[Suite: e2e] NodePool STS Permissions"
+
+var _ = ginkgo.Describe(nodePoolTestName, ginkgo.Ordered, label.E2E, func() {
```
btoll, line 31: Is this the correct label (`label.E2E`)?

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,460 @@
...
+		if testNodePoolName != "" && ocmConnection != nil {
+			ginkgo.By("Cleaning up test NodePool")
+			deleteNodePool(ocmConnection, clusterID, testNodePoolName)
+		}
+		if ocmConnection != nil {
+			ocmConnection.Close()
+		}
+	})
+
+	ginkgo.Describe("Authentication and Prerequisites", func() {
```
btoll, line 91: I don't think we need to test this here.

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,291 @@
...
+			}
+		}
+	})
+
+	ginkgo.It("should successfully create NodePool", func(ctx context.Context) {
+		ginkgo.By("Getting existing NodePool configuration")
+
+		existingNodePools, err := h.Dynamic().Resource(nodePoolGVR).Namespace(clusterNamespace).List(ctx, metav1.ListOptions{})
+		expect.NoError(err, "Failed to list existing NodePools")
+		Expect(len(existingNodePools.Items)).To(BeNumerically(">", 0), "No existing NodePools found to reference")
```
btoll, line 102: What would be the use case where there isn't a node pool?  And, if there isn't, can't we still proceed here?  (This may be a lack of my understanding, by the way.  Feel free to correct me :) )

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,291 @@
...
+
+		ginkgo.By("Creating test NodePool to validate STS permissions")
+
+		refNodePool := existingNodePools.Items[0]
+		var subnet string
+		if spec, found, err := unstructured.NestedMap(refNodePool.Object, "spec"); found && err == nil {
+			if platform, found, err := unstructured.NestedMap(spec, "platform"); found && err == nil {
+				if aws, found, err := unstructured.NestedMap(platform, "aws"); found && err == nil {
```
btoll, line 110: Should there be separate tests for each platform?  Does that matter?

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,291 @@
...
+					if s, found, err := unstructured.NestedString(aws, "subnet"); found && err == nil {
+						subnet = s
+					}
+				}
+			}
+		}
```
btoll, line 116: Would a helper function be more readable here?  You could give it a nodepool spec and a platform and it gives you a subnet?

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,291 @@
...
+				},
+				"platform": map[string]interface{}{
+					"aws": map[string]interface{}{
+						"instanceType": "m5.large",
+					},
+				},
+			},
+		}
+
+		if subnet != "" {
```
btoll, line 140: What should happen if a subnet isn't found?

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,291 @@
...
+			spec := nodePoolSpec["spec"].(map[string]interface{})
+			platform := spec["platform"].(map[string]interface{})
+			aws := platform["aws"].(map[string]interface{})
+			aws["subnet"] = subnet
+		}
+
+		nodePoolObj := &unstructured.Unstructured{Object: nodePoolSpec}
+		_, err = h.Dynamic().Resource(nodePoolGVR).Namespace(clusterNamespace).Create(ctx, nodePoolObj, metav1.CreateOptions{})
```
btoll, line 148: If this test is creating a new nodepool should it also then tear it down?

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,291 @@
...
+		expect.NoError(client.List(context.Background(), &nodeList))
+
+		if len(nodeList.Items) == 0 {
+			ginkgo.Skip("No nodes found - cannot run NodePool tests")
+		}
+
+		initialNodeCount = len(nodeList.Items)
+
+		for _, node := range nodeList.Items {
+			if label, exists := node.Labels["hypershift.openshift.io/nodePool"]; exists {
```
btoll, line 59: It looks like you're using the string `hypershift.openshift.io/nodePool` several times in this suite, so it would good to make that a `const`.

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,291 @@
...
+			if len(nodeList.Items) > initialNodeCount {
+				for _, node := range nodeList.Items {
+					if label, exists := node.Labels["hypershift.openshift.io/nodePool"]; exists {
+						if strings.Contains(label, testNodePoolName) && isNodeReady(node) {
+							newNodes = append(newNodes, node)
+						}
+					}
+				}
+			}
+			return len(newNodes) > 0, nil
```
btoll, line 173: Shouldn't this comparison be that of `len(newNodes) == len(totalNodes)`?

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,291 @@
...
+		}, wait.WithTimeout(20*time.Minute), wait.WithInterval(30*time.Second))
```
btoll, line 174: 20 minutes seems like an awfully long time.  Can this be lessened?

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,291 @@
...
+
+		expect.NoError(err, "NodePool failed to provision nodes - STS permissions (ec2:RunInstances) may be missing")
+		Expect(len(newNodes)).To(BeNumerically(">", 0), "No new nodes found")
```
btoll, line 177: Again, is this the correct assertion?

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,291 @@
...
+				RestartPolicy: corev1.RestartPolicyNever,
+			},
+		}
+
+		expect.NoError(client.Create(ctx, pod), "Failed to create test pod")
+
+		ginkgo.By("Waiting for workload to complete successfully")
+
+		err := wait.For(conditions.New(client).PodPhaseMatch(pod, corev1.PodSucceeded), wait.WithTimeout(5*time.Minute))
+		expect.NoError(err, "Workload scheduling failed on NodePool")
```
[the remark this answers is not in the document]
Amarthya-v: yes, it's guaranteed because,  we use Node selector with the specific Nodepool label and the test nodepool has only 1 node, If no matching node exists, pod stays Pending (which would fail our test).

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,324 @@
...
+			if platform, found, err := unstructured.NestedMap(spec, "platform"); found && err == nil {
+				if aws, found, err := unstructured.NestedMap(platform, "aws"); found && err == nil {
+					if subnet, found, err := unstructured.NestedString(aws, "subnet"); found && err == nil {
+						return subnet
+					}
+				}
+			}
+		}
+	}
+	return ""
```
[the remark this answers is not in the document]
Amarthya-v: Yes, if the subnet is empty, Hypershift picks one from the cluster’s VPC to ensure networking works, even when Nodepools don’t have a subnet set.

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,291 @@
...
+
+		testNodePoolName = fmt.Sprintf("test-%d", time.Now().Unix()%100000)
+	})
+
+	ginkgo.AfterAll(func(ctx context.Context) {
+		if testNodePoolName != "" {
+			ginkgo.By("Cleaning up test NodePool")
+			err := h.Dynamic().Resource(nodePoolGVR).Namespace(clusterNamespace).
+				Delete(ctx, testNodePoolName, metav1.DeleteOptions{})
+			if err != nil && !apierrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
Amarthya-v: Yes, it’s needed. NotFound just means the Nodepool is already deleted. Without it, cleanup would log false errors instead of real failures.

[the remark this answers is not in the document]
btoll: Right, why would it give a NotFound if it's just deleted the nodepool?  Wouldn't it be some other error?

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,324 @@
...
+		ginkgo.By("Cleaning up test NodePool")
+
+		err := h.Dynamic().Resource(nodePoolGVR).Namespace(clusterNamespace).
+			Delete(ctx, testNodePoolName, metav1.DeleteOptions{})
+		if err != nil && !apierrors.IsNotFound(err) {
+			ginkgo.GinkgoLogr.Error(err, "Failed to cleanup test NodePool", "name", testNodePoolName)
+		}
+		_ = wait.For(func(ctx context.Context) (bool, error) {
+			_, err := h.Dynamic().Resource(nodePoolGVR).Namespace(clusterNamespace).
+				Get(ctx, testNodePoolName, metav1.GetOptions{})
```
btoll, line 76: Why is this trying to get the nodepool if it was just deleted?

btoll: If it's async, then wrap it in a Wait api.

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,324 @@
...
+func detectSTSPermissionError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	errorMsg := strings.ToLower(err.Error())
+	stsPatterns := []string{
+		"accessdenied", "unauthorizedoperation", "forbidden",
+		"invalid iam role", "sts permissions", "assumerolewithwebidentity",
+	}
```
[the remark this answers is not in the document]
btoll: You've resolved this, does that mean that you're confident that nothing else but an STS error would contain these keywords?

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,477 @@
...
+	Expect(hasInternalIP).To(BeTrue(), fmt.Sprintf("Node %s missing InternalIP", node.Name))
+
+	Expect(node.Labels["node.kubernetes.io/instance-type"]).ToNot(BeEmpty(),
+		fmt.Sprintf("Node %s missing instance type", node.Name))
+}
+
+func getClusterNamespace(h *helper.H) (string, error) {
+	if name := viper.GetString(config.Cluster.Name); name != "" {
+		return "clusters-" + name, nil
```
btoll, line 292: You're using `fmt.Sprintf` in other functions but you're not using it here.  Probably be good to be consistent.

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,477 @@
...
+		expect.NoError(err, "Failed to create resource-intensive pod")
+		defer client.Delete(ctx, pod)
+
+		err = wait.For(conditions.New(client).PodPhaseMatch(pod, corev1.PodSucceeded),
+			wait.WithTimeout(5*time.Minute), wait.WithInterval(5*time.Second))
+		expect.NoError(err, "Resource-intensive workload failed")
+	})
+
+	ginkgo.It("rejects operations on non-existent NodePools", func(ctx context.Context) {
+		if clusterNamespace == "" || !canAccessNodePoolAPI(h, clusterNamespace) {
```
btoll, line 195: Once here, `clusterNamespace` should never by an empty string, should it?  And, can the `canAccessNodePoolAPI` test be done before any tests are run?

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,477 @@
...
+		clusterNamespace, err = getClusterNamespace(h)
+		if err != nil {
+			ginkgo.Skip(fmt.Sprintf("Cannot determine cluster namespace: %v", err))
+		}
+
+		testNodePoolName = fmt.Sprintf("sts-test-%d", time.Now().Unix()%100000)
+	})
+
+	ginkgo.AfterAll(func() {
+		if createdNodePool && testNodePoolName != "" && clusterNamespace != "" {
```
btoll, line 66: Is this conditional necessary?  Are any of the tests doing any node pool cleanup?  And, if they are, should they be?

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,477 @@
...
+		expect.NoError(err, "Failed to find NodePool nodes")
+		Expect(len(nodePoolNodes)).To(BeNumerically(">", 0), "No NodePool nodes found")
+
+		for _, node := range nodePoolNodes {
+			validateAWSIntegration(node)
+		}
+	})
+
+	ginkgo.It("creates new NodePool successfully", func(ctx context.Context) {
+		if clusterNamespace == "" {
```
btoll, line 85: Once here, clusterNamespace should never by an empty string, should it? And, can the canAccessNodePoolAPI test be done before any tests are run?

Note, this isn't consistent with the others.

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,477 @@
...
+	ginkgo.It("handles invalid configurations gracefully", func(ctx context.Context) {
+		pod := createPodWithSelector(h.CurrentProject(), map[string]string{
+			nodePoolLabel: "non-existent-nodepool",
+		})
+
+		err := client.Create(ctx, pod)
+		expect.NoError(err, "Should create pod with invalid selector")
+		defer client.Delete(ctx, pod)
+
+		time.Sleep(30 * time.Second)
```
btoll, line 159: Why this arbitrary sleep?

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,477 @@
...
+		Expect(foundPod).ToNot(BeNil(), "Pod should exist")
+		Expect(foundPod.Status.Phase).To(Equal(corev1.PodPending), "Pod should remain pending")
+	})
+
+	ginkgo.It("handles resource-intensive workloads", func(ctx context.Context) {
+		nodePoolNodes, err := findReadyNodePoolNodes(ctx, client)
+		expect.NoError(err, "Failed to find ready NodePool nodes")
+		Expect(len(nodePoolNodes)).To(BeNumerically(">", 0), "No ready NodePool nodes available")
+
+		targetNode := &nodePoolNodes[0]
```
btoll, line 182: Same comment as above about the possibility of several nodepool nodes.

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,477 @@
...
+						corev1.ResourceMemory: resource.MustParse("128Mi"),
+					},
+				},
+			}},
+			RestartPolicy: corev1.RestartPolicyNever,
+		},
+	}
+}
+
+func createPodWithSelector(namespace string, nodeSelector map[string]string) *corev1.Pod {
```
btoll, line 392: It looks like this function is virtually equivalent to `createTestPod`?

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,477 @@
...
+						corev1.ResourceMemory: resource.MustParse("64Mi"),
+					},
+				},
+			}},
+			RestartPolicy: corev1.RestartPolicyNever,
+		},
+	}
+}
+
+func createResourceIntensivePod(namespace, nodeName string) *corev1.Pod {
```
btoll, line 416: You could probably also roll this into the same generic `createPod` function.

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,474 @@
...
+		if err != nil {
+			ginkgo.Skip(fmt.Sprintf("Cannot determine cluster namespace: %v", err))
+		}
+
+		canAccessAPI = canAccessNodePoolAPI(h, clusterNamespace)
+		testNodePoolName = fmt.Sprintf("sts-test-%d", time.Now().Unix()%100000)
+	})
+
+	ginkgo.AfterAll(func() {
+		if createdNodePool {
```
btoll, line 69: Wouldn't you want to move this entire block into the that starts on line 95?

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,474 @@
...
+		expect.NoError(err, "Failed to find NodePool nodes")
+		Expect(len(nodePoolNodes)).To(BeNumerically(">", 0), "No NodePool nodes found")
+
+		for _, node := range nodePoolNodes {
+			validateAWSIntegration(node)
+		}
+	})
+
+	ginkgo.Describe("NodePool creation and provisioning", func() {
+		ginkgo.It("creates new NodePool successfully", func(ctx context.Context) {
```
btoll, line 96: Wouldn't you want to create the nodepool in a `BeforeAll` and then you can get rid of the `createdNodePool` boolean?

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,474 @@
...
+		}
+
+		_, err := h.Dynamic().Resource(nodePoolGVR).Namespace(clusterNamespace).
+			Get(ctx, "non-existent-nodepool", metav1.GetOptions{})
+
+		Expect(err).To(HaveOccurred(), "Should fail accessing non-existent NodePool")
+		Expect(apierrors.IsNotFound(err)).To(BeTrue(), "Should return NotFound error")
+	})
+
+	ginkgo.It("rejects duplicate NodePool creation", func(ctx context.Context) {
```
btoll, line 246: Why not move this into the `Describe` block above for the nodepool?

pkg/e2e/verify/nodepool.go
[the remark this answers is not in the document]
btoll: Can you please answer this instead of just resolving it?

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,474 @@
...
+	patterns := []string{
+		"accessdenied", "unauthorizedoperation", "forbidden",
+		"invalid iam role", "sts permissions", "not authorized",
+		"assumerolewithwebidentity", "invalididentitytoken",
+		"authfailure", "signaturedoesnotmatch",
+	}
+	for _, pattern := range patterns {
+		if strings.Contains(msg, pattern) {
+			ginkgo.Fail(fmt.Sprintf(
+				"STS_PERMISSION_ERROR: %s failed due to missing AWS permissions (%s): %v. "+
```
btoll, line 460: This should be cleaned up.  There's no need to concat strings within a `fmt.Sprintf` method call.

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,474 @@
...
+
+	if config.nodeName != "" {
+		pod.Spec.NodeName = config.nodeName
+	}
+
+	if config.nodeSelector != nil {
+		pod.Spec.NodeSelector = config.nodeSelector
+	}
+
+	if config.cpuLimit != "" && config.memLimit != "" {
```
btoll, line 308: Do you really need resource limits?  Aren't these short-lived pods?  Are you anticipating resource spikes that would need to be throttled?

pkg/e2e/verify/nodepool.go
```diff
@@ -0,0 +1,474 @@
...
+
+	current := h.CurrentProject()
+	if strings.HasPrefix(current, "clusters-") {
+		return current, nil
+	}
+
+	return "", fmt.Errorf("could not determine cluster namespace")
+}
+
+func canAccessNodePoolAPI(h *helper.H, namespace string) bool {
```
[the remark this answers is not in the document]
btoll: Can you isolate all the tests that rely on this API to the same block?

PR #3017, opened 2025-08-19
osde2e-upgrade-runner

Ramakrishn07:
Task - https://issues.redhat.com/browse/OSD-30938

This PR introduces a new Jenkins job script to replicate the functionality of the existing Prow-based osde2e upgrade jobs. The script provisions an OSD AWS cluster on the previous minor version (Y-1), performs the upgrade to the latest minor version (Y), and runs the full osde2e test suite.


btoll:
@Ramakrishn07 Can you change the name?  I think it would be better to name it based on its intent rather than on the env vars used.  If they ever change then the script name would no longer be reflective of it.

PR #3137, opened 2026-03-04
SDCICD-1715 update readme  file

ritmun:
This updated info should reflect in https://pkg.go.dev/github.com/openshift/osde2e#readme-tests


ritmun:
stale, should be reworked on when required

PR #3148, opened 2026-03-11
Fix Slack channel to use saas configured channel

YiqinZhang:
Fixes Slack notifications not being sent for ad-hoc test suite failures when LLM log analysis is disabled.

Decouple notification queueing from EnableAnalysis flag — queueNotification is always called on failure
sendFailureNotification uses SLACK_CHANNEL (SaaS parameter) directly for channel resolution
Per-suite and global notifications are mutually exclusive to prevent duplicates

Notification paths

| Scenario | Per-suite notification | Global notification |
|---|---|---|
| No suites, test failed | None (queue empty) | Sent |
| Suites configured, test failed, analysis on | Sent (with LLM content) | Skipped |
| Suites configured, test failed, analysis off | Sent (without LLM content) | Skipped |
| Provision failed, suites configured | None (tests never ran) | Sent (fallback) |
| Provision failed, no suites | None | Sent |

[SDCICD-1758](https://issues.redhat.com/browse/SDCICD-1758)

### DIFF
--- a/pkg/common/config/config.go
+++ b/pkg/common/config/config.go
@@ -14,7 +14,7 @@ import (
 )
 
 // channel id for #hcm-cicd-notifications
-const DefaultNotificationsChannel = "C06HQR8HN0L"
+const defaultNotificationsChannel = "C06HQR8HN0L"
 
 type Secret struct {
 	FileLocation string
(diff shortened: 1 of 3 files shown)


ritmun:
I think this update is not fixing the issue of suite level slacks not going out. 

The pending collection of notifications is staying empty, because it's still dependent on llm enabled flag, and it's set to false. This is why test suite notifications(deferred) aren't going out. Slack should be independent of log analysis per https://issues.redhat.com/browse/SDCICD-1767.

 Continuing discussion in [slack](https://redhat-internal.slack.com/archives/CMK13BP4J/p1773325844297969?thread_ts=1773254690.132599&cid=CMK13BP4J)

PR #3204, opened 2026-04-08
[SREP-4255] Add --security-group flag to cleanup command

smarthall:
Add a --security-group flag to the osde2e cleanup command that removes leftover security groups from orphaned VPCs whose CloudFormation stacks are in DELETE_FAILED state.

Leftover security groups (e.g. from OCPBUGS-74960) block CloudFormation stack deletion, causing all VPC cleanup to fail silently. This flag is designed to run before --vpc so that the subsequent stack deletion can succeed.

The implementation:
- Finds osde2e VPCs with DELETE_FAILED CloudFormation stacks
- Skips VPCs belonging to active clusters
- Revokes all ingress/egress rules to clear cross-SG dependencies
- Deletes all non-default security groups in those VPCs
- Reports results to Slack summary

### DIFF
--- a/cmd/osde2e/cleanup/cmd.go
+++ b/cmd/osde2e/cleanup/cmd.go
@@ -280,6 +280,7 @@ func run(_ context.Context) (msg Message, err error) {
 	}
 	log.Printf("Found %d active clusters for cleanup operations\n", len(activeClusters))
 
+	// Security groups exist inside VPCs so we want to make sure they are cleaned up before we try to clean up the VPCs
 	if args.securityGroup {
 		sgDeletedCounter := 0
 		sgFailedCounter := 0
--- a/cmd/osde2e/cleanup/cmd.go
+++ b/cmd/osde2e/cleanup/cmd.go
@@ -48,6 +48,7 @@ var args struct {
 	sendSummary     bool
 	ec2             bool
 	vpc             bool
+	securityGroup   bool
 }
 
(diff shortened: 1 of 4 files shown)


cmd/osde2e/cleanup/cmd.go
```diff
@@ -147,6 +149,13 @@ func init() {
 		"Cleanup vpc resources",
 	)
 
+	flags.BoolVar(
+		&args.securityGroups,
+		"security-groups",
```
bmeng: Can we use the singular here to follow the same convention as others?

smarthall: amended the previous commit to address this.

cmd/osde2e/cleanup/cmd.go
```diff
@@ -269,6 +280,16 @@ func run(_ context.Context) (msg Message, err error) {
 	}
 	log.Printf("Found %d active clusters for cleanup operations\n", len(activeClusters))
 
+	if args.securityGroup {
```
ritmun, line 284: Could you add a comment here saying this must be done prior to vpc cleanup? Without context for this PR, subsequently some engineer could move this after vpc and disrupt the functioning.

smarthall: Addressed in commit 22957027807d6f163608240c3d250d6c82e48fac

PR #3205, opened 2026-04-09
[SREP-3734] Use internal registry image for executor artifact sidecar

smarthall:
## Summary
- Replace `busybox:latest` with the internal OpenShift CLI image (`image-registry.openshift-image-registry.svc:5000/openshift/cli:latest`) for the `pause-for-artifacts` sidecar container in the ad-hoc test executor
- The `busybox:latest` pull from Docker Hub is subject to rate limits and transient 503 errors, which caused a race condition where the sidecar hadn't started by the time artifact collection was attempted after a fast-completing test
- The internal CLI image is already used elsewhere in the codebase (`pkg/e2e/osd/rebalance_infra_nodes.go`) and is pre-cached on cluster nodes

## Test plan
- [ ] Verify the `pause-for-artifacts` container starts reliably in ad-hoc test executor jobs
- [ ] Confirm artifact collection succeeds for fast-completing test suites
- [ ] Run `periodic-ci-openshift-osde2e-main-nightly-4.21-osd-aws` and verify no `container not found` errors

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/common/executor/executor.go
+++ b/pkg/common/executor/executor.go
@@ -242,7 +242,7 @@ func (e *Executor) buildJobSpec(namespace string, image string) *batchv1.Job {
 						},
 						{
 							Name:    "pause-for-artifacts",
-							Image:   "busybox:latest",
+							Image:   "image-registry.openshift-image-registry.svc:5000/openshift/cli:latest",
 							Command: []string{"tail", "-f", "/dev/null"},
 							VolumeMounts: []corev1.VolumeMount{
 								{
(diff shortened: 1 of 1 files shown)


smarthall:
I made this PR as there were issues with the tests receiving errors when pulling from the docker registry, and because of this the output and artifacts of a test was being lost.

I chose this image as it is already [used elsewhere in the tests](https://github.com/openshift/osde2e/blob/main/pkg/e2e/osd/rebalance_infra_nodes.go#L69), so I presumed that this wasn't a concern. If that's not the case we might need to address that existing code too.

I feel that depending on the ocp image registry is the more reliable option then depending on a third party registry. However the other containers in the pods pull from quay, would you be more comfortable with a `ubi-minimal` image from quay or the red hat registry? Would you like me to update the image used for the rebalancing tests too?

wking (approved):
/lgtm

+1 from me on testing based on the components we sell to customers and need to maintain, and not testing unrelated external services that we can't fix when they break.

ritmun:
//lgtm

PR #3216, opened 2026-04-21
[SDCICD-1816]  size Prometheus token to workload and ensure cluster expiry covers krkn-ai run

varunraokadaparthi:
- Size the Prometheus token lifetime to generations × population × 5 min instead of the fixed default so it does not expire mid-run.
- Extend the cluster OCM expiration before krkn-ai runs when the current expiry is too short, and restore it to a short grace period afterward.

Co-Authored-By: Cursor <noreply@cursor.com> 

## Summary by CodeRabbit

* **New Features**
  * Automatic cluster expiration extension during KrknAI runs to prevent interruptions. Cluster expiration is extended before execution begins and can be restored to its original state afterward.

* **Tests**
  * Added unit tests for cluster expiration duration calculations.

* **Chores**
  * Updated upstream Go module dependencies to latest compatible versions.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,11 +21,11 @@ require (
 	github.com/mitchellh/mapstructure v1.5.0
 	github.com/onsi/ginkgo/v2 v2.28.1
 	github.com/onsi/gomega v1.39.1
-	github.com/openshift-online/ocm-sdk-go v0.1.498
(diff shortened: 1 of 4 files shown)


varunraokadaparthi:
@coderabbitai review

PR #3218, opened 2026-04-23
Revert "[SREP-3734] Use internal registry image for executor artifact sidecar"

ritmun:
Reverts openshift/osde2e#3205


ritmun:
see https://redhat-internal.slack.com/archives/CMK13BP4J/p1776938469361319

PR #3262, opened 2026-06-18
[SDCICD-1858] Sanitize artifacts before S3 upload

YiqinZhang:
## Summary

- S3 upload path now redacts secrets (AWS keys, tokens, passwords, private keys, etc.) from artifact files before uploading to `osde2e-logs` bucket
- Enable sanitization for LLM analysis aggregator prompt path (was wired but never activated)
- Fail-open design: sanitization failures do not block uploads

## Test plan

- [x] Unit tests for `prepareUploadBody` (secret redaction, fail-open, nil sanitizer)
- [x] `go vet` clean
- [x] All existing tests pass unchanged

JIRA: https://issues.redhat.com/browse/SDCICD-1858

Co-authored-by: [Cursor](https://cursor.com)

## Release Notes

* **New Features**
  * S3 uploads now automatically detect and redact sensitive values in text-based files before uploading, enhancing data security and privacy protection.

* **Tests**
  * Added test coverage for the new file sanitization functionality during upload operations.

### DIFF
--- a/pkg/e2e/adhoctestimages/adhoctestimages.go
+++ b/pkg/e2e/adhoctestimages/adhoctestimages.go
@@ -173,7 +173,7 @@ func runLogAnalysisForAdHocTestImage(ctx context.Context, logger logr.Logger, te
 		},
 		PromptTemplate:  "default",
 		FailureContext:  err.Error(),
-		SanitizerConfig: &sanitizer.Config{},
+		SanitizerConfig: &sanitizer.Config{EnableAudit: false},
 	}
 
 	engine, err := analysisengine.New(ctx, engineConfig)
--- a/pkg/e2e/adhoctestimages/adhoctestimages.go
+++ b/pkg/e2e/adhoctestimages/adhoctestimages.go
@@ -12,6 +12,7 @@ import (
 	. "github.com/onsi/gomega"
 	"github.com/openshift/osde2e-common/pkg/clients/ocm"
 	"github.com/openshift/osde2e/internal/analysisengine"
+	"github.com/openshift/osde2e/internal/sanitizer"
 	viper "github.com/openshift/osde2e/pkg/common/concurrentviper"
 	"github.com/openshift/osde2e/pkg/common/config"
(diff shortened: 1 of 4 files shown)


pkg/e2e/adhoctestimages/adhoctestimages.go
```diff
@@ -170,8 +171,9 @@ func runLogAnalysisForAdHocTestImage(ctx context.Context, logger logr.Logger, te
 				Version:       viper.GetString(config.Cluster.Version),
 			},
 		},
-		PromptTemplate: "default",
-		FailureContext: err.Error(),
+		PromptTemplate:  "default",
+		FailureContext:  err.Error(),
+		SanitizerConfig: &sanitizer.Config{},
```
ritmun: The log says "sanitized %s: redacted %d secret(s)", relPath" , so shouldn't this have those rules set in config here?

pkg/e2e/e2e.go
```diff
@@ -307,8 +308,9 @@ func (o *E2EOrchestrator) AnalyzeLogs(ctx context.Context, testErr error) error
 				Version:       viper.GetString(config.Cluster.Version),
 			},
 		},
-		PromptTemplate: "default",
-		FailureContext: testErr.Error(),
+		PromptTemplate:  "default",
+		FailureContext:  testErr.Error(),
+		SanitizerConfig: &sanitizer.Config{},
```
ritmun: does this config also require some configs set in it?

pkg/e2e/adhoctestimages/adhoctestimages.go
YiqinZhang: Sanitization rules are loaded internally via [getDefaultRules()](https://github.com/YiqinZhang/osde2e/blob/main/internal/sanitizer/rules.go) in sanitizer.New() and run unconditionally.
Config only controls audit logging, content size limits, and strict mode. 
Updated the config to use EnableAudit: false explicitly so the intent is clear.

PR #3283, opened 2026-09-01
 [SDCICD-1948] : Add Secrets Manager leftover cleanup for unmanaged ROSA OIDC keys

minlei98:
[SDCICD-1948] Add Secrets Manager leftover cleanup for unmanaged ROSA OIDC keys

Force-delete only osde2e-owned leftovers, and require a complete int/stage/prod
cluster and OIDC ARN skip set so live keys from other clusters are not removed.

[SDCICD-1948]: https://redhat.atlassian.net/browse/SDCICD-1948?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ritmun (commented):
One placement thought that spans both PRs: the AWS secrets-cleanup logic added to `osde2e-common` in openshift/osde2e-common#652 is fairly osde2e-specific (it keys off the `osde2e` ownership marker) and pulls new deps into a shared library. Would it fit better here in osde2e, alongside the existing s3/ec2/vpc/iam/securitygroups cleanups under `pkg/common/aws`? The generic OCM OIDC-config helpers seem like the natural fit for `-common`. Curious what you think.

